PR #19, opened 2020-12-11
Token rules from truffleHog

eddynaka:


michaelcfanning:
Closing this. We shouldn't be picking up rules from trufflehog, which has a copyleft license. :)

PR #20, opened 2020-12-11
Incrementing Token rules

eddynaka:


Src/Plugins/PlaintextSecrets/PlaintextSecrets.json
```diff
@@ -4,15 +4,40 @@
       "Name": "DoNotExposePlaintextSecrets",
       "Id": "SEC1001",
       "Level": "Error",
+      "DefaultNameRegex": "(?i)\\.(azure|bat|cmd|config|cpp|cs|cscfg|definitions|dtsx|ini|java|jsx?|json|keys|loadtest|m|md|php|properties|ps1|psm1|pubxml|py|resx|sample|sql|ste|swift|test|tsx?|txt|waz|xml)$",
```
michaelcfanning, line 7: >cpp [](start = 56, length = 3)

Add c, hxx, h, hpp? Let's continue to discuss the list of file extensions offline.

Src/Plugins/PlaintextSecrets/PlaintextSecrets.json
```diff
@@ -4,15 +4,40 @@
...
           "MessageArguments": { "secretKind": "Azure DevOps personal access token" }
+        },
+        {
+          "SubId": "FacebookOAuthToken",
+          "ContentsRegex": "(?i)facebook.{0,30}['\"\\s]([0-9a-f]{32})['\"\\s]",
+          "MessageArguments": { "secretKind": "Facebook OAuth token" }
+        },
+        {
+          "SubId": "TwitterOAuthToken",
+          "ContentsRegex": "(?i)twitter.{0,30}['\"\\s]([0-9a-zA-Z]{35,44})['\"\\s]",
```
michaelcfanning, line 24: >0-9a-zA-Z [](start = 56, length = 9)

We should test but I don't think this is required, because the entire match is case-insensitive due to (?i). Should be 0-9a-z? Definitely test if we change. :)

Src/Plugins/PlaintextSecrets/PlaintextSecrets.json
```diff
@@ -4,15 +4,40 @@
...
+          "MessageArguments": { "secretKind": "Twitter OAuth token" }
+        },
+        {
+          "SubId": "GoogleOAuth",
+          "ContentsRegex": "\"client_secret\":\"([a-zA-Z0-9-_]{24})\"",
```
michaelcfanning, line 29: >zA [](start = 52, length = 2)

use (?i) in this pattern and drop this to a-z0-9. Actually, let's please use a common pattern for this everywhere, and let's make it a-z0-9 or  0-9a-z, whichever you prefer. This consistency will help with readability. ALSO: we need to be absolutely sure that all these tokens actually permit a mixture of upper and lowercase alpha. I'm not %100 sure, so let's chase that down.

eddynaka: validated locally and it matches with a client_secret that i have. The format is:
15 letters or numbers (upper or lower case) + "-" + 2 letters or numbers (upper or lower case) + "-" + 5 letters or numbers (upper or lower case).

---
In reply to: [541058036](https://github.com/microsoft/sarif-pattern-matcher/pull/20#discussion_r541058036) [](ancestors = 541058036)

michaelcfanning (approved):
:shipit:

PR #21, opened 2020-12-11
Provide enanced dynamic validation capability.

michaelcfanning:
@eddynaka

### DIFF
--- a/Src/SarifPatternMatcher/Validation.cs
+++ b/Src/SarifPatternMatcher/Validation.cs
@@ -5,9 +5,46 @@ namespace Microsoft.CodeAnalysis.SarifPatternMatcher
 {
     public enum Validation
     {
+        /// <summary>
+        /// A default value that reflects an uninitialized state,
+        /// (i.e., no useful validation state is available).
+        /// </summary>
         None = 0,
+
+        /// <summary>
+        /// A validator for the current matched pattern
+        /// could not be found.
+        /// </summary>
         ValidatorNotFound = None,
-        Valid,
+
+        /// <summary>
+        /// A validator for the current matched pattern returned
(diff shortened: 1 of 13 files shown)


Src/SarifPatternMatcher/Validation.cs
```diff
@@ -5,9 +5,46 @@ namespace Microsoft.CodeAnalysis.SarifPatternMatcher
...
+        /// A default value that reflects an uninitialized state,
+        /// (i.e., no useful validation state is available).
+        /// </summary>
         None = 0,
+
+        /// <summary>
+        /// A validator for the current matched pattern
+        /// could not be found.
+        /// </summary>
         ValidatorNotFound = None,
```
eddynaka, line 18: >None [](start = 28, length = 4)

If we set None, would this be able to distinguish a value 0 from `None` or `ValidatorNotFound`?

eddynaka:
>           "ContentsRegex": "(?i)facebook.{0,30}['\"\\s]([0-9a-f]{32})['\"\\s]",

should we add (?<fingerprint> here as well?

---
Refers to: Src/Plugins/PlaintextSecrets/PlaintextSecrets.json:19 in ac2601f. [](commit_id = ac2601fedc1235ce30a48fdad00ba9a466277a3b, deletion_comment = False)

Src/SarifPatternMatcher/Validation.cs
michaelcfanning: Yes, so None is basically an entirely uninitialized value. 

---
In reply to: [541451683](https://github.com/microsoft/sarif-pattern-matcher/pull/21#discussion_r541451683) [](ancestors = 541451683)

eddynaka (approved):

PR #33, opened 2020-12-17
Refactor plugins tests

eddynaka:
## Changes
- removed the project `Test.FunctionalTests.AllPlugins.csproj`
- moved the file `EndToEndTests.cs` to src\Plugins directory
- renamed Test.UnitTests.PluginName to Tests.PluginName, with that, we will add all tests from that PluginName to that project only
  - added compile reference to the `EndToEndTests.cs`
- renamed enum and EndToEndTests.cs namespace (they were wrong)


michaelcfanning (approved):
:shipit:

PR #34, opened 2020-12-17
Test base64 decoding. Return no results when validator returns 'no match'.

michaelcfanning:
@eddynaka


Src/Plugins/PlaintextSecrets/PlaintextSecrets.json
```diff
@@ -10,7 +10,7 @@
       "MatchExpressions": [
         {
           "SubId": "AzureDevOpsPersonalAccessToken",
-          "Base64DecodedContentLength": 70,
+          "MatchLengthToDecode": 52,
```
michaelcfanning, line 13: >MatchLengthToDecode [](start = 11, length = 19)

Note this change! We now specify the length of the non-encoded value (as this is a known value). We then *compute* the encoded length from this. Changed the property name to reflect the difference in meaning.

Src/Plugins/Test.FunctionalTests.AllPlugins/TestData/PlaintextSecrets/ExpectedOutputs/VariousPats.sarif
```diff
@@ -10,168 +10,90 @@
...
-                "artifactLocation": {
-                  "uri": "src/Plugins/Test.FunctionalTests.AllPlugins/TestData/PlaintextSecrets/Inputs/VariousPats.txt",
-                  "uriBaseId": "SRC_ROOT"
+              ],
+              "message": {
+                "text": "An exception of type 'ArgumentOutOfRangeException' was raised analyzing 'VariousPats.txt' for check 'SearchSkimmer' (which has been disabled). The exception may have resulted from a problem related to parsing the analysis target, and not specific to the rule, however."
+              },
+              "level": "error",
+              "exception": {
+                "kind": "ArgumentOutOfRangeException",
```
michaelcfanning, line 35: >ArgumentOutOfRangeException [](start = 25, length = 27)

Thanks to us now honoring 'no match' properly, this SARIF log now has two false positive eliminated from it (two patterns that don't pass the PAT CRC check).

Src/SarifPatternMatcher/SearchSkimmer.cs
```diff
@@ -16,13 +16,13 @@ namespace Microsoft.CodeAnalysis.SarifPatternMatcher
 {
     public class SearchSkimmer : Skimmer<AnalyzeContext>
     {
-        private const string Base64DecodingFormatString = @"\b(?i)[0-9a-z\/+]{{{0}}}==";
```
michaelcfanning, line 19: >}}}=="; [](start = 81, length = 7)

Oops. This regex specified padding when it shouldn't have. The padding depends on the length arg injected into this format string. We now compute the padding (later).

Src/SarifPatternMatcher/SearchSkimmer.cs
```diff
@@ -159,15 +174,42 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
...
+
+                if (_validators != null)
                 {
-                    // TODO add a trace for this condition.
-                    continue;
+                    state = _validators.Validate(matchExpression.SubId, fingerprint, dynamic);
+
+                    switch (state)
+                    {
+                        case Validation.NoMatch:
```
michaelcfanning, line 185: >NoMatch [](start = 40, length = 7)

We need tests for all these cases! I will author some next.

eddynaka (approved):
:shipit:

PR #44, opened 2020-12-18
Fixing analysis for file only

### LINKED ISSUE #42: Review Sensitive File should return only one result
Today, if we run our test for the `ReviewSensitiveFile`, you will see that the output sarif is expecting many results because it is validating many times in the method `RunMatchExpression`.

What I would expect is: if we analyze a file that is an issue, we will just show one result pointing to the file.

What is happening today: it is analyzing all characters inside the file.

cc: @michaelcfanning

eddynaka:
Fixes #42


Src/SarifPatternMatcher/SearchSkimmer.cs
```diff
@@ -175,98 +175,150 @@ public override void Analyze(AnalyzeContext context)
...
+                (!string.IsNullOrEmpty(matchExpression.SubId) ?
+                    "/" + matchExpression.SubId :
+                    string.Empty);
 
-            foreach (FlexMatch flexMatch in _engine.Matches(searchText, matchExpression.ContentsRegex))
+            FailureLevel level = matchExpression.Level != 0 ?
+                matchExpression.Level :
+                DefaultConfiguration.Level;
+
+            if (!string.IsNullOrEmpty(matchExpression.ContentsRegex))
```
michaelcfanning, line 187: There is significant conditional statement complexity in this method, we should start to refactor into methods,

michaelcfanning:
Great change! But I think we need to add some more focused unit tests that cover all possibilities in this scenario, it's too easy to miss changes that fly by in the log file. I checked in this code without observing that fake.sarif was so enormous for example! oops.

See SearchSkimmerTests for the beginnings of some patterns. Also consider whether we need a test at the AnalyzeCommand level, i'm not sure we do but interested in your opinion.

Src/SarifPatternMatcher/SearchSkimmer.cs
eddynaka: sure! separated in two methods.

---
In reply to: [545985868](https://github.com/microsoft/sarif-pattern-matcher/pull/44#discussion_r545985868) [](ancestors = 545985868)

eddynaka:
just added more tests and fixed the code coverage. Let me know what do you think :)

---
In reply to: [748224385](https://github.com/microsoft/sarif-pattern-matcher/pull/44#issuecomment-748224385) [](ancestors = 748224385)

Src/Test.UnitTests.SarifPatternMatcher/SearchSkimmerTests.cs
```diff
@@ -15,19 +15,35 @@ namespace Microsoft.CodeAnalysis.SarifPatternMatcher.SarifPatternMatcher
...
+            };
+        }
+
+        private static MatchExpression CreateEmptyMatchExpression()
+        {
+            return new MatchExpression();
+        }
+
         [Fact]
-        public void SearchSkimmer_DetectsBase64EncodedPattern()
```
michaelcfanning, line 30: >SearchSkimmer_DetectsBase64EncodedPattern [](start = 20, length = 41)

This is our test naming convention to put the name of the type under test as a prefix. it helps with search in the test explorer. You probably don't like it because of style rules that flag the underscore?? :) we should have a stylecop ruleset specifically for test code that turns this one off.

michaelcfanning (approved):
:shipit:

Src/Test.UnitTests.SarifPatternMatcher/SearchSkimmerTests.cs
eddynaka: I just think that that is strange. Today, it's organized like this:
1. project name
  - namespace
     - class name
         - test name

So, in theory, its a duplicated work, since test name = classname_testname. Either way, both works :)

---
In reply to: [546093717](https://github.com/microsoft/sarif-pattern-matcher/pull/44#discussion_r546093717) [](ancestors = 546093717)

PR #47, opened 2020-12-19
Copying .spam to sln and fixing targets

### LINKED ISSUE #48: Copy json files from Security to solution folder
Idea: when we build the solution, our JSON files inside the Security project should be copied to a folder called `.spam` where the solution exists.

Suggestion: use MSBuild tasks to accomplish this.

eddynaka:
Fixes #48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 
 # Dev machine files
 appsettings.local.json
-msbuild.binlog
 
 # User-specific files
 *.suo
@@ -224,3 +223,6 @@ GeneratedArtifacts/
 _Pvt_Extensions/
 ModelManifest.xml
 *.binlog
+
+# SPAM configuration
+.spam/
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -13,8 +13,15 @@
     </Content>
   </ItemGroup>
 
-  <Target Name="PostBuild" AfterTargets="PostBuildEvent">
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,7 +3,6 @@
 
 # Dev machine files
 appsettings.local.json
-msbuild.binlog
```
eddynaka, line 6: Removed because we already have this in line 226, which is more generic

.gitignore
```diff
@@ -224,3 +223,6 @@ GeneratedArtifacts/
 _Pvt_Extensions/
 ModelManifest.xml
 *.binlog
+
+# SPAM configuration
+.spam/
```
eddynaka, line 228: we will ignore .spam folders

Src/Plugins/Security/Security.csproj
```diff
@@ -8,9 +8,13 @@
     <Content Include="*.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
-    <Content Include="build\SEC1001.SecurePlaintextSecrets.targets">
+    <Content Include="build\Security.targets">
```
eddynaka, line 11: fixed to correct naming

Src/Plugins/Security/Security.csproj
```diff
@@ -8,9 +8,13 @@
...
       <PackagePath>build\</PackagePath>
     </Content>
   </ItemGroup>
+
+  <Target Name="PostBuild" AfterTargets="PostBuildEvent">
+    <Exec Command="xcopy /Y /R &quot;$(MSBuildThisFileDirectory)\*.json&quot; &quot;$(SolutionDir).spam\*.json&quot;" />
```
eddynaka, line 17: after build, we will copy to .spam inside sln

PR #59, opened 2020-12-22
When we build, package will generate .spam/Security folder with content

### LINKED ISSUE #58: We shouldn't flatten all files to .spam folder
If we do this, we can't easily support plug-ins with validator files that have different dependencies.

We should probably have a top-level directory with the plugin name:

.spam\Security

eddynaka:
Fixes #58


michaelcfanning (approved):
:shipit:

PR #69, opened 2020-12-29
Add utilities class for validation plugins.

michaelcfanning:
@eddynaka


Targets/build.plugins.props
```diff
@@ -2,6 +2,7 @@
 
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)..\src\Plugins\ValidationState.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)..\src\Plugins\Utilities.cs" />
```
eddynaka, line 5: Where is the file? Couldn't find it

eddynaka (approved):

PR #71, opened 2020-12-30
Adjust failure level appropriately when dynamic validation is in play.

michaelcfanning:
@eddynaka

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -292,23 +292,27 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
                             // The validator returned a bad value.
                             // TODO: we should log this condition
                             // and then continue processing.
+                            level = FailureLevel.Error;
                             break;
                         }
 
                         case Validation.Valid:
                         {
+                            level = FailureLevel.Error;
                             validationState = " which was determined to be valid";
                             break;
                         }
 
                         case Validation.Invalid:
                         {
+                            level = FailureLevel.Warning;
                             validationState = " which was determined to be invalid";
                             break;
                         }
 
                         case Validation.InvalidForConsultedAuthorities:
                         {
+                            level = FailureLevel.Warning;
                             validationState = " which was determined to be invalid for all consulted authorities";
                             break;
                         }


eddynaka (approved):

michaelcfanning:
>                         case Validation.ValidatorReturnedIllegalValue:

Interested in your thoughts here. The plugins can return an arbitrary failure level when they perform work. But, I think we should have some default handling based on certain dynamic validation return values. But we shouldn't override the default/plugin failure level in other cases. See if you agree I've made good decisions. :) Since this condition indicates a misbehaving plug-in, I've made it an error for now. Really, we should probably log a separate error notification and then process the result as is.

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:290 in 5a0ac21. [](commit_id = 5a0ac2193b06774395610efbc97b65ec6791c1ec, deletion_comment = False)

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -292,23 +292,27 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
                             // The validator returned a bad value.
                             // TODO: we should log this condition
                             // and then continue processing.
+                            level = FailureLevel.Error;
                             break;
                         }
 
                         case Validation.Valid:
                         {
+                            level = FailureLevel.Error;
```
michaelcfanning, line 301: >level [](start = 28, length = 5)

If we conclusively validate something, it's important! So we make it an error.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -292,23 +292,27 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
...
                             validationState = " which was determined to be valid";
                             break;
                         }
 
                         case Validation.Invalid:
                         {
+                            level = FailureLevel.Warning;
```
michaelcfanning, line 308: >level [](start = 28, length = 5)

If you conclusively marked something as invalid, that's a warning only, it shouldn't ever break a process (like an error will).

Src/Sarif.PatternMatcher/SearchSkimmer.cs
eddynaka: yeah, we should think how we will log, so we can decide in the future based on data.

eddynaka: but, ur decision is perfect based on our knowledge so far

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -292,23 +292,27 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
...
                             validationState = " which was determined to be invalid";
                             break;
                         }
 
                         case Validation.InvalidForConsultedAuthorities:
```
michaelcfanning, line 313: >InvalidForConsultedAuthorities [](start = 40, length = 30)

We did some checking and it *looks* invalid, so, we'll set it to a warning. This assumes the user has configured the scan to check for the most critical authorities.

michaelcfanning:
>                         case Validation.Unknown:

So, in all cases, when the answer is Unknown, we preserve the failure level set by the rule/plugin.

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:320 in 5a0ac21. [](commit_id = 5a0ac2193b06774395610efbc97b65ec6791c1ec, deletion_comment = False)

PR #80, opened 2021-01-02
Fingerprints and multiline rules

michaelcfanning:
#eddynaka

### DIFF
--- a/Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SocialSecurityNumbers.sarif
+++ b/Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SocialSecurityNumbers.sarif
@@ -23,9 +23,6 @@
                   "text": "'{0}' contains an apparent {1}."
                 }
               },
-              "defaultConfiguration": {
-                "level": "note"
-              },
               "helpUri": "https://github.com/microsoft/sarif-pattern-matcher"
             }
           ]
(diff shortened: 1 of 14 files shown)


Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SocialSecurityNumbers.sarif
```diff
@@ -23,9 +23,6 @@
                   "text": "'{0}' contains an apparent {1}."
                 }
               },
-              "defaultConfiguration": {
```
michaelcfanning, line 26: >defaultConfiguration [](start = 15, length = 20)

We have an issue with specifying a default configuration currently. If you do this, you can't remap a result level to 'Warning' because that's a default value which is stripped at serialization time. Let's talk about this later!

Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -23,5 +23,11 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             HelpText = "Enable dynamic validations, if any (may compromise " +
                        "performance and/or result in calls across the network).")]
         public bool DynamicValidation { get; internal set; }
+
+        [Option(
+            'm',
+            "multiline",
+            HelpText = "Enable multiline regular expression search.")]
```
michaelcfanning, line 30: >Enable [](start = 24, length = 6)

I haven't actually plumbed this through yet. Will do tomorrow.

eddynaka (approved):

PR #83, opened 2021-01-05
Fix fingerprint emit. Fix unauthorized reporting.

michaelcfanning:
@eddynaka @cfaucon 

We weren't actually persisting fingerprints properly, now fixed. Also, the 'unauthorized' helper omitted the '#' delimiter, with the result that these issues were elevated to error (because the validator output could not be parsed).

We need to make this condition produce a useful error! Probably any result in these circumstances s/be dropped to a warning, I'll make that change soon.


eddynaka (approved):
:shipit:

PR #90, opened 2021-01-08
Push data to match expressions

michaelcfanning:
@eddynaka 

This change provides a model for populating replacement strings directly from disk. This will allow us to store complex regular expressions without requiring any escaping. This will greatly facilitate developing complex patterns. :)

Also, we should switch to instantiating a skimmer *per rule*. We have a problem today, we support two models: one in which a definition refers to a single rule, with multiple match expressions. The other in which a parent definition flows common information to each match expression. This flexibility isn't serving us well. :) We'll talk more about cleaning it up. For now, I'm aggressively pushing inherited data (and replacing global strings) in match expressions before passing them to the skimmers...

### DIFF
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -16,10 +16,17 @@
     <Content Include="build\Sarif.PatternMatcher.Security.targets">
       <PackagePath>build\</PackagePath>
     </Content>
+    <Content Include="Security.SharedStrings.txt">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
   </ItemGroup>
 
   <ItemGroup>
     <SpamFiles Include="$(MSBuildThisFileDirectory)*.json" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="Security.SharedStrings.txt" />
   </ItemGroup>
 
   <Target Name="CopyingSpam" AfterTargets="Build">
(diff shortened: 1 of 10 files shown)


Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -101,6 +106,67 @@ public class AnalyzeCommand : MultithreadedAnalyzeCommandBase<AnalyzeContext, An
             return skimmers;
         }
 
+        internal static void PushInheritedData(SearchDefinitions definitions, Dictionary<string, string> sharedStrings)
```
michaelcfanning, line 109: >PushInheritedData [](start = 29, length = 17)

We should have @jameswinkler provide some unit tests here! Great starter task.

Src/Plugins/Security/Security.csproj
```diff
@@ -16,12 +16,19 @@
     <Content Include="build\Sarif.PatternMatcher.Security.targets">
       <PackagePath>build\</PackagePath>
     </Content>
+    <Content Include="Security.SharedStrings.txt">
```
eddynaka, line 19: >Security.SharedStrings.txt [](start = 22, length = 26)

do we need to change the targets file as well?

michaelcfanning: Yes!

---
In reply to: [553691486](https://github.com/microsoft/sarif-pattern-matcher/pull/90#discussion_r553691486) [](ancestors = 553691486)

michaelcfanning: Fixed it! Please merge if you are satisfied.

---
In reply to: [553726183](https://github.com/microsoft/sarif-pattern-matcher/pull/90#discussion_r553726183) [](ancestors = 553726183,553691486)

eddynaka (approved):

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
jameswinkler: Looking into it

PR #96, opened 2021-01-09
Validate PFX files

eddynaka:


Scripts/BuildAndTest.ps1
```diff
@@ -63,9 +63,12 @@ If (Test-Path "..\bld") {
 if (-not $NoBuild) {
     Write-Information "Building RE2.Native.sln (MSBuild)..."
     MSBuild "$RepoRoot\Src\RE2.Native.sln" /p:Configuration=$Configuration /p:Platform="Any CPU"
+    
+    Write-Information "Restoring SarifPatternMatcher.sln (dotnet)..."
+    dotnet restore $RepoRoot\Src\SarifPatternMatcher.sln
```
eddynaka, line 68: >dotnet restore $RepoRoot\Src\SarifPatternMatcher.sln [](start = 4, length = 52)

Azure pipeline is having a hard time without this. So, for now i'm forcing the restore.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -411,7 +411,119 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
...
+                    region: null,
+                    flexMatch: null,
+                    null,
+                    matchExpression.Fixes,
+                    arguments);
+
+            context.Logger.Log(reportingDescriptor, result);
+        }
+
+        private string NormalizeValidator(string validatorMessage)
```
michaelcfanning, line 526: >NormalizeValidator [](start = 23, length = 18)

This should be named 'NormalizedValidatorMessage',my mistake from previous change.

eddynaka: Ok, will rename :)

---
In reply to: [554470118](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470118) [](ancestors = 554470118)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
+{
+    internal static class PFXCryptographicKeyFileValidator
```
michaelcfanning, line 11: >PFXCryptographicKeyFileValidator [](start = 26, length = 32)

Please correct casing of this to PfxCryptographicKeyfileValidator. Note casing of 'Keyfile' this is one word. Change all rule id references as well. I notice that we aren't always capitalizing API as Api as well, we'll make a pass later.

One thing, you may have trouble changing the casing of this file name, I don't know. It's possible that your last casing will be retained when I sync (hopefully). To be sure, you can temporarily rename the file to foo.cs, push a commit, then rename to final casing (this works for sure).

[https://en.wikipedia.org/wiki/Keyfile](https://en.wikipedia.org/wiki/Keyfile)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+#pragma warning disable IDE0060 // Remove unused parameter
+        public static string IsValid(
+            ref string matchedPattern,
+            ref Dictionary<string, string> groups,
+            ref bool performDynamicValidation,
+            ref string failureLevel,
+            ref string fingerprint)
+        {
+#pragma warning restore IDE0060
+            fingerprint = $"[cert={matchedPattern}]";
```
michaelcfanning, line 22: >fingerprint [](start = 12, length = 11)

What's the input matched pattern here? Is it just the file name? We could use that, I suppose, but better would be to find an identifier, like the public key token, that is consistent no matter what the binary container file is named.

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+            // This plugin does not perform any dynamic validation.
```
michaelcfanning, line 24: >This plugin does not perform any dynamic validation [](start = 15, length = 51)

This isn't true! :) remove the comment and change line 28 to return true. Make sure not to proceed on this code path if the input value is false, though.

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: its the fullpath. c:\path\file.pfx, for example

---
In reply to: [554470486](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470486) [](ancestors = 554470486)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: if that's the case...should we change the behavior for the public azure devops validator? right now, it is the same thing.

---
In reply to: [554470513](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470513) [](ancestors = 554470513)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: ok

---
In reply to: [554470314](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470314) [](ancestors = 554470314)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+                return ex.Message switch
+                {
+                    "Cannot find the original signer." => TryLoadCertificateCollection(certificatePath),
+                    "The specified network password is not correct." => ValidationState.Unknown,
+                    _ => ValidationState.Unknown,
+                };
+            }
+            catch (Exception)
+            {
+                return ValidationState.Unknown;
```
michaelcfanning, line 54: >ValidationState.Unknown [](start = 23, length = 23)

see ValidatorBase for our standard helper to generate a return value based on an exception.

CreateReturnValueForException. It looks like that helper might need reworking or we need to create a second helper for this scenario. We shouldn't be returning HostUnknown here, probably just Unknown. We *do* want to emit the public key token to make clear the identity of the thing under analysis.

eddynaka: for the public token, if the file is a bundle, what would be our public token? the concat of all?

---
In reply to: [554470932](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470932) [](ancestors = 554470932)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private static ValidationState TryLoadCertificate(string certificatePath)
+        {
+            X509Certificate2 certificate = null;
+            try
+            {
+                // If this certificate needs a password or it is a bundle, it will throw an exception.
+                certificate = new X509Certificate2(certificatePath);
+
+                return certificate.PrivateKey != null ? ValidationState.Authorized : ValidationState.NoMatch;
```
michaelcfanning, line 41: Sweet! the only problem here is that you're not describing the asset (in this case the public key token of the certificate). The NoMatch return is interesting here! Generally, we drop NoMatch items entirely on the floor. This may confuse people for this case. We might want to consider a new validation state for killed or incomplete or public only credentials. another option would be to raise a NotApplicable message, this check isn't applicable in this case because we're got a public key only result. Stick with NoMatch right now, but let's file a tracking issue.

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: Another question: today we have the enum "Authorized", but this validator should return something like "Match" or something related. Don't know

---
In reply to: [554471205](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554471205) [](ancestors = 554471205,554470932)

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveFiles/ExpectedOutputs/without_password.sarif
```diff
@@ -0,0 +1,69 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
```
michaelcfanning, line 2: >$schema [](start = 3, length = 7)

Let's please encode the rule id in our test files to make things clear, so this file should be SEC1003.PfxCryptographicKeyfile_without_password.sarif. Yes? moving forward, need to push this standard consistently across all code bases.

michaelcfanning:
Where are your test inputs? Remember depending on configuration you may be required to add a force arg to git to get it to submit binary files...

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveFiles/ExpectedOutputs/without_password.sarif
eddynaka: ok. Will do that in the following PRs

---
In reply to: [554471425](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554471425) [](ancestors = 554471425)

eddynaka:
pushed. didnt see that it wasnt there.

---
In reply to: [757365389](https://github.com/microsoft/sarif-pattern-matcher/pull/96#issuecomment-757365389) [](ancestors = 757365389)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: changed to [thumbprint=value]

---
In reply to: [554470631](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554470631) [](ancestors = 554470631,554470486)

Src/Plugins/Security/PFXCryptographicKeyFileValidator.cs
eddynaka: added a few static methods in validatorbase which has a file in the name, so we can distinguish between content vs file

---
In reply to: [554471302](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554471302) [](ancestors = 554471302,554471205,554470932)

Src/Plugins/ValidatorBase.cs
```diff
@@ -81,7 +93,7 @@ public static string CreateReturnValueForCompromisedAsset(string asset, string u
         /// set to null if no fingerprint can be computed that definitively identifies the secret.
         /// </param>
         /// <returns>Return the validation state.</returns>
-        protected abstract string IsValidHelper(
```
michaelcfanning, line 84: >IsValidHelper [](start = 34, length = 13)

You can't change this, please revert. I will show you the pattern to use this base class.

Src/Plugins/ValidatorBase.cs
```diff
@@ -56,6 +56,18 @@ public static string CreateReturnValueForCompromisedAsset(string asset, string u
                 nameof(ValidationState.Authorized) + $"#The '{user}' account is compromised for '{asset}'.";
         }
 
+        public static string CreateReturnValueForFileException(Exception e, string asset)
```
michaelcfanning, line 59: >CreateReturnValueForFileException [](start = 29, length = 33)

This is not a good helper, because it knows about a very specific exception message. That kind of logic should always be in a rule. i.e., you should have a handler that catches the specific kind of exception that has this message.

eddynaka: reverted

---
In reply to: [554483316](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554483316) [](ancestors = 554483316)

Src/Plugins/ValidatorBase.cs
```diff
@@ -56,6 +56,18 @@ public static string CreateReturnValueForCompromisedAsset(string asset, string u
...
+        {
+            if (e.Message.Equals("The specified network password is not correct."))
+            {
+                return nameof(ValidationState.Unknown) + $"#The file '{asset}' is secure and we could not validate.";
+            }
+
+            return nameof(ValidationState.Unknown) +
```
michaelcfanning, line 66: >ValidationState [](start = 26, length = 15)

This helper should only contain this code and the helper should be called CreateReturnValueForUnhandledException

Src/Plugins/ValidatorBase.cs
eddynaka: reverted this.

---
In reply to: [554482975](https://github.com/microsoft/sarif-pattern-matcher/pull/96#discussion_r554482975) [](ancestors = 554482975)

michaelcfanning:
>         public static string CreateReturnValueForException(Exception e, string asset)

This is a bad method name, I will fix later.

---
Refers to: Src/Plugins/ValidatorBase.cs:22 in 56940b5. [](commit_id = 56940b58606e3975f0d7331184d5627b8789c1d3, deletion_comment = False)

michaelcfanning (approved):
:shipit:

PR #107, opened 2021-01-14
Correct rule ids (make them opaque). Provide actual readable names. P…

michaelcfanning:
…lumb everything through. @eddynaka


eddynaka (approved):
:shipit:

PR #120, opened 2021-01-18
Adding e-mail fingerprint

eddynaka:


Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102.EmailAddress_with_email_addresses.sarif
```diff
@@ -835,7 +907,10 @@
                 }
               }
             }
-          ]
+          ],
+          "fingerprints": {
+            "ValidationFingerprint/v1": "[acct=addressPrecededByColonAndFollowedByPeriod][host=fake.com]"
```
michaelcfanning, line 912: >fake.com [](start = 95, length = 8)

FYI, we should always prefer 'example.com'. fake.com is an actual domain! specializing in fake plants. next time we're in this code, we can update.

michaelcfanning (approved):
:shipit:

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102.EmailAddress_with_email_addresses.sarif
eddynaka: fixed

---
In reply to: [559791510](https://github.com/microsoft/sarif-pattern-matcher/pull/120#discussion_r559791510) [](ancestors = 559791510)

PR #125, opened 2021-01-21
Enable NetAnalyzers

eddynaka:
## Changes
- Enable NETAnalyzers
- Fix a few error / warning / messages

## Why
- From our security rules, we should use Roslyn analyzers in IDE/build


jameswinkler (approved):
:shipit:

PR #128, opened 2021-01-24
Aws credentials

michaelcfanning:
@eddynaka


Src/Plugins/Security/AwsCredentialsValidator.cs
```diff
@@ -0,0 +1,147 @@
...
+                message = $"An unexpected exception was caught attempting to authenticate AWS id '{id}': {e.Message}";
+                return nameof(ValidationState.Unknown);
+            }
+            catch (Exception e)
+            {
+                message = $"An unexpected exception was caught attempting to authentic AWS id '{id}': {e.Message}";
+                return nameof(ValidationState.Unknown);
+            }
+
+/*          var client = new HttpClient();
```
michaelcfanning, line 106: We should be able to replace use of the AWS SDK with direct REST API calls, just didn't do sufficient research to make it happen. Relevant code is at [https://github.com/aws/aws-sdk-net/](https://github.com/aws/aws-sdk-net/), specifically, the code in t\sdk\src\Core\Amazon.Runtime\Internal\Auth\AWS4Signer.cs

eddynaka (approved):

PR #133, opened 2021-01-25
Rule updates

michaelcfanning:
@eddynaka 
@jameswinkler

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -215,7 +215,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
                 {
                     if (context != null)
                     {
-                        // An illegal state '{0}' was returned validating a result for check '{1}'.
+                        // An illegal state was returned running check '{0}' against '{1}' ({2}).
                         context.Logger.LogToolNotification(
                             Errors.CreateNotification(
                                 context.TargetUri,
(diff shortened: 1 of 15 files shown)


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -427,9 +428,10 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
                                 FailureLevel.Error,
                                 exception: null,
                                 persistExceptionStack: false,
-                                messageFormat: SpamResources.ERR998_ValidatorReturnedIllegalValidationState,
```
michaelcfanning, line 430: >ERR998_ValidatorReturnedIllegalValidationState [](start = 61, length = 46)

Oops, we used the wrong format string here and also failed to report args properly. For this error and also the illegal return value.

Src/Sarif.PatternMatcher/Validation.cs
```diff
@@ -51,7 +51,7 @@ public enum Validation
         /// <summary>
         /// The host is unknown.
         /// </summary>
-        HostUnknown,
+        UnknownHost,
```
michaelcfanning, line 54: >UnknownHost [](start = 8, length = 11)

Oops, changed the plug in enum and didn't adjust this one. I'll add a unit test for this next.

eddynaka (approved):
:shipit:

PR #137, opened 2021-01-27
Experiments with various OAUTH client id/secret pairs.

michaelcfanning:
@eddynaka, let's be sure to have a talk about this one before merge. :)


eddynaka:
>                         SetProperty(currentKey, value);

u could add trim here and remove from the places that u add

---
Refers to: Src/Sarif.PatternMatcher/Fingerprint.cs:242 in 918b34d. [](commit_id = 918b34d48db4c2b8b95fb089132c48695e4790f9, deletion_comment = False)

eddynaka (approved):

PR #143, opened 2021-01-30
Reading assemblies before loading

eddynaka:


Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -297,6 +297,16 @@ private Assembly CurrentDomain_AssemblyResolve(object sender, ResolveEventArgs a
 
             this._resolvedNames.Add(assemblyName);
 
+            AppDomain currentDomain = AppDomain.CurrentDomain;
+            Assembly[] assemblies = currentDomain.GetAssemblies();
+            foreach (Assembly assembly in assemblies)
+            {
+                if (assembly.FullName.Split(',')[0] == assemblyName)
```
eddynaka, line 304: using the full name didn't work. so using the name of the assembly only

PR #159, opened 2021-02-03
Make PostgreSQL validator order-insensitive

jameswinkler:
Make many corrections to regex, test data, and processing after a match.  Also add more test data and delete unused files


eddynaka:
>         protected override string IsValidDynamicHelper(ref string fingerprintText,

remember to do two tries:
1. with database
2. if the result is unknown from (1), do without

---
Refers to: Src/Plugins/Security/PostgreSqlConnectionStringValidator.cs:82 in a679146. [](commit_id = a679146449543b8ce2ccb0bb3bf661c82afab2ff, deletion_comment = False)

Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -207,6 +207,13 @@
...
           "ContentsRegex": "$MySqlConnectionStringAdo",
           "MessageArguments": { "secretKind": "ADO or ODBC MySQL connection string" }
+        },
+        {
+          "Id": "SEC101/045",
+          "Name": "DoNotExposePlaintextSecrets/PostgreSqlConnectionString",
+          "SubId": "Ado",
+          "ContentsRegex": "$PostgreSqlConnectionStringAdo",
+          "MessageArguments": { "secretKind": "ADO PostgreSQL connection string" }
         }
```
eddynaka, line 217: remove this, since a few lines above we have the same

jameswinkler: Done.  This removed a lot of redundancies in the respective sarif file too.

---
In reply to: [569782450](https://github.com/microsoft/sarif-pattern-matcher/pull/159#discussion_r569782450) [](ancestors = 569782450)

jameswinkler:
>         protected override string IsValidDynamicHelper(ref string fingerprintText,

As discussed, we'll do this or something similar in another PR.

---
In reply to: [772856093](https://github.com/microsoft/sarif-pattern-matcher/pull/159#issuecomment-772856093) [](ancestors = 772856093)

---
Refers to: Src/Plugins/Security/PostgreSqlConnectionStringValidator.cs:82 in a679146. [](commit_id = a679146449543b8ce2ccb0bb3bf661c82afab2ff, deletion_comment = False)

eddynaka (approved):

PR #160, opened 2021-02-03
Order-insensitive for SQL connections

eddynaka:


Src/Plugins/Security/SqlConnectionStringValidator.cs
```diff
@@ -54,9 +63,29 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
...
+            }
+            else
+            {
+                host = ParseExpression(RegexEngine, matchedPattern, HostExpression);
+                database = ParseExpression(RegexEngine, matchedPattern, DatabaseExpression);
+                account = ParseExpression(RegexEngine, matchedPattern, AccountExpression);
+                password = ParseExpression(RegexEngine, matchedPattern, PasswordExpression);
+            }
+
+            if (string.IsNullOrEmpty(host)
```
jameswinkler, line 85: >IsNullOrEmpty [](start = 23, length = 13)

Can you make these "IsNullOrWhitespace" ?  IMO there's pretty much no reason to use IsNullOrEmpty

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_046.SqlConnectionString.ps1
```diff
@@ -1,5 +1,7 @@
 # This is a well-formed but invalid ADO sql connection string
 server=tcp:some-database-name.database.windows.net,1433;Initial Catalog=catalog-db;Persist Security Info=False;User ID=username;Password=password;MultipleActiveResultSets=False;Encrypt=True;TrustServerCertificate=False;Connection Timeout=30;
+Initial Catalog=catalog-db;Persist Security Info=False;Password=password;MultipleActiveResultSets=False;Encrypt=True;User ID=username;TrustServerCertificate=False;Connection Timeout=30;server=tcp:some-database-name.database.windows.net,1433;
```
jameswinkler, line 3: For my sanity, could you create some examples that are wrapped in XML, like we might see in some config files, or perhaps defined in code?  You don't normally see connection strings in just plain text like this.  I want to verify that quotes/new line characters/etc don't have any unexpected effects.

Src/Plugins/ValidatorBase.cs
```diff
@@ -144,6 +146,12 @@ public static string ReturnUnknownHost(ref string message, string host)
             return nameof(ValidationState.Unknown);
         }
 
+        public static string ParseExpression(IRegex regexEngine, string matchedPattern, string expression)
```
jameswinkler, line 149: The two methods added here serve almost the exact same function as the two i added to the PostgeSQL validator.  Could you either consolidate the two, or create a TODO or work item to do so?

michaelcfanning: So, by the way, it turns out that RE2 supports groups, which we were confused about early on (and so didn't plumb the support through). If we exposed this, we wouldn't need to do this parse value stuff. Anyone interested in fixing up RE2 to close this gap? Let me know.

---
In reply to: [569657561](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569657561) [](ancestors = 569657561)

Src/Plugins/ValidatorBase.cs
```diff
@@ -181,6 +189,30 @@ public static string ReturnUnknownHost(ref string message, string host)
...
+            {
+                return value;
+            }
+
+            value = value.Substring(value.IndexOf('=') + 1);
+
+            int index = 0;
+            foreach (char ch in value)
+            {
+                if (ch == ' ' || ch == '\t')
```
michaelcfanning, line 204: >ch [](start = 20, length = 2)

maybe use char.IsWhiteSpace for this?

Src/Plugins/ValidatorBase.cs
```diff
@@ -181,6 +189,30 @@ public static string ReturnUnknownHost(ref string message, string host)
...
+                {
+                    index++;
+                    continue;
+                }
+
+                break;
+            }
+
+            return value.Substring(index);
```
michaelcfanning, line 213: >Substring [](start = 25, length = 9)

should this helper call trim() or otherwise ensure no trailing white space? Also, shouldn't this helper be internal and unit-tested? easy to do.

Src/Plugins/ValidatorBase.cs
eddynaka: where should we fix that? if u could point me i could take a look :)


---
In reply to: [569695240](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569695240) [](ancestors = 569695240,569657561)

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_046.SqlConnectionString.ps1
michaelcfanning: Yes, we have other multiline test patterns that Eddy put together, we should go recreate those. This was a very early rule and the pattern wasn't established yet.

---
In reply to: [569655372](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569655372) [](ancestors = 569655372)

Src/Sarif.PatternMatcher/Fingerprint.cs
```diff
@@ -167,6 +171,11 @@ public override string ToString()
                 components.Add($"[{PersonalAccessTokenKeyName}={this.PersonalAccessToken.Trim()}]");
             }
 
+            if (!string.IsNullOrEmpty(Port))
```
michaelcfanning, line 174: >Port [](start = 38, length = 4)

i wouldn't add this component. instead, I'd work this into the host value, which supports a port value. what do you think?

Src/Sarif.PatternMatcher/Fingerprint.cs
```diff
@@ -127,6 +131,11 @@ public override string ToString()
                 components.Add($"[{AccountKeyName}={this.Account.Trim()}]");
             }
 
+            if (!string.IsNullOrEmpty(Database))
```
michaelcfanning, line 134: >Database [](start = 38, length = 8)

I have wondered previously whether we should call this 'Resource' or something. in other words, this qualifier is for some host- or server-resident thing.

thoughts?

Src/Sarif.PatternMatcher/Fingerprint.cs
eddynaka: for sqlserver that is true, but for the pr james is working (postgres) the port is another field and can be separate from the host.
Server=servername;User Id=username;Port=1234;Password=xyz.
For sqlserver, the host will contain server + port if exists

---
In reply to: [569697684](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569697684) [](ancestors = 569697684)

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -30,7 +30,7 @@
   $HockeyAppToken=(?si)hockey.{0,50}[`'"\s>]?(?<refine>[0-9a-f]{32})([^0-9a-f]|$)
   $MySqlConnectionStringAdo=(?i)Server\s*=(?-i)(?<host>[0-9a-z\-]{3,63}\.mysql\.database\.azure\.com.*)(,|;).*Uid=(?<account>[a-z\@\-]{1,120})(,|;).*Pwd=(?<password>[^;]{8,128});
   $PostgreSqlConnectionStringAdo=(?i)Host\s*=\s*(?-i)(?<host>[0-9a-z\-_\.]{3,91})((?i);Port\s*=\s*(?<port>[0-9]{1,5}))?;Database\s*=\s*(?-i)(?<database>[0-9a-z\-_]{3,91})(?i);Username\s*=\s*(?<account>[^,;]+);Password\s*=\s*(?<password>[^,;\s]+)
-  $SqlConnectionStringAdo=(?i)Server\s*=\s*tcp:(?<host>[^,;]+)(,|;).*(User ID|Uid)\s*=\s*(?<account>[^;]+);.*(Password|Pwd)\s*=\s*(?<password>[^;]+);
-  $SqlConnectionStringJdbc=(?i)jdbc:sqlserver:\/\/(?<host>[^:]+):.*;user\s*=\s*(?<account>[^@]+)@[^;]+;password\s*=\s*(?<password>[^;]+);
-  $SqlConnectionStringPhp=(?i)sqlsrv:server\s*=\s*tcp:(?<host>[^,]+),[^,]+,\s*"(?<account>[^"]+)",\s*"(?<password>[^"]+)"
+  $SqlConnectionStringAdo=(?i)(((server|data source)\s*=\s*[^;]+|(initial catalog|database)\s*=\s*[^;]+|(user id|uid)\s*=\s*[^;]+|(password|pwd)\s*=\s*[^;]+).+?){4}
```
michaelcfanning, line 33: >$SqlConnectionStringAdo [](start = 2, length = 23)

these changes add database but don't solve the order-sensitivity issue (which is the name of this PR). we must be deferring that work? in other words, this regex will trim user data if it happens to follow the password data.

eddynaka: in theory it is working order-insensitive. if u check, i added {4} based on the whole group. With that, it will look for 4 patterns that might contain server/database/user/password. when it finds 4, it will get a match

---
In reply to: [569699776](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569699776) [](ancestors = 569699776)

eddynaka: what i didn't do yet is the same thing for jdbc and php...it's getting trickier since we have only one validator. so, i just fixed to get dabase for now

---
In reply to: [569700938](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569700938) [](ancestors = 569700938,569699776)

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_046.SqlConnectionString.ps1
eddynaka: added more tests and more combinatios, xml style, appsettings style

---
In reply to: [569697107](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569697107) [](ancestors = 569697107,569655372)

Src/Plugins/ValidatorBase.cs
jameswinkler: I'll do it in my branch, disregard!

---
In reply to: [569696512](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569696512) [](ancestors = 569696512,569695240,569657561)

jameswinkler (approved):
:shipit:

Src/Plugins/ValidatorBase.cs
```diff
@@ -144,6 +146,36 @@ public static string ReturnUnknownHost(ref string message, string host)
...
+        internal static string ParseValue(string value)
+        {
+            if (string.IsNullOrEmpty(value))
+            {
+                return value;
+            }
+
+            value = value.Substring(value.IndexOf('=') + 1);
+
+            int index = 0;
```
jameswinkler, line 164: This code seems intended find the index of the first non-whitespace character, so we can use it to make a substring below.  Why not just trim?  Trim hits both sides if I'm not mistaken.  And we use it anyway.

eddynaka: removed that part and added trim only

---
In reply to: [569751725](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569751725) [](ancestors = 569751725)

Src/Sarif.PatternMatcher/Fingerprint.cs
michaelcfanning: My idea is that we consider using the URI authority thing for this. i.e., we could bundle the user and port information with host.

[userinfo@]host[:port]

---
In reply to: [569699026](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569699026) [](ancestors = 569699026,569697684)

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/Fingerprint.cs
eddynaka: @jameswinkler we will have to take a look at this in ur pr.

---
In reply to: [569768638](https://github.com/microsoft/sarif-pattern-matcher/pull/160#discussion_r569768638) [](ancestors = 569768638,569699026,569697684)

PR #161, opened 2021-02-03
Improving GitHub PAT search

eddynaka:


Src/Plugins/Security/GitHubPatValidator.cs
```diff
@@ -13,10 +15,19 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
     internal class GitHubPatValidator : ValidatorBase
     {
         internal static GitHubPatValidator Instance;
+        internal static IRegex RegexEngine;
+
+        private const string PatExpression = "[0-9a-z]{40}";
```
michaelcfanning, line 20: >PatExpression [](start = 29, length = 13)

this needs to be case-insensitive (?i), right?

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -25,7 +25,7 @@
   $AwsCredentials=(?s)\b(?<id>(A3T[0-9A-Z]|AKIA|AGPA|AROA|AIPA|ANPA|ANVA|ASIA)[0-9A-Z]{16}).+?\b(?i)(?<key>[0-9a-z\/+]{40})
   $FacebookAppCredentials=(?si)(facebook|fb)(.{0,20})?[`'"\s>](?<id>[0-9]{13,17})[^0-9].+?(facebook|fb)(.{0,20})?[`'"\s>](?<key>[0-9a-f]{32})([^0-9a-f]|$)
   $GitHubAppCredentials=(?s)\b(?<id>Iv1.[0-9a-z]{16}).+?\b(?i)(?<key>[0-9a-z]{35,40})([^0-9a-f]|$)
-  $GitHubPat=(?si)github.{0,50}(?<refine>[0-9a-z]{40})([^0-9a-z]|$)
+  $GitHubPat=(?si)github(?<refine>.{0,50}[0-9a-z]{40})([^0-9a-z]|$)
```
michaelcfanning, line 28: >github [](start = 18, length = 6)

I think we should probably include 'github' in the refinement, because it is helpful for understanding the region snippet. your .Contains check in the rule should still work.

michaelcfanning (approved):
:shipit:

Src/Plugins/Security/GitHubPatValidator.cs
eddynaka: lower case only

---
In reply to: [569700983](https://github.com/microsoft/sarif-pattern-matcher/pull/161#discussion_r569700983) [](ancestors = 569700983)

PR #165, opened 2021-02-04
Remove ordering in MySql regex

jameswinkler:
Update validator, sarif, etc.  Port optional.


Src/Plugins/Security/MySqlConnectionStringValidator.cs
```diff
@@ -15,6 +16,12 @@ internal class MySqlConnectionStringValidator : ValidatorBase
         internal static MySqlConnectionStringValidator Instance;
         internal static IRegex RegexEngine;
 
+        private const string HostRegex = "(?i)(Server\\s*=\\s*(?<host>[\\w\\-.]{3,90}))";
```
eddynaka, line 19: >\\ [](start = 58, length = 2)

we could use @"" and remove the double slashes

jameswinkler: Done.

---
In reply to: [570524896](https://github.com/microsoft/sarif-pattern-matcher/pull/165#discussion_r570524896) [](ancestors = 570524896)

eddynaka (approved):

PR #168, opened 2021-02-05
Update regex to not scan entire files

jameswinkler:
Having .* or .+ would scan entire files in search of something that follows the expression.  Put reasonable limits on how far we'll search and update test files


michaelcfanning (approved):
:shipit:

PR #173, opened 2021-02-07
Improving sqlserver issues

eddynaka:


Src/Plugins/Security/SEC101_046.SqlConnectionStringValidator.cs
```diff
@@ -166,7 +166,7 @@ private static string ValidateConnectionString(ref string message, string host,
                         if (match.Success)
                         {
                             message = match.Value;
-                            return nameof(ValidationState.Unknown);
+                            return nameof(ValidationState.Unauthorized);
```
eddynaka, line 169: >return nameof(ValidationState.Unauthorized); [](start = 28, length = 44)

If we get at this point, that means that we are not authorized to make a request, since the IP is restricted. So, changing to `Unauthorized`.

Src/Plugins/ValidatorBase.cs
```diff
@@ -93,6 +93,11 @@ public static string ReturnUnhandledException(ref string message, Exception e, s
                 return ReturnUnknownHost(ref message, asset);
             }
 
+            if (TestExceptionForMessage(e, "The network path was not found", asset))
```
eddynaka, line 96: >The network path was not found [](start = 44, length = 30)

Adding a new condition to return UnknownHost, since this means the same thing.

Src/Plugins/Security/SEC101_046.SqlConnectionStringValidator.cs
michaelcfanning: Right, but we don't know if the credential is bad or not here, so the situation is ambiguous. That is, we could have a valid cred (and are getting blocked by the firewall) or we could have an invalid cred (and are getting blocked by the firewall). So I don't think this is the right change. We haven't determined conclusively that the credential should not be revoked.

---
In reply to: [571507753](https://github.com/microsoft/sarif-pattern-matcher/pull/173#discussion_r571507753) [](ancestors = 571507753)

eddynaka: This was an idea to circuit-break and don't make a second request without database name.
If we have a valid cred and got blocked -> unauthorized
if we have an invalid cred and got blocked -> its unauthorized as well, for example, good server and bad password. 

---
In reply to: [571652630](https://github.com/microsoft/sarif-pattern-matcher/pull/173#discussion_r571652630) [](ancestors = 571652630,571507753)

PR #180, opened 2021-02-08
Cleanup regex, update validator to match

jameswinkler:
Also, a few more results in sarif files


Src/Plugins/Security/SEC101_045.PostgreSqlConnectionStringValidator.cs
```diff
@@ -54,15 +51,12 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
...
             string port = ParseExpression(RegexEngine, matchedPattern, PortRegex);
-            string account = ParseExpression(RegexEngine, matchedPattern, AccountRegex);
-            string password = ParseExpression(RegexEngine, matchedPattern, PasswordRegex);
             string database = ParseExpression(RegexEngine, matchedPattern, DatabaseRegex);
 
-            if (string.IsNullOrWhiteSpace(host) ||
-                string.IsNullOrWhiteSpace(account) ||
-                string.IsNullOrWhiteSpace(password))
+            if (!groups.TryGetValue("host", out string host)
+                || !groups.TryGetValue("account", out string account)
```
michaelcfanning, line 58: >  ||  [](start = 14, length = 5)

Actually we agreed this morning is that our standard is to leave these pipes on the lines above, so that the conditions are left aligned.

jameswinkler: Oh, when you asked me, I was saying that I liked how the pipes lined up, not the first character of the condition.  I'm not married to either format in any case.

---
In reply to: [572382722](https://github.com/microsoft/sarif-pattern-matcher/pull/180#discussion_r572382722) [](ancestors = 572382722)

michaelcfanning (approved):
:shipit:

eddynaka (approved):

PR #185, opened 2021-02-10
Static analysis caching

michaelcfanning:
We will only fire a single result per file, per fingerprint. This helps eliminate noise due to over-aggressive regex matching.

@eddynaka @jameswinkler

### DIFF
--- a/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_045.PostgreSqlConnectionString.sarif
+++ b/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_045.PostgreSqlConnectionString.sarif
@@ -62,18 +62,18 @@
                   "startLine": 4,
                   "startColumn": 45,
                   "endLine": 4,
-                  "endColumn": 115,
+                  "endColumn": 117,
                   "charOffset": 91,
-                  "charLength": 70,
+                  "charLength": 72,
                   "snippet": {
-                    "text": "host=my_host;Port=5555;Database=my_db;Username=my_user;Password=my_pw\""
+                    "text": "host=my_host;Port=5555;Database=my_db;Username=my_user;Password=my_pw_1\""
                   }
                 }
               }
             }
           ],
           "fingerprints": {
-            "ValidationFingerprint/v1": "[acct=my_user][host=my_host][pwd=my_pw][port=5555][resource=my_db]"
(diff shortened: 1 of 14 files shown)


Src/Plugins/ValidatorBase.cs
```diff
@@ -41,19 +42,47 @@ protected ValidatorBase()
...
+            string state =
                 validator.IsValidStaticHelper(ref matchedPattern,
                                               ref groups,
                                               ref failureLevel,
                                               ref fingerprint,
                                               ref message);
+
+            if (state != nameof(ValidationState.NoMatch))
+            {
+                string scanTarget = groups["scanTargetFullPath"];
```
michaelcfanning, line 74: >groups [](start = 36, length = 6)

This convention is how we flow the scan target path to the plug-ins. Now, we could have created a new parameter instead. But this identifier is a well-known thing in the scanner format strings, so I thought it was fine to preserve this well-known value in the groups object.

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_006.GitHubPat.ps1
```diff
@@ -5,11 +5,11 @@
 
 # AppSettings style
 <appSettings>
-  <add key="GitHubPat" value="ff34885a8624460a855540c6592698d2f1812843" />
+  <add key="GitHubPat" value="dead885a8624460a855540c6592698d2f1812843" />
```
michaelcfanning, line 8: >dead885a8624460a855540c6592698d2f1812843 [](start = 30, length = 40)

Note that due to this change when you create test cases, each case must produce a unique fingerprint, or some results will be filtered away. :) For all test changes, I've tweaked some portion of the key/secret to make sure each case produces a unique fingerprint.

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_045.PostgreSqlConnectionString.sarif
```diff
@@ -296,57 +296,18 @@
...
-                }
-              }
-            }
-          ],
-          "fingerprints": {
-            "ValidationFingerprint/v1": "[acct=my_user][host=my_host][pwd=my_pw][port=5555]"
-          }
-        },
-        {
-          "ruleId": "SEC101/045",
```
michaelcfanning, line 314: >SEC101 [](start = 21, length = 6)

the missing results in all SARIF files is the outcome of our change! i.e., we were produce multiple results for what was effectively the same test case. :)

eddynaka (approved):

PR #187, opened 2021-02-11
Reverting last sql change

eddynaka:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -29,7 +29,7 @@
   $GoogleApiKey=\b(?<refine>AIza(?i)[0-9a-z-_]{35})([^0-9a-z-_]|$)
   $HockeyAppToken=(?si)hockey.{0,50}[`'"\s>]?(?<refine>[0-9a-f]{32})([^0-9a-f]|$)
   $MySqlConnectionStringAdo=(?i)(Port\s*=\s*([0-9]{4,5}).*)?(((Server\s*=\s*(?<host>[\w\-.]{3,90}))|(Uid=(?-i)(?<account>[a-z\@\-]{1,120})(?i))|(Pwd=(?<password>[^;]{8,128}))).*?){3}(.*Port\s*=\s*([0-9]{4,5}))?
-  $SqlConnectionStringAdo=(?i)(((server|data source)\s*=\s*(?<host>[^;]{0,128})|(initial catalog|database)\s*=\s*(?<database>[^;|"']{0,128})|(user id|uid)\s*=\s*(?<account>[^;|"']{0,128})|(password|pwd)\s*=\s*(?<password>[^;|"']{1,128})).{1,128}?){4}
+  $SqlConnectionStringAdo=(?i)(((server|data source)\s*=\s*[^;]{0,50}|(initial catalog|database)\s*=\s*[^;]{0,50}|(user id|uid)\s*=\s*[^;]{0,50}|(password|pwd)\s*=\s*[^;]{0,50}).+?){4}
```
michaelcfanning, line 32: >{4} [](start = 181, length = 3)

are you sure it is this version of the SQL rule that is ok? it is still performing the order-insensitive match. Don't we need to go back further to the fixed order version of the rule? or not?

eddynaka: this was the change that we had in package  59 (the latest run from chris). So, at least we will be back to the same quantity of results.
We will have to check later.
If we increase 0,50 to 0,128 the regex wont end. So, we have multiple issues in the regex.

---
In reply to: [574853592](https://github.com/microsoft/sarif-pattern-matcher/pull/187#discussion_r574853592) [](ancestors = 574853592)

michaelcfanning (approved):
:shipit:

PR #210, opened 2021-02-15
Mailgun detection and validator (for new style API keys).

michaelcfanning:
@eddynaka

### DIFF
--- a/Src/Plugins/Security/SEC101_026.MailgunApiKeyValidator.cs
+++ b/Src/Plugins/Security/SEC101_026.MailgunApiKeyValidator.cs
@@ -102,14 +102,15 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 
                     default:
                     {
-                        message = $"An unexpected response was returned : '{response.StatusCode}'";
+                        message = $"An unexpected response code was returned attempting to "+
+                                  $"validate the '{account}' account: '{response.StatusCode}'";
                         break;
                     }
                 }
             }
             catch (Exception e)
             {
-                return ReturnUnhandledException(ref message, e);
+                return ReturnUnhandledException(ref message, e, asset: account);
             }
 
             return ReturnUnknownAuthorization(ref message, asset: account);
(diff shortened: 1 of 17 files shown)


Src/Plugins/Security/SEC101_026.MailgunApiKeyValidator.cs
```diff
@@ -0,0 +1,118 @@
...
+                        break;
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                return ReturnUnhandledException(ref message, e);
+            }
+
+            return ReturnUnknownAuthorization(ref message, asset: account);
```
eddynaka, line 115: >return ReturnUnknownAuthorization(ref message, asset: account); [](start = 12, length = 63)

if the status code is NotFound, for example. We would change the message in line 105. but, when we get  here, we would change again.
With that, i would move this and return nameof(ValidationState.Unknown) in line 106 instead of the break.

Src/Plugins/Security/Security.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="GoogleApi" Version="3.10.5" />
     <PackageReference Include="MySqlConnector" Version="1.2.1" />
     <PackageReference Include="Npgsql" Version="5.0.3" />
+    <PackageReference Include="RestSharp" Version="106.11.7" />
```
eddynaka, line 37: ><PackageReference Include="RestSharp" Version="106.11.7" /> [](start = 4, length = 59)

Can't we replace restsharp with  httpclient?

michaelcfanning: I did! We just need to zap this reference, it isn't used. Sorry!

---
In reply to: [575900109](https://github.com/microsoft/sarif-pattern-matcher/pull/210#discussion_r575900109) [](ancestors = 575900109)

Src/Plugins/Security/SEC101_026.MailgunApiKeyValidator.cs
michaelcfanning: we should return unknown at 106 and make this an exceptional path (as a non-feasible code path) or do what I did, which is to construct a message on line 105 and now simply return unknown on line 115. what do you think?

---
In reply to: [575899988](https://github.com/microsoft/sarif-pattern-matcher/pull/210#discussion_r575899988) [](ancestors = 575899988)

eddynaka (approved):
:shipit:

PR #213, opened 2021-02-15
Improvement in Postgres regex

eddynaka:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -34,7 +34,7 @@
   $SEC101/023.AmazonMwsAuthToken=\b(?<refine>amzn\\.mws\\.[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})([^0-9a-f]|$)
   $SEC101/026.MailgunApiKey=(?is)((^|[^0-9a-z\-])(?<key>[0-9a-z]{32}-[0-9a-z]{8}-[0-9a-z]{8}).{0,50}?|[^0-9a-z](?<account>[0-9a-z]+?\.mailgun\.org).{0,50}?){2}
   $SEC101/031.NuGetApiKey=\b(?<refine>(?<key>oy2[0-9a-z]{43}))(^[0-9a-z]|$)
-  $SEC101/045.PostgreSqlConnectionStringAdo=((?i)(((Port\s*=\s*[0-9]{1,5})|((database|db)\s*=\s*[^;]+));.*)?((host|server)\s*=\s*(?<host>[\w\-_\.]{3,91})|(username|uid|user id)\s*=\s*(?<account>[^,;\"\s]+)|(password|pwd)\s*=\s*(?<password>[^,;\"\s]+)).+?){3}(.*((Port\s*=\s*[0-9]{1,5})|((database|db)\s*=\s*[^;]+)))?
+  $SEC101/045.PostgreSqlConnectionStringAdo=((?i)(((Port\s*=\s*[0-9]{1,5})|((database|db)\s*=\s*[^;"<]+));.*)?((host|server)\s*=\s*(?<host>[\w\-_\.]{3,91})|(username|uid|user id)\s*=\s*(?<account>[^,;"<\s]+)|(password|pwd)\s*=\s*(?<password>[^,;"<\s]+)).+?){3}(.*((Port\s*=\s*[0-9]{1,5})|((database|db)\s*=\s*[^;"<]+)))?
```
michaelcfanning, line 37: >*[^;"<]+));.*)?(( [](start = 95, length = 17)

you didn't include the XML pattern in your other change. This breaks us for passwords with these characters. I think we probably need patterns (restricted by file type) that are specific to XML, file types that use double quotes or apostrophes for encapsulation, etc. We should developer a clear pattern, this condition is pervasive. Right now, we hack around on an ad hoc basis. :) Maybe James could work on a master pattern for all rules to follow?

eddynaka: yeah, i only added cases where i saw in the results that we got so far.

---
In reply to: [576400277](https://github.com/microsoft/sarif-pattern-matcher/pull/213#discussion_r576400277) [](ancestors = 576400277)

PR #217, opened 2021-02-15
Unicode escaping

michaelcfanning:
@eddynaka 

Some unicode characters are escaped by URL.LocalPath. This breaks file reads. We can't simply URL decode these values, as some apparently URL decoded file names are actually valid on disk. The solution is to simply use the OriginalString provided to us originally by the file system.


eddynaka (approved):
:shipit:

PR #219, opened 2021-02-16
Consolidate LinkedInClientID with LinkedInSecretKey

eddynaka:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -31,6 +31,7 @@
   $SEC101/006.GitHubPat=(?si)(?<refine>github.{0,50}[^0-9a-z][0-9a-z]{40})([^0-9a-z]|$)
   $SEC101/007.GitHubAppCredentials=(?s)((\b(?<id>Iv1\.[0-9a-z]{16}).{1,50}?)|(\b(?i)(?<key>[0-9a-z]{35,40})\b.{0,50}?)){2}
   $SEC101/008.AwsCredentials=(?s)\b(?<id>(A3T[0-9A-Z]|AKIA|AGPA|AROA|AIPA|ANPA|ANVA|ASIA)[0-9A-Z]{16}).{1,500}?\b(?i)(?<key>[0-9a-z\/+]{40})
+  $SEC101/009.LinkedInCredentials=(?si)linkedin(.{0,50})?\b(?-i)(?<id>[0-9a-z]{14})\b.{0,50}?\b(?i)(?<key>[0-9a-z]{16})\b
```
michaelcfanning, line 34: >\b [](start = 119, length = 2)

I don't think \b will work for a line-break, just for your information.

and so if the key happens to run up against the very end of a line, it won't match. you might want to test this, but i think it's what i observed, \b will match ^ but not $ in our engine.

michaelcfanning: This check potentially looks noisy. In general, I'm not sure of the immediate value of producing client id and secret pairs relative to other work. Maybe clarify offline?

---
In reply to: [576997304](https://github.com/microsoft/sarif-pattern-matcher/pull/219#discussion_r576997304) [](ancestors = 576997304)

michaelcfanning (approved):
:shipit:

PR #220, opened 2021-02-16
Improvement in HttpAuthorizationRequestHeader validation

eddynaka:
This PR fixes the following issues:
- example.com and example.com/ should return the same fingerprint
- adding a dynamic validator

### DIFF
--- a/Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
+++ b/Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
@@ -1,7 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
 
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 {
@@ -28,6 +32,13 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                                                ref message);
         }
 
+        public static string IsValidDynamic(ref string fingerprint, ref string message)
+        {
+            return ValidatorBase.IsValidDynamic(Instance,
+                                                ref fingerprint,
+                                                ref message);
+        }
+
         protected override string IsValidStaticHelper(ref string matchedPattern,
                                                       ref Dictionary<string, string> groups,
                                                       ref string failureLevel,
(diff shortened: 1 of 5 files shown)


Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -44,10 +55,63 @@ static HttpAuthorizationRequestHeaderValidator()
...
+            var fingerprint = new Fingerprint(fingerprintText);
+            string host = fingerprint.Host;
+
+            try
+            {
+                using HttpClient client = CreateHttpClient();
+                client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic", Guid.NewGuid().ToString());
+
+                using HttpResponseMessage responseDummy = client
+                    .GetAsync(host + (fingerprint.Resource ?? string.Empty), HttpCompletionOption.ResponseHeadersRead)
```
michaelcfanning, line 78: >fingerprint.Resource [](start = 38, length = 20)

please lift this into its own variable. generally, it's helpful to show early in the validator what data from the fingerprint is used in the method. this approach also helps lead to more concise lines of code, as a variable named 'resource' is fewer columns than 'fingerprint.Resource'. Note that using a variable will allow you to handle the conditional assignment to string.empty as well, also shortening code. :)

michaelcfanning:
>         public static string ReturnUnauthorizedAccess(ref string message,

Can we also include a regex fix in this change? Or are you doing that next?

---
Refers to: Src/Plugins/ValidatorBase.cs:166 in a521770. [](commit_id = a521770bf29bc2ca5b1032d5bb374141813eb6ee, deletion_comment = False)

michaelcfanning (approved):
:shipit:

eddynaka:
>         public static string ReturnUnauthorizedAccess(ref string message,

i was trying. but, my changes was always making catastrophic issues...will probably need some help to understand.

---
In reply to: [780101549](https://github.com/microsoft/sarif-pattern-matcher/pull/220#issuecomment-780101549) [](ancestors = 780101549)

---
Refers to: Src/Plugins/ValidatorBase.cs:166 in a521770. [](commit_id = a521770bf29bc2ca5b1032d5bb374141813eb6ee, deletion_comment = False)

Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
eddynaka: changed the code. Let me know what do you think :)

---
In reply to: [577118840](https://github.com/microsoft/sarif-pattern-matcher/pull/220#discussion_r577118840) [](ancestors = 577118840)

eddynaka:
>         public static string ReturnUnauthorizedAccess(ref string message,

replaced \s for \n. that fixed the catastrophic error.

---
In reply to: [780102582](https://github.com/microsoft/sarif-pattern-matcher/pull/220#issuecomment-780102582) [](ancestors = 780102582,780101549)

---
Refers to: Src/Plugins/ValidatorBase.cs:166 in a521770. [](commit_id = a521770bf29bc2ca5b1032d5bb374141813eb6ee, deletion_comment = False)

PR #221, opened 2021-02-16
Add region to fingerprint

jameswinkler:


michaelcfanning (changes requested):
:clock1:

Src/Sarif.PatternMatcher/Fingerprint.cs
```diff
@@ -171,6 +175,11 @@ public override string ToString()
                 components.Add($"[{PortKeyName}={this.Port.Trim()}]");
             }
 
+            if (!string.IsNullOrEmpty(Region))
```
michaelcfanning, line 178: >Region [](start = 38, length = 6)

Is this data center region? What's the driving scenario for this change? I don't see rule changes coming in with it.

jameswinkler: There was a PR coming in the private repo that was going to utilize "region" but Eddy twisted my arm into putting that information into the "resource" field instead.  So I cancelled/closed this PR.

---
In reply to: [577213550](https://github.com/microsoft/sarif-pattern-matcher/pull/221#discussion_r577213550) [](ancestors = 577213550)

PR #222, opened 2021-02-16
enable net472 support

eddynaka:
## Changes
- with this pr we will enable net472 support.

## Reason:
- for the previous package if we added to a netframework project, it wasn't moving the files correctly to the output folder


michaelcfanning (approved):
:shipit:

PR #225, opened 2021-02-17
Shannon entropy

michaelcfanning:
Add utility for generating a normalized specific shannon entropy metric for a given string of a given total symbol set.

### DIFF
--- a/Src/Plugins/ValidatorBase.cs
+++ b/Src/Plugins/ValidatorBase.cs
@@ -177,7 +177,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins
             foreach (char ch in charCounts.Keys)
             {
                 double count = charCounts[ch];
-                double frequency = count / (double)input.Length;
+                double frequency = count / input.Length;
                 entropy += -(frequency * Math.Log(frequency, countOfPossibleSymbols));
             }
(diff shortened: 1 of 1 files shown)


Src/Plugins/ValidatorBase.cs
```diff
@@ -157,6 +157,33 @@ public static string ReturnUnhandledException(ref string message, Exception e, s
...
+            foreach (char ch in input)
+            {
+                charCounts.TryGetValue(ch, out double count);
+                charCounts[ch] = ++count;
+            }
+
+            foreach (char ch in charCounts.Keys)
+            {
+                double count = charCounts[ch];
+                double frequency = count / (double)input.Length;
```
eddynaka, line 180: >double [](start = 44, length = 6)

i think we dont need this cast

v-luisangell (approved):

Src/Plugins/ValidatorBase.cs
michaelcfanning: Yes, that's true. Fixed.

---
In reply to: [577963751](https://github.com/microsoft/sarif-pattern-matcher/pull/225#discussion_r577963751) [](ancestors = 577963751)

eddynaka (approved):

PR #226, opened 2021-02-17
Add constant to validator base

jameswinkler:


v-luisangell (approved):

Src/Plugins/ValidatorBase.cs
```diff
@@ -198,6 +198,11 @@ public static string ReturnUnknownHost(ref string message, string host)
             return nameof(ValidationState.Unknown);
         }
 
+        public static string ReturnUnexpectedResponseCode(int statusCode)
+        {
+            return string.Format("An unexpected response was returned : '{0}'", statusCode);
```
michaelcfanning, line 203: >Format [](start = 26, length = 6)

This isn't right, we don't want to return an integer value, we want the readable enum value.

michaelcfanning: I'll add this in my current change, will update your helper.

---
In reply to: [578064418](https://github.com/microsoft/sarif-pattern-matcher/pull/226#discussion_r578064418) [](ancestors = 578064418)

michaelcfanning (changes requested):
:clock1:

Src/Plugins/ValidatorBase.cs
jameswinkler: I've updated it to accept an HttpsStatusCode enum instead of an int, is that what you had in mind?

---
In reply to: [578064596](https://github.com/microsoft/sarif-pattern-matcher/pull/226#discussion_r578064596) [](ancestors = 578064596,578064418)

eddynaka: Michael pushed a pr that contains something similar:
```
public static string CreateUnexpectedResponseCodeMessage(HttpStatusCode status)
        {
            return $"An unexpected HTTP response code was received: '{status}'.";
        }
```

---
In reply to: [578066926](https://github.com/microsoft/sarif-pattern-matcher/pull/226#discussion_r578066926) [](ancestors = 578066926,578064596,578064418)

PR #229, opened 2021-02-18
Improving GoogleApiKeyValidator

### LINKED ISSUE #227: Analyzing Google API keys with referer/APK SHA1 cert restrictions should return pass results
Results such as the following aren't necessarily warnings, they may reflect proper security applied to an API key. Generally, it's best not to expose a key but this can be difficult in some web site scenarios or when embedding a key in a mobile app (even in this latter case, though, Google recommends encrypting or obfuscating).

For these use cases, an API key should be restricted, for example, by applying a referer filter or enforcing a check against the mobile app identity. When our validator detects these conditions today, we fire warnings but arguably, these are 'pass' conditions! :)

e:\repros\test.txt(3,1-40): warning SEC101/003: 'test.txt' contains an apparent Google API key, the validity of which could not be determined by runtime analysis (an unexpected exception was caught attempting to validate api key: RequestDenied: API keys with referer restrictions cannot be used with this API).

e:\repros\test.txt(6,1-40): warning SEC101/003: 'test.txt' contains an apparent Google API key, the validity of which could not be determined by runtime analysis (an unexpected exception was caught attempting to validate api key: RequestDenied: This IP, site or mobile application is not authorized to use this API key. Request received from IP address 2601:600:877f:8c60:3033:d967:d038:6ee3, with empty referer).

eddynaka:
Fixes #227

### DIFF
--- a/Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
+++ b/Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
@@ -83,8 +83,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             }
             catch (Exception e)
             {
-                var googleException = e as GoogleApiException;
-                if (googleException != null)
+                if (e is GoogleApiException googleException)
                 {
                     switch (e.Message)
                     {
@@ -109,9 +108,13 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                             return nameof(ValidationState.Expired);
                         }
 
-                        case Invalid:
                         case IpNotAuthorized:
                         case RefererRestricted:
+                        {
+                            return nameof(ValidationState.Pass);
+                        }
+
+                        case Invalid:
                         {
                             return nameof(ValidationState.NoMatch);
                         }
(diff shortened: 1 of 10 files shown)


michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
```diff
@@ -61,8 +61,10 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
             const string Deleted = "RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console).";
             const string KeyNotAuthorized = "RequestDenied: This API key is not authorized to use this service or API.";
             const string ProjectNotAuthorized = "RequestDenied: This API project is not authorized to use this API.";
+            const string IpNotAuthorized = "RequestDenied: This IP, site or mobile application is not authorized to use this API key.";
```
michaelcfanning, line 64: >IpNotAuthorized [](start = 25, length = 15)

This is not the right change. 'NoMatch' indicates a false positive. Both these conditions indicate that we *did* in fact find a match and that the key was actually secured in some way. We can't take this change, we can discuss more offline.

eddynaka: I could add Pass to the ValidationState and in the SearchSkimmer, if we get a pass, we change the kind to pass as well. Should we leave error as warning or should we change as well?

---
In reply to: [578522731](https://github.com/microsoft/sarif-pattern-matcher/pull/229#discussion_r578522731) [](ancestors = 578522731)

PR #231, opened 2021-02-19
Validating certificates if they are already loaded

eddynaka:


Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -222,7 +225,7 @@ public ValidatorsCache(IEnumerable<string> validatorBinaryPaths = null, IFileSys
                     try
                     {
                         assemblyBaseFolder = Path.GetDirectoryName(validatorPath);
-                        assembly = _fileSystem.AssemblyLoadFrom(validatorPath);
+                        assembly = Assembly.Load(_fileSystem.FileReadAllBytes(validatorPath));
```
michaelcfanning, line 228: >Assembly.Load [](start = 35, length = 13)

Do we eventually require adding AssemblyLoad to IFileSystem?

michaelcfanning (approved):
:shipit:

PR #241, opened 2021-02-27
Capture a few more scenarios for MySql

jameswinkler:


Src/Plugins/Security/SEC101_047.MySqlConnectionStringValidator.cs
```diff
@@ -62,20 +59,32 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
...
+            // However, the credentials can still be valid, so we should return "unknown".
+            // Grab the empty host here and then short circuit in dynamic validation.
+
+            Fingerprint fingerprint = new Fingerprint()
+            {
+                Account = account,
+                Password = password,
+            };
+
+            if (!groups.TryGetNonEmptyValue("host", out string host))
```
eddynaka, line 78: >groups.TryGetNonEmptyValue("host", out string host) [](start = 17, length = 51)

this is strange. if we  want to do this, i think its better to create a new rule just to grab account and password

michaelcfanning (approved):
:shipit:

PR #245, opened 2021-03-01
Add new PSCredentialsValidator

jameswinkler:


Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -242,6 +242,27 @@
...
+          "Id": "SEC101/048",
+          "Name": "DoNotExposePlaintextSecrets/PSCredential",
+          "SubId": "Constructor",
+          "ContentsRegex": "$SEC101/048.PSCredentialConstructor",
+          "MessageArguments": { "secretKind": "PSCredential constructor" }
+        },
+        {
+          "Id": "SEC101/048",
+          "Name": "DoNotExposePlaintextSecrets/PSCredential",
+          "SubId": "Object Initializer",
```
eddynaka, line 256: >"SubId": "Object Initializer", [](start = 10, length = 30)

we could remove the subid since the rule is almost the same thing, right? and it would work the same way

jameswinkler: You mean by merging the regexes?  I'm trying to be more specific so that they have better performance.

eddynaka: our code handles this without subid. So we can create the same rule without subids.
just remove subid and test :)
let me know if that does not work

---
In reply to: [585748165](https://github.com/microsoft/sarif-pattern-matcher/pull/245#discussion_r585748165) [](ancestors = 585748165)

jameswinkler: Done.

---
In reply to: [585749742](https://github.com/microsoft/sarif-pattern-matcher/pull/245#discussion_r585749742) [](ancestors = 585749742,585748165)

eddynaka (approved):

PR #246, opened 2021-03-02
Adding limit to size of file when analyzing

eddynaka:


Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -35,5 +35,11 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             "deny-regex",
             HelpText = "A regular expression used to suppress scanning for any file or directory path that matches the regex.")]
         public string FileNameDenyRegex { get; internal set; }
+
+        [Option(
+            "file-size",
+            HelpText = "The file max size in kilobytes that will be analyzed.",
```
michaelcfanning, line 41: >The file max size in kilobytes that will be analyzed [](start = 24, length = 52)

Update slightly: The maximum file size (in kilobytes) that will be analyzed.

michaelcfanning (approved):
:shipit:

PR #248, opened 2021-03-02
Correcting Ids

jameswinkler:


Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -250,13 +250,13 @@
           "MessageArguments": { "secretKind": "PSCredential constructor" }
         },
         {
-          "Id": "SEC101/048",
+          "Id": "SEC101/049",
```
eddynaka, line 253: >049 [](start = 24, length = 3)

build is stuck. can u take a look?

michaelcfanning: If you have any test data for these checks (and you author good tests!), you probably need to update rule ids in the output SARIF.

---
In reply to: [585883455](https://github.com/microsoft/sarif-pattern-matcher/pull/248#discussion_r585883455) [](ancestors = 585883455)

jameswinkler: Yup, it was the output file.  Whoops.

---
In reply to: [585905492](https://github.com/microsoft/sarif-pattern-matcher/pull/248#discussion_r585905492) [](ancestors = 585905492,585883455)

eddynaka (approved):

PR #250, opened 2021-03-03
Check if all rules have validators

jameswinkler:


Src/Plugins/Tests.Security/ReflectionTests.cs
```diff
@@ -0,0 +1,63 @@
...
+using Newtonsoft.Json.Linq;
+
+using Xunit;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
+{
+    public class ReflectionTests
+    {
+        [Fact]
+        public void All101Rules_ShouldHaveValidators()
```
michaelcfanning, line 20: >All101Rules_ShouldHaveValidators [](start = 20, length = 32)

Nice work, James!

One suggestion, you can simply your assembly retrieval by simply finding a type in that project and grabbing the Assembly from it. This will ensure that binary is in memory in all cases, in case of running this test entirely on its own, for example,

typeof(SendGridKeyValidator).Assembly

jameswinkler: Thanks Michael!

---
In reply to: [586019894](https://github.com/microsoft/sarif-pattern-matcher/pull/250#discussion_r586019894) [](ancestors = 586019894)

michaelcfanning (approved):
:shipit:

michaelcfanning:
love this change, let's ship it!

eddynaka (approved):
:shipit:

PR #257, opened 2021-03-03
Add new PicaticApiKeyValidator

jameswinkler:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -42,6 +42,7 @@
   $SEC101/013/Pem.CryptographicPrivateKey=(?si)-{5}BEGIN (DSA|EC|OPENSSH|PGP|RSA|SSH2 ENCRYPTED)?\s*PRIVATE (KEY BLOCK|KEY)-{5}.*?(VERSION: [^\n]+\n)*\n(?<key>[^:]*?)-{5}END (DSA|EC|OPENSSH|PGP|RSA|SSH2 ENCRYPTED)?\s*PRIVATE (KEY BLOCK|KEY)-{5}
   $SEC101/014.FacebookAccessToken=\b(?<refine>EAACEdEose0cBA(?i)[0-9a-z]+)
   $SEC101/016.StripeApiKey=\b(?<refine>(?<key>(?:r|s)k_(live|test)_(?i)[0-9a-z]{24,99}))([^0-9a-z]|$)
+  $SEC101/019.PicaticApiKey=\b(?<refine>sk_(live|test)_[0-9a-z]{32})
```
michaelcfanning, line 45: >refine [](start = 33, length = 6)

use this: 

(?<refine>(?key>sk_(live|test)_[0-9a-z]{32}))

I know it's weird. What <refine> does is to specify what match pattern is set to. This is the region that will be selected in Visual Studio when you double-click the result. It's the region of special interest. However, in this case, that region also happens to be the key. And it's better if our validators operate against these well-known group names to populate fingerprints rather than knowing that a rule happens to render the key as the matchPattern argument.

My note applies in lots of places!

michaelcfanning (approved):
:shipit:

PR #261, opened 2021-03-04
Drop existing  `warning` levels to `note`. 

michaelcfanning:
And introduce `warning` for some lower value `authorized` states.

@eddynaka @jameswinkler


eddynaka (approved):
:shipit:

PR #268, opened 2021-03-04
Changing failureLevel when NoMatch

eddynaka:


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -226,6 +226,8 @@ public override void Analyze(AnalyzeContext context)
                     // The validator determined the match is a false positive.
                     // i.e., it is not the kind of artifact we're looking for.
                     // We should suspend processing and move to the next match.
+
+                    level = FailureLevel.Note;
```
michaelcfanning, line 230: >Note [](start = 41, length = 4)

This isn't right, no match is not a failure. This should be 'None' if you want an explicit value.

Src/Sarif.PatternMatcher/ValidatingVisitor.cs
```diff
@@ -38,7 +38,7 @@ public override Result VisitResult(Result node)
                 // {0:scanTarget}' contains {1:validationPrefix}{2:encoding}{3:secretKind}{4:validationSuffix}{5:validatorMessage}
                 string message = null;
 
-                FailureLevel level = node.Level;
+                FailureLevel level = default;
```
michaelcfanning, line 41: >level [](start = 29, length = 5)

This is right! and a good change. Seeing this I think I understand the bug, was NoMatch getting the default level??

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -249,7 +251,7 @@ public override void Analyze(AnalyzeContext context)
                                 validatorMessage));
                     }
 
-                    level = FailureLevel.Error;
+                    level = FailureLevel.Note;
```
michaelcfanning, line 254: >Note [](start = 41, length = 4)

This isn't right, 'Error' is correct. We always raise visibility on catastrophic errors. Users will then tell us about them so that we can fix. With your change, this bad engine problem could remain hidden for a long time.

PR #270, opened 2021-03-04
Filename should be unique

eddynaka:


Src/Sarif.PatternMatcher.Function/HttpAnalyzeFunction.cs
```diff
@@ -28,7 +28,7 @@ public static class HttpAnalyzeFunction
 
             try
             {
-                string fileName = "temp.txt";
+                string fileName = $"{Guid.NewGuid()}.txt";
```
michaelcfanning, line 31: > $"{Guid.NewGuid()}.txt" [](start = 33, length = 24)

[Dynamic content from web page was analyzed]:{Guid.NewGuid()}}

michaelcfanning: Would this name be too long? Remember to walk in the shoes of the customer: what do they think when they see a guid? Instead, let's do our best to communicate what's happening here: we analyzed content delivered by the web page, so we don't have a real file name.


---
In reply to: [588495862](https://github.com/microsoft/sarif-pattern-matcher/pull/270#discussion_r588495862) [](ancestors = 588495862)

eddynaka: it is. its a hot fix but we can think later how to improve that. We can't have similar names because if we do that we can have one search as error and the next as nomatch. Reason, we have a cache per path/result causing this.

---
In reply to: [588499993](https://github.com/microsoft/sarif-pattern-matcher/pull/270#discussion_r588499993) [](ancestors = 588499993,588495862)

PR #274, opened 2021-03-08
Enable NPM Validator

eddynaka:


Src/Plugins/Tests.Security/ReflectionTests.cs
```diff
@@ -21,6 +21,11 @@ public void All101Rules_ShouldHaveValidators()
         {
             try
             {
+                if (!File.Exists("SEC101.SecurePlaintextSecrets.json"))
```
eddynaka, line 24: >if (!File.Exists("SEC101.SecurePlaintextSecrets.json")) [](start = 16, length = 55)

sometimes the CI does not find the file.

Src/Plugins/Security/SEC101_017.NpmApiKeyValidator.cs
```diff
@@ -0,0 +1,177 @@
...
+                foreach (Object obj in tokensRoot.Tokens)
+                {
+                    if (!key.Contains(obj.Token))
+                    {
+                        continue;
+                    }
+
+                    if (obj.Readonly)
+                    {
+                        message = "The token has read permissions.";
```
eddynaka, line 126: >token [](start = 39, length = 5)

should we rename the rule to NpmTokenValidator instead of ApiKey?

michaelcfanning: Yes. Also, you might consider using apostrophes for these messages, the token has 'read' permissions. The token has 'automation' permissions. Etc.

---
In reply to: [590425352](https://github.com/microsoft/sarif-pattern-matcher/pull/274#discussion_r590425352) [](ancestors = 590425352)

michaelcfanning: Looks like these are called NPM author tokens? So NpmAuthorToken?

---
In reply to: [590446403](https://github.com/microsoft/sarif-pattern-matcher/pull/274#discussion_r590446403) [](ancestors = 590446403,590425352)

michaelcfanning (approved):
:shipit:

PR #281, opened 2021-03-09
Add google service account key validator dynamic validation and unit tests

jameswinkler:


Src/Plugins/Security/SEC101_053.GoogleServiceAccountKeyValidator.cs
```diff
@@ -0,0 +1,117 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+
+using Google.Apis.Auth.OAuth2;
+
+using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.Utilities;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
```
michaelcfanning, line 12: >Security [](start = 62, length = 8)

Hello! Let's talk offline. I don't think you can fit this class of validation into our model.

PR #288, opened 2021-03-12
Adding argument to disable cache

eddynaka:


Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -41,5 +41,9 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             HelpText = "The maximum file size (in kilobytes) that will be analyzed.",
             Default = 1024)]
         public int FileSizeInKilobytes { get; internal set; }
+
+        [Option(
+            "disable-validation-caching")]
```
michaelcfanning, line 46: >disable-validation-caching [](start = 13, length = 26)

disable-dynamic-validation-caching

DisableDynamicValidationCaching

Src/Plugins/ValidatorBase.cs
```diff
@@ -113,6 +119,11 @@ protected ValidatorBase()
             return validationState;
         }
 
+        public static void DisableValidationCaching(bool disableValidationCaching)
```
michaelcfanning, line 122: >DisableValidationCaching [](start = 27, length = 24)

Just use a static property for this!

Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -277,10 +309,19 @@ public ValidatorsCache(IEnumerable<string> validatorBinaryPaths = null, IFileSys
                             isValidDynamic = null;
                         }
 
-                        ruleToMethodMap[typeName] = new ValidationMethodPair
+                        MethodInfo disableValidationCaching = type.BaseType.GetMethod(
```
michaelcfanning, line 312: >GetMethod [](start = 76, length = 9)

Let's use a property for this. It's fine to stick with 'ValidationMethods' as a name. You can retrieve the setter for the property and store it there.

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/AnalyzeOptions.cs
eddynaka: renamed

---
In reply to: [592834129](https://github.com/microsoft/sarif-pattern-matcher/pull/288#discussion_r592834129) [](ancestors = 592834129)

Src/Plugins/ValidatorBase.cs
eddynaka: changing to a property or field is a problem. We would need the object to change and we don't have it.
I will leave as a method for now and we can try later to change again

---
In reply to: [592834390](https://github.com/microsoft/sarif-pattern-matcher/pull/288#discussion_r592834390) [](ancestors = 592834390)

PR #289, opened 2021-03-12
Fixing fingerprint ordering

eddynaka:
Fix to prevent new ordering issues.

Using the dictionary and saving, we can order by key then use the value.

### DIFF
--- a/Src/Sarif.PatternMatcher/Fingerprint.cs
+++ b/Src/Sarif.PatternMatcher/Fingerprint.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
@@ -123,6 +124,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
         public string GetValidationFingerprintText() => ToString(this, denyList: s_assetOnlyKeys);
 
 #pragma warning disable SA1107 // Code should not contain multiple statements on one line
+
         public void SetProperty(string keyName, string value)
         {
             switch (keyName)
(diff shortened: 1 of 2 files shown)


eddynaka:
>             if (!computedFingerprint.Equals(fingerprintText))

I think we should remove this to prevent issues. What do you think?

---
Refers to: Src/Sarif.PatternMatcher/Fingerprint.cs:71 in d7d8192. [](commit_id = d7d81922cea23272c0d4c8ee328c93738472fdeb, deletion_comment = False)

Src/Sarif.PatternMatcher/Fingerprint.cs
```diff
@@ -157,91 +160,91 @@ internal static string ToString(Fingerprint f, ISet<string> denyList)
...
 
             if (!string.IsNullOrEmpty(f.Uri) && !denyList.Contains(UriKeyName))
             {
-                components.Add($"[{UriKeyName}={f.Uri.Trim()}]");
+                components.Add(UriKeyName, $"[{UriKeyName}={f.Uri.Trim()}]");
             }
 
             return components.Count > 0 ?
-                string.Concat(components) :
+                string.Concat(components.OrderBy(c => c.Key).Select(v => v.Value)) :
```
michaelcfanning, line 247: >components [](start = 30, length = 10)

Good change. We could have incorporated the IsNullOrEmptyCheck here as well?

eddynaka: just added. what about my other question? what do u think?

---
In reply to: [593336660](https://github.com/microsoft/sarif-pattern-matcher/pull/289#discussion_r593336660) [](ancestors = 593336660)

michaelcfanning (approved):
:shipit:

PR #294, opened 2021-03-13
Improving Facebook validator

eddynaka:


Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
```diff
@@ -66,20 +68,77 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
             string id = fingerprint.Id;
             string key = fingerprint.Key;
 
-            using var httpClient = new HttpClient();
+            string result = RetrieveInformation(
```
michaelcfanning, line 71: >result [](start = 19, length = 6)

I'd call this 'state', as its the validation state rendered as a string.

michaelcfanning (approved):
:shipit:

PR #295, opened 2021-03-13
Improving validators

eddynaka:
## Changes
- Improved false positives for httpauthorization
- Fixed wrong response.error usage in slack validator
- Ignoring tcp connections from postgres validator
- improving postgres exception logging (first inner then exception if exists)


michaelcfanning (changes requested):
:clock1:

michaelcfanning (approved):
:shipit:

PR #298, opened 2021-03-14
Update google API key rule. Fix reporting helper. Update windows SDK.

michaelcfanning:
@eddynaka, sorry had to delete my old windows SDK and our current version doesn't appear to be archived! so went ahead and just refreshed to a more recent release. it's not a super lengthy install. sorry for the inconvenience, though.

Here's the downloader. Be sure to uninstall version 18362, it's 2+ GB! :)

https://go.microsoft.com/fwlink/p/?linkid=2120843

### DIFF
--- a/Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
+++ b/Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
@@ -59,7 +59,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                                                        ref string message)
         {
             const string EnableBilling = "RequestDenied: You must enable Billing on the Google Cloud Project at https://console.cloud.google.com/project/_/billing/enable Learn more at https://developers.google.com/maps/gmp-get-started";
-            const string Deleted = "RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console).";
+            const string Deleted = "RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console.";
             const string KeyNotAuthorized = "RequestDenied: This API key is not authorized to use this service or API.";
             const string ProjectNotAuthorized = "RequestDenied: This API project is not authorized to use this API.";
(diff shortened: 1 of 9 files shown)


Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
```diff
@@ -92,6 +92,7 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
                             return nameof(ValidationState.AuthorizedError);
                         }
 
+                        case ProjectNotAuthorized:
```
eddynaka, line 95: >ProjectNotAuthorized [](start = 29, length = 20)

are we sure about this? This caused the previous issue...

Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
```diff
@@ -59,7 +59,7 @@ public static string IsValidDynamic(ref string fingerprint, ref string message)
                                                        ref string message)
         {
             const string EnableBilling = "RequestDenied: You must enable Billing on the Google Cloud Project at https://console.cloud.google.com/project/_/billing/enable Learn more at https://developers.google.com/maps/gmp-get-started";
-            const string Deleted = "RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console).";
+            const string Deleted = "RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console.";
```
eddynaka, line 62: >Deleted [](start = 25, length = 7)

we could try to include this:
RequestDenied: API keys with referer restrictions cannot be used with this API
RequestDenied: This API project was not found. This API project may have been deleted or may not be authorized to use this API. You may need to enable the API under APIs in the console
RequestDenied: Google has disabled the use of APIs from this API project

and also, we would have to use a contains to validate the following (check ip address):
RequestDenied: This IP, site or mobile application is not authorized to use this API key. Request received from IP address 52.161.72.40, with empty referer

also, this rules does not end with period.

eddynaka (changes requested):
:clock1:

PR #305, opened 2021-03-15
FileContent exception handling

eddynaka:


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -151,7 +151,33 @@ public override void Analyze(AnalyzeContext context)
...
+                try
+                {
+                    context.FileContents = _fileSystem.FileReadAllText(filePath);
+                }
+                catch (Exception e)
+                {
+                    if (e is IOException || e is UnauthorizedAccessException)
+                    {
+                        // We should only log here, since when we throw,
+                        // the analyzer will catch the exception and log.
```
michaelcfanning, line 163: // We should log and return here because we want the rule to continue to run. i.e., the issue is likely in permissions with the scan target, not a general problem with the rule. In other cases, we 'throw', which will result in the rule getting disabled.

PR #306, opened 2021-03-15
Enhanced reporting

eddynaka:


Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -43,7 +43,14 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
         [Option(
-            "disable-dynamic-validation-caching")]
+            "disable-dynamic-validation-caching",
+            HelpText = "Disable caching from dynamic validation.")]
         public bool DisableDynamicValidationCaching { get; internal set; }
+
+        [Option(
+            "enhanced-reporting",
+            HelpText = "Enable enhanced reporting. If possible, it will fetch data from " +
+                       "the provider and add more information to the reporting message.")]
```
michaelcfanning, line 53: "Enable enhanced reporting provided by dynamic validators (when --dynamic-validation is also enabled)."

Src/Sarif.PatternMatcher/ValidatingVisitor.cs
```diff
@@ -37,12 +39,14 @@ public override Result VisitResult(Result node)
                 // Our validation messages currently look like so.
                 // {0:scanTarget}' contains {1:validationPrefix}{2:encoding}{3:secretKind}{4:validationSuffix}{5:validatorMessage}
                 string message = null;
+                IDictionary<string, string> options = new Dictionary<string, string>();
```
michaelcfanning, line 42: >IDictionary<string, string> [](start = 16, length = 27)

use 'var'

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/ValidatingVisitor.cs
eddynaka: the method below needs IDictionary, so we need to enforce using the interface

PR #312, opened 2021-03-17
Hash validation fingerprint

eddynaka:


Src/Sarif.PatternMatcher/MatchHasher.cs
```diff
@@ -0,0 +1,46 @@
...
+
+using System;
+using System.Security.Cryptography;
+using System.Text;
+
+using Microsoft.Strings.Interop;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
+{
+    public static class MatchHasher
```
michaelcfanning, line 12: >MatchHasher [](start = 24, length = 11)

Can we move this helper to the Fingerprint class? Thanks! :)

michaelcfanning (approved):
:shipit:

PR #314, opened 2021-03-18
Merge PS and Sql Credential Validators

### LINKED ISSUE #272: Merge SqlCredential and PSCredential
These two validators are almost identical in test cases and code.  The only real difference are the names which can easily be merged with regex.  

Merge the two validators.

jameswinkler:
Fixes #272


eddynaka (approved):
:shipit:

PR #315, opened 2021-03-19
Adding dropbox validators

eddynaka:
## Changes
- adding dropbox access token validation
  - two different token types (no expiration and short expiration)
- adding dropbox key and secret validation


Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
```diff
@@ -0,0 +1,133 @@
...
+                    {
+                        string body = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+
+                        // App deleted.
+                        if (body.Contains("This app is currently disabled"))
+                        {
+                            return nameof(ValidationState.Expired);
+                        }
+
+                        // Request was successfull but AccessToken does not have access.
```
michaelcfanning, line 102: >successfull [](start = 39, length = 11)

s/be 'successful'

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
```diff
@@ -0,0 +1,133 @@
...
+                    {
+                        return nameof(ValidationState.AuthorizedError);
+                    }
+
+                    case HttpStatusCode.BadRequest:
+                    {
+                        string body = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+
+                        // App deleted.
+                        if (body.Contains("This app is currently disabled"))
```
michaelcfanning, line 97: >Contains [](start = 33, length = 8)

Do we have a more comprehensive string we could encode here? Could this be a StartsWith?

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
```diff
@@ -0,0 +1,133 @@
...
+                        {
+                            return nameof(ValidationState.Expired);
+                        }
+
+                        // Request was successfull but AccessToken does not have access.
+                        if (body.Contains("file_requests.read"))
```
michaelcfanning, line 103: >Contains [](start = 33, length = 8)

Same here, this is a pretty obscure token 'file_requests.read', could be encode something more comprehensive? For really long strings that compromise readability, you can create const strings in the class.

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
```diff
@@ -0,0 +1,133 @@
...
+                        break;
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                return ReturnUnhandledException(ref message, e);
+            }
+
+            return nameof(ValidationState.Unknown);
```
michaelcfanning, line 130: >nameof [](start = 19, length = 6)

If you move this to the default case, I think this code path becomes unfeasible and you wouldn't be able to have anything here. this would co-locate this return path with the message construction. probably a clearer pattern to use (I know, of course, that we have preferred the pattern you have here in other rules. just a thought to explore, not a necessary change.

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -45,6 +45,9 @@
   $SEC101/016.StripeApiKey=\b(?<refine>(?<key>(?:r|s)k_(live|test)_(?i)[0-9a-z]{24,99}))([^0-9a-z]|$)
   $SEC101/017.NpmAuthorToken=(?i)npm.{0,100}[^0-9a-z](?-i)(?<refine>(?<key>[a-z0-9]{8}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{4}-[a-z0-9]{12}))[^0-9A-Za-z]
   $SEC101/019.PicaticApiKey=\b(?<refine>sk_(live|test)_[0-9a-z]{32})(?i)([^0-9a-z]|$)
+  $SEC101/020.DropboxAccessTokenNoExpiration=(?si)dropbox.{1,50}[^0-9a-z](?<refine>[0-9a-z_-]{64})([^0-9a-z]|$)
+  $SEC101/020.DropboxAccessTokenShortExpiration=(?si)dropbox.{1,50}\b(?<refine>sl\.(?i)[0-9a-z-_]{135,139})(?i)([^0-9a-z]|$)
```
michaelcfanning, line 49: >DropboxAccessTokenShortExpiration [](start = 14, length = 33)

It'd be nice if we could distingish a secret that never expires from other secrets. Is that possible for us? IOW, it is a more serious concern to flag a key that never expires than one that is fixed in time-frame (in terms of validity).

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/Security.SharedStrings.txt
eddynaka: The only way to do that is checkin the size. no expiration has a fixed size (64) characters. different from that  is short lived.
i could also  make a check and see if its a short lived, return as authorized warning
otherwise, authorized error

---
In reply to: [598009225](https://github.com/microsoft/sarif-pattern-matcher/pull/315#discussion_r598009225) [](ancestors = 598009225)

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
eddynaka: updated to ends with. the first piece of the string shows an error message that can be used in other scenarios

---
In reply to: [598008133](https://github.com/microsoft/sarif-pattern-matcher/pull/315#discussion_r598008133) [](ancestors = 598008133)

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
eddynaka: changed to const format and added a better message

---
In reply to: [598008335](https://github.com/microsoft/sarif-pattern-matcher/pull/315#discussion_r598008335) [](ancestors = 598008335)

Src/Plugins/Security/SEC101_020.DropboxAccessTokenValidator.cs
eddynaka: I didn't do that because there are two places that i use this:
1. default case
2. bad request not specified message.  so, i would leave this for now and think in a  better strategy later :)

---
In reply to: [598008839](https://github.com/microsoft/sarif-pattern-matcher/pull/315#discussion_r598008839) [](ancestors = 598008839)

michaelcfanning (approved):
:shipit:

PR #316, opened 2021-03-19
MongoDb validator

eddynaka:


Src/Plugins/Security/SEC101_050.MongoDbConnectionStringValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            var fingerprint = new Fingerprint(fingerprintText, false);
+
+            string host = fingerprint.Host;
+            string account = fingerprint.Account;
+            string password = fingerprint.Password;
+
+            try
+            {
+                var dbClient = new MongoClient($"mongodb+srv://{account}:{password}@{host}/?connectTimeoutMS=3000");
```
michaelcfanning, line 80: >dbClient [](start = 20, length = 8)

Just curious, are you sure this type isn't disposable? thanks for adding a timeout!

Src/Plugins/Security/SEC101_050.MongoDbConnectionStringValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+                dbClient.ListDatabases();
```
michaelcfanning, line 81: >dbClient [](start = 16, length = 8)

Likely, there's some interesting data at this level we could put in the result string. 'The following databases are compromised: DB1, DB2, DB3'. Or other things. Might be worth looking around just adding the db names as I suggest.

Src/Plugins/Security/SEC101_050.MongoDbConnectionStringValidator.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+            catch (Exception e)
+            {
+                if (e is MongoAuthenticationException mae)
+                {
+                    if (e.Message.Contains("Unable to authenticate"))
```
michaelcfanning, line 87: >Contains [](start = 34, length = 8)

Same note, if possible, prefer Equals comparisons to exact strings. If there's uncertainty (like when we see some strings with a period at the end and others not), StartsWith is useful. Contains is always the most permissive check. Now, you might think that's valuable here and want to make a case for it. I don't know. :)

michaelcfanning (approved):
:shipit:

Src/Plugins/Security/SEC101_050.MongoDbConnectionStringValidator.cs
eddynaka: done

---
In reply to: [598010493](https://github.com/microsoft/sarif-pattern-matcher/pull/316#discussion_r598010493) [](ancestors = 598010493)

Src/Plugins/Security/SEC101_050.MongoDbConnectionStringValidator.cs
eddynaka: changed to startswith

---
In reply to: [598010958](https://github.com/microsoft/sarif-pattern-matcher/pull/316#discussion_r598010958) [](ancestors = 598010958)

PR #317, opened 2021-03-19
Add new Cmdkey validator

jameswinkler:


Src/Plugins/Security/SEC101_055.CmdkeyValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                return true;
+            }
+
+            if (potentialPassword.Length < 2)
+            {
+                return true;
+            }
+
+            if (potentialPassword.First() == '$')
```
michaelcfanning, line 82: >potentialPassword [](start = 16, length = 17)

Just index into the string for this, potentialPassword[0]. But really, do we want this? Couldn't this character exist in a BAT or CMD file? Not for PS of course.

Src/Plugins/Security/SEC101_055.CmdkeyValidator.cs
```diff
@@ -0,0 +1,116 @@
...
+                Account = account,
+                Password = password,
+            }.ToString();
+
+            // Manual use of the regex against CodeAsData produces highly confident results,
+            // so return AuthorizedWarning rather than Unknown.
+            return nameof(ValidationState.AuthorizedWarning);
+        }
+
+        private static bool IsInvalidPassword(string potentialPassword)
```
michaelcfanning, line 70: >IsInvalidPassword [](start = 28, length = 17)

Have we got this right? Which is clearer:

IsValidPassword and !IsValidPassword

or 
IsInvalidPassord and !InvalidPassword

my open question to you.

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -69,5 +69,6 @@
   $SEC101/053.GoogleServiceAccountKeyConsoleFormat="private_key":\s*"-----BEGIN PRIVATE KEY-----\\n(?s)(?<key>[^"]+)(?-s)\\n-----END PRIVATE KEY-----\\n(?s).{1,200}(?-s)"client_id"\s*:\s*"(?<account>[\w.-]+)"
   $SEC101/053.GoogleServiceAccountKeyConsoleRestFormat="name": "projects\/[\w-]+\/serviceAccounts\/[\w@.-]+\/keys\/(?<key>[^"]+)",
   $SEC101/054.AlibabaAccessKey=\b(?<account>LTAI([a-zA-Z0-9]{12}|[a-zA-Z0-9]{20}))\b(?s).{0,50}(?-s)\b(?<password>[\w]{30})\b
+  $SEC101/055.Cmdkey=cmdkey.{1,200}\/pass:\s*(?<password>[^\["'\s`%\\]{2,})
```
michaelcfanning, line 72: >Cmdkey [](start = 14, length = 6)

you're not capturing enough data here! cmdkey allows the user to be specified (add it to the account fingerprint element) and target name (which goes into the fingerprint host slot). Please note, if you have a server name, an account and a password, you can try connecting to it! So we also need a dynamic validator.

 cmdkey /add:targetname /user:username /pass:password

jameswinkler: Regarding the username/account, I make a second pass in the validator to find it.  Passwords are more sensitive and I didn't want to exclude exposed passwords because we couldn't correctly parse the username.

Regarding dynamic validation, my understanding is that cmdkey is designed to store *any* kind of credentials.  Assuming we could identify a username and server, we don't know what that server is, what it's listening for, or how to connect to it.  Is it a VM?  A SQL Server database?  Etc.  I don't understand how dynamic validation is possible unless we run through every technology of which we are aware.  Thoughts?

---
In reply to: [598013129](https://github.com/microsoft/sarif-pattern-matcher/pull/317#discussion_r598013129) [](ancestors = 598013129)

Src/Plugins/Security/SEC101_055.CmdkeyValidator.cs
jameswinkler: I sincerely had this debate with myself moments before checking in because I actually reversed the boolean logic by mistake.  I chose to leave the method as "IsInvalid" because I can't say with confidence that a password that passes every check within is "valid".  There may be patterns which clearly identify an invalid passwords that I have not implemented.  I can say for sure that passwords failing one of these conditions is almost certainly Invalid.  I can't say that passwords passing all these conditions is valid.

---
In reply to: [598012574](https://github.com/microsoft/sarif-pattern-matcher/pull/317#discussion_r598012574) [](ancestors = 598012574)

Src/Plugins/Security/SEC101_055.CmdkeyValidator.cs
jameswinkler: Brainfart, that's much simpler.

It's hard to say what will be valid characters in all contexts, I suspect that in some contexts $ is allowed.  And the regex for this validator does indeed allow it...except for the very first character.  So yes, we may miss some passwords, but only those passwords which put $ in the first position.  In the data I viewed, I saw *many* more potential false positives where a powershell variable was being used.  So I think this is a worthwhile tradeoff.  Let me know if you think differently.

---
In reply to: [598012251](https://github.com/microsoft/sarif-pattern-matcher/pull/317#discussion_r598012251) [](ancestors = 598012251)

Src/Plugins/Security/Security.SharedStrings.txt
jameswinkler: As discussed, we're merging this regex with AadUser and NetworkCredentials validators

---
In reply to: [598015790](https://github.com/microsoft/sarif-pattern-matcher/pull/317#discussion_r598015790) [](ancestors = 598015790,598013129)

PR #326, opened 2021-03-23
Fixing PAT regex

eddynaka:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -75,5 +75,5 @@
   $SEC101/053.GoogleServiceAccountKeyConsoleFormat="private_key":\s*"-----BEGIN PRIVATE KEY-----\\n(?s)(?<key>[^"]+)(?-s)\\n-----END PRIVATE KEY-----\\n(?s).{1,200}(?-s)"client_id"\s*:\s*"(?<account>[\w.-]+)"
   $SEC101/053.GoogleServiceAccountKeyConsoleRestFormat="name": "projects\/[\w-]+\/serviceAccounts\/[\w@.-]+\/keys\/(?<key>[^"]+)",
   $SEC101/054.AlibabaAccessKey=\b(?<account>LTAI([a-zA-Z0-9]{12}|[a-zA-Z0-9]{20}))\b(?s).{0,50}(?-s)\b(?<password>[\w]{30})\b
-  $SEC101/102.AdoPat=(^|[^2-7a-z])(?<refine>[2-7a-z]{52})([^2-7a-z]|$)
+  $SEC101/102.AdoPat=(^|[^2-7a-z])(?<refine>(?<key>[2-7a-z]{52}))([^2-7a-z]|$)
```
eddynaka, line 78: >key [](start = 47, length = 3)

without this, we were getting more characters than needed.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -459,7 +459,7 @@ private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext c
 
                 Regex regex =
                     CachedDotNetRegex.GetOrCreateRegex(matchExpression.ContentsRegex,
-                                                      RegexDefaults.DefaultOptionsCaseInsensitive);
+                                                      RegexDefaults.DefaultOptionsCaseSensitive);
```
eddynaka, line 462: >DefaultOptionsCaseSensitive [](start = 68, length = 27)

we were not considering casing here, which is wrong. Since we might have a regex that is case sensitive.

Src/Plugins/Security/Security.SharedStrings.txt
eddynaka: we also have this regex:
\b(?<pat>[2-7a-z]{52})(\b|$)  

what should we use?

---
In reply to: [600038758](https://github.com/microsoft/sarif-pattern-matcher/pull/326#discussion_r600038758) [](ancestors = 600038758)

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -75,5 +75,5 @@
   $SEC101/053.GoogleServiceAccountKeyConsoleFormat="private_key":\s*"-----BEGIN PRIVATE KEY-----\\n(?s)(?<key>[^"]+)(?-s)\\n-----END PRIVATE KEY-----\\n(?s).{1,200}(?-s)"client_id"\s*:\s*"(?<account>[\w.-]+)"
   $SEC101/053.GoogleServiceAccountKeyConsoleRestFormat="name": "projects\/[\w-]+\/serviceAccounts\/[\w@.-]+\/keys\/(?<key>[^"]+)",
   $SEC101/054.AlibabaAccessKey=\b(?<account>LTAI([a-zA-Z0-9]{12}|[a-zA-Z0-9]{20}))\b(?s).{0,50}(?-s)\b(?<password>[\w]{30})\b
-  $SEC101/102.AdoPat=(^|[^2-7a-z])(?<refine>[2-7a-z]{52})([^2-7a-z]|$)
+  $SEC101/102.AdoPat=\b(?<pat>[2-7a-z]{52})(\b|$)
```
michaelcfanning, line 78: >(\b|$) [](start = 43, length = 6)

the previous pattern should actually work at the end and it's more precise.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
michaelcfanning: Ouch! by the way, the spacing is wrong on this argument, it should be moved one column to the right. 

---
In reply to: [600039240](https://github.com/microsoft/sarif-pattern-matcher/pull/326#discussion_r600039240) [](ancestors = 600039240)

PR #333, opened 2021-03-25
Add rank to all results (which is the normalized shannon entropy of t…

michaelcfanning:
…he password/key component of the fingerprint).

@cfaucon @eddynaka @jameswinkler


Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -120,6 +160,15 @@ private enum ParseState
...
         public string PersonalAccessToken { get; set; }
 
+        /// <summary>
+        /// Gets or sets a value that is the count of the valid symbols that
+        /// may appear in the fingerprint element that represents an actual
+        /// secret, such as a password or key. This value, when present, will
+        /// be used to compute the rank of the fingerprint (which is itself
+        /// the normalized Shannon entropy of the key or password).
+        /// </summary>
+        public int SecretSymbolSetCount { get; set; }
```
michaelcfanning, line 170: >SecretSymbolSetCount [](start = 19, length = 20)

Note that this change can't really be leveraged until we make another change which is to simply return a Fingerprint value from every analysis method. We previously used a string in order to avoid taking a dependency on the type, but no that we have an SDK it's silly to translate everything through a string return value.

eddynaka (approved):
:shipit:

PR #334, opened 2021-03-25
Improving slack webhook

### LINKED ISSUE #330: Slack web hook rule should break fingerprint into key and id components.
@eddynaka, @jameswinkler, this is high priority.

We currently persist the complete slack web URL as the secret. This prevents us from, say, examining the shannon entropy of the key element of the URL. I think it breaks down like so.

Can we convert the fingerprint to use the id and key elements asap? Thanks!

https://hooks.slack.com/services/{id}/{key}

eddynaka:
Fixes #330


michaelcfanning (approved):
:shipit:

Issue #331, opened 2021-03-25
GoogleGCMServiceAccountValidator casing incorrect and fingerprint is wrong


michaelcfanning:
@eddynaka , @jameswinkler , this is a small fix but high priority

The correct casing of this rule name is GoogleGcmServiceAccountValidator. In .NET, you never case more than two letters in sequence as upper-case.

Also, the fingerprint for this check is broken. Resource is not the right property to use. A 'resource' is a sub-component of a container. So, a database is a resource of a database server.
 
This rule in general looks weak. What's the insecurity here? If we can't detect an exposure of interest, that can be validated, please just consider deleting this check for now.

eddynaka:
I will exclude GCM since we are not capturing anything relevant and we can't validate.

PR #337, opened 2021-03-25
Removing Hockey and GCM validator

### LINKED ISSUE #332: Please add tests or delete rules for Hockey app and Facebook app credentials
@jameswinkler, @eddynaka

We have no meaningful tests for these checks. Please review these rules for value. And build them out if they should be retained or delete them if they are weak.

eddynaka:
Fixes #332 , #331


michaelcfanning (approved):
:shipit:

PR #339, opened 2021-03-25
Akamai validator

### LINKED ISSUE #338: Author Akamai detection (and validator if feasible)
@eddynaka, this one looks useful, let's make it a priority if we can.

eddynaka:
Fixes #338

### DIFF
--- a/Src/Plugins/Security/SEC101_015.AkamaiKeyValidator.cs
+++ b/Src/Plugins/Security/SEC101_015.AkamaiKeyValidator.cs
@@ -1,0 +1,110 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.Utilities;
+using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
+{
+    public class AkamaiKeyValidator : ValidatorBase
+    {
+        internal static AkamaiKeyValidator Instance;
+
+        static AkamaiKeyValidator()
+        {
+            Instance = new AkamaiKeyValidator();
+        }
+
+        public static string IsValidStatic(ref string matchedPattern,
+                                           ref Dictionary<string, string> groups,
+                                           ref string failureLevel,
+                                           ref string fingerprint,
+                                           ref string message)
+        {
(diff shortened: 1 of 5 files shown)


Src/Plugins/Security/SEC101_015.AkamaiKeyValidator.cs
```diff
@@ -0,0 +1,110 @@
...
+            var fingerprint = new Fingerprint(fingerprintText, false);
+
+            string id = fingerprint.Id;
+            string key = fingerprint.Key;
+            string host = fingerprint.Host;
+            string pwd = fingerprint.Password;
+
+            using HttpClient httpClient = CreateHttpClient();
+            httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue(
+                $"EG1-HMAC-SHA256",
```
michaelcfanning, line 81: >EG1-HMAC-SHA256 [](start = 18, length = 15)

I wonder why this header value is named this way? I don't see that we've done any HMAC generation!

eddynaka: Don't know as well.
[https://developer.akamai.com/legacy/introduction/Client_Auth.html](https://developer.akamai.com/legacy/introduction/Client_Auth.html)

---
In reply to: [601832564](https://github.com/microsoft/sarif-pattern-matcher/pull/339#discussion_r601832564) [](ancestors = 601832564)

michaelcfanning: Hello! you have only done part of the work. :) First you need to create the string that you've created! Then initialize an HMAC 256 signing object using the signing key. This will give you a signature. Then you create a string such as the following to use as the actual authorization header.

 Authorization:EG1-HMAC-SHA256 client_token=akaa-xxxxxxxxxxxxxxxx-xxxxxxxxxxxxxxxx;
 access_token=akaa-xxxxxxxxxxxxxx-xxxxxxxxxxxxxx;timestamp=20130817T02:49:13+0000;
 nonce=dd9957e2-4fe5-48ca-8d32-16a772ac6d8f;signature=xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
EdgeGrid Request Signature

---
In reply to: [601834361](https://github.com/microsoft/sarif-pattern-matcher/pull/339#discussion_r601834361) [](ancestors = 601834361,601832564)

michaelcfanning: Here's code to produce the HMAC. Double-checking your initial string, not sure that it's right! I'll call you.


using (var hmac = new HMACSHA256(Convert.FromBase64String(key)))

---
In reply to: [601840967](https://github.com/microsoft/sarif-pattern-matcher/pull/339#discussion_r601840967) [](ancestors = 601840967,601834361,601832564)

PR #348, opened 2021-03-31
MovingFiles after building

eddynaka:


Src/Plugins/Security/Security.csproj
```diff
@@ -59,8 +55,17 @@
...
-  <Target Name="CopyingSpam" AfterTargets="Build">
+  <Target Name="CopyingSecurityToSpam" AfterTargets="Build">
+
+    <ItemGroup>
+      <SecurityFile Include="$(OutputPath)Security.dll" />
+      <SpamFiles Include="$(MSBuildThisFileDirectory)*.json" />
+      <SdkFile Include="$(OutputPath)Sarif.PatternMatcher.Sdk.dll" />
+    </ItemGroup>
+
+    <Copy SourceFiles="@(SdkFile)" DestinationFolder="$(SolutionDir).spam\Security\" OverwriteReadOnlyFiles="true" SkipUnchangedFiles="false"  ContinueOnError="true" />
```
eddynaka, line 66: >ContinueOnError="true" [](start = 143, length = 22)

we need this because it will try to find the dll for net472 and netstandard2.0. Since we dont ship net472, lets ignore this.

Src/Plugins/Security/Security.csproj
```diff
@@ -59,8 +55,17 @@
...
     <Copy SourceFiles="@(SpamFiles)" DestinationFolder="$(SolutionDir).spam\Security\" OverwriteReadOnlyFiles="true" SkipUnchangedFiles="false" />
+    <Copy SourceFiles="@(SecurityFile)" DestinationFolder="$(SolutionDir).spam\Security\" OverwriteReadOnlyFiles="true" SkipUnchangedFiles="false" ContinueOnError="true" />
```
eddynaka, line 68: >ContinueOnError="true" [](start = 147, length = 22)

there are times it has a race condition, generating an error. but the dll is copied to the  dest folder. so, lets ignore as well.

michaelcfanning (approved):
:shipit:

PR #349, opened 2021-03-31
Fingerprint simplification

michaelcfanning:
@eddynaka, @cfaucon, @jameswinkler 

Look out. Here it comes.


Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -132,9 +132,9 @@
         },
         {
           "Id": "SEC101/015",
-          "Name": "DoNotExposePlaintextSecrets/AkamaiKey",
-          "ContentsRegex": "$SEC101/015.AkamaiKey",
-          "MessageArguments": { "secretKind": "Akamai key" }
+          "Name": "DoNotExposePlaintextSecrets/AkamaiCredentials",
```
michaelcfanning, line 135: >AkamaiCredentials [](start = 47, length = 17)

This secret kind requires multiple components, and so by convention the proper name is 'credentials'.

Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -92,9 +92,16 @@ public ValidatorsCache(IEnumerable<string> validatorBinaryPaths = null, IFileSys
...
-                        obj: null, arguments);
+                try
+                {
+                    validationState =
+                        (ValidationState)isValidStaticMethodInfo.Invoke(
+                            obj: null, arguments);
+                }
+                catch (TargetInvocationException e)
+                {
+                    throw e.InnerException;
```
michaelcfanning, line 103: >InnerException [](start = 28, length = 14)

Now the console output and SARIF logs will show us the actual exception in the called component. Hooray!

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -14,21 +14,14 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
     public struct Fingerprint
```
michaelcfanning, line 14: >Fingerprint [](start = 18, length = 11)

The heart of the change. We delete 'hmac' as unused. We rename 'acct' key to 'account' (as nothing else is abbreviated after this change)'

The following secret keys are all replaced by 'secret' -> key, pat, pwd, skey/128, skey/256, sasToken

Src/Sarif.PatternMatcher/ValidatorsCache.cs
eddynaka: should we move this to the previous try/finally?

---
In reply to: [604541612](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r604541612) [](ancestors = 604541612)

Src/Plugins/Security/SEC101_025.SendGridApiKeyValidator.cs
```diff
@@ -51,11 +51,14 @@ public static ValidationState IsValidDynamic(ref Fingerprint fingerprint, ref st
                                                       out Fingerprint fingerprint)
         {
             fingerprint = default;
-            string key = groups["key"];
+            if (!groups.TryGetValue("secret", out string secret))
```
jameswinkler, line 54: Shoudl this be "TryGetNonEmptyValue" ?

eddynaka: in theory, the regex would only be caught if secret exists. so, IMO, we would not need it

---
In reply to: [605118483](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r605118483) [](ancestors = 605118483)

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
jameswinkler: Is it possible we could have multiple secrets (perhaps different types) in the same match/result?

---
In reply to: [604542444](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r604542444) [](ancestors = 604542444)

Src/Plugins/Security/SEC101_025.SendGridApiKeyValidator.cs
jameswinkler: True, but regex may change in the future.  Excluding null/empty results improves our fault tolerance.  No huge performance hit so there's no reason to NOT do it, IMO

---
In reply to: [605119939](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r605119939) [](ancestors = 605119939,605118483)

Src/Sarif.PatternMatcher/ValidatorsCache.cs
jameswinkler: Agree that a single try/catch is cleaner and I'm pretty sure behaviorally equivalent

---
In reply to: [604854469](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r604854469) [](ancestors = 604854469,604541612)

michaelcfanning: Yes! good note, will do.

---
In reply to: [605224616](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r605224616) [](ancestors = 605224616,604854469,604541612)

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
michaelcfanning: Generally, we expect one result per exposed secret. There is a possible problem, though (which may be what you're actually saying), if a single compromised credential has multiple secrets that contribute to it. In this case, currently, you'd need to define a mini-language for the secret value to express all the component secrets. :( If/when this problem emerges, we could take a look at making a change. if you have any immediate thoughts, let me know.

---
In reply to: [605134680](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r605134680) [](ancestors = 605134680,604542444)

jameswinkler: Nah, just playing devil's advocate.  Makes sense to me, and I never know where to put things so this simplifies that decision-making significantly.

---
In reply to: [605228417](https://github.com/microsoft/sarif-pattern-matcher/pull/349#discussion_r605228417) [](ancestors = 605228417,605134680,604542444)

eddynaka (approved):
:shipit:

PR #352, opened 2021-04-02
Tweak HttpAuthorizationRequestHeader

jameswinkler:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,7 +23,7 @@
 
 # $FacebookAppCredentials=(?si)(((facebook|fb)(.{0,20})?[`'"\s>](?<id>[0-9]{13,17})[^0-9].{1,500}?)|((facebook|fb)(.{0,20})?[`'"\s>](?<secret>[0-9a-f]{32})([^0-9a-f]?)).{1,500}?){2}
 
-  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?<host>(http|ftp|https):\/\/[\w_-]+(?:(?:\.[\w_-]+)+))(?<resource>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?<secret>[a-z0-9\/+_.=]{10,})
+  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?<host>(http|ftp|https):\/\/[\w_.-]{1,200})(?<resource>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?<secret>[a-z0-9\/+_.=]{10,})
```
eddynaka, line 26: >(?<secret>[a-z0-9\/+_.=]{10,}) [](start = 202, length = 30)

we could change to:
(?<secret>[^'".....]+), with that, we wouldn't be restricted to at least 10 characters. What do u think? (change the ... for all the possible combinations that aren't accepted in a basic, probably ><; ' "

jameswinkler: Done.

---
In reply to: [606340616](https://github.com/microsoft/sarif-pattern-matcher/pull/352#discussion_r606340616) [](ancestors = 606340616)

michaelcfanning: Is there a new test here you could add? Why was this change helpful? Generally, any change driven by an observed bug/unhandled exception, etc., needs to come with a test.

---
In reply to: [606362763](https://github.com/microsoft/sarif-pattern-matcher/pull/352#discussion_r606362763) [](ancestors = 606362763,606340616)

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/Security.SharedStrings.txt
jameswinkler: The first change, to the host, is functionally equivalent.  ?: is a non-capturing match, not supported by RE2.  So no behavioral changes expected and no unit tests required.  I can't speak to the second change, it was something Eddy suggested.  @eddynaka , do you want to comment or propose a unit test?

---
In reply to: [606364018](https://github.com/microsoft/sarif-pattern-matcher/pull/352#discussion_r606364018) [](ancestors = 606364018,606362763,606340616)

michaelcfanning (approved):
:shipit:

PR #353, opened 2021-04-02
Rule name conventions

michaelcfanning:
@eddynaka @jameswinkler 

Rename MailgunApiKey to MailgunApiCredentials.
Rename AlibabaAccessKey to AlibabaCloudCredentials.

`Credentials` in our naming convention indicates a bundle (i.e., more than one) components that together allow access to a secured thing. For example, a user name and password, or an Azure client id and secret.

We reserve `Key` for a standalone token that on its own provides access, such as an ADO personal access token.

An `id` and `secret` combination may appear in rules that refer to connection strings, credentials or (in a single case for slack) a webhook.

A standalone use of ?<secret> generally occurs in rule names that refer to a `pat`, `token`, or `key`.


Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_026.MailgunApiCredentials.ps1
```diff
@@ -1,5 +1,5 @@
 # AppSettings style
 <appSettings>
   <add key="MailgunDomain" value="deadbeef0deadbeef0deadbeef.mailgun.org" />
-  <add key="MailgunApiKey" value="0deadbeef00deadbeef00deadbeef000-deadbeef-deadbeef" />
+  <add key="MailgunApiCredentials" value="0deadbeef00deadbeef00deadbeef000-deadbeef-deadbeef" />
```
jameswinkler, line 4: The difference between "Key" and "Credentials" is purely internal, right?  Input files should not change, if I'm not mistaken.

jameswinkler (approved):
:shipit:

eddynaka (approved):
:shipit:

PR #358, opened 2021-04-02
Part in asset fingerprint only

michaelcfanning:
@eddynaka, this prevents the new `part` property from surfacing in a validation fingerprint.

Note that `part` stands for 'Provider And Resource Type'. :)

I need this change to unblock. I will add a reflection driven test later.


eddynaka (approved):
:shipit:

PR #362, opened 2021-04-05
Cloudant

michaelcfanning:
@eddynaka @jameswinkler


Src/Plugins/Security/SEC101_035.CloudantCredentialsValidator.cs
```diff
@@ -75,52 +79,76 @@ public static ValidationState IsValidDynamic(ref Fingerprint fingerprint, ref st
...
+
+                using var client = new HttpClient(handler)
                 {
-                    var handler = new HttpClientHandler();
-                    handler.Credentials = new NetworkCredential(id, password);
+                    BaseAddress = new Uri(uri),
+                };
 
-                    using var client = new HttpClient(handler)
+                using HttpResponseMessage response = clientWithNoCredentials
```
eddynaka, line 119: >clientWithNoCredentials [](start = 53, length = 23)

this should be client

eddynaka:
>                 !groups.TryGetNonEmptyValue("host", out string host) ||

we should remove host

---
Refers to: Src/Plugins/Security/SEC101_015.AkamaiCredentialsValidator.cs:54 in 8da8d8e. [](commit_id = 8da8d8ec668518342b29e7bae0c2ceddf52804c5, deletion_comment = False)

Src/Plugins/Security/SEC101_035.CloudantCredentialsValidator.cs
jameswinkler, line 119: Should this be response = "client.etc().etc()" ?

eddynaka (approved):
:shipit:

PR #43, opened 2020-12-18
Adding fingerprint tag

eddynaka:
Fixes #35

PR #372, opened 2021-04-10
Add new validator for Nuget Credentials

jameswinkler:


Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_043.NuGetCredentials.ps1
```diff
@@ -0,0 +1,36 @@
...
+    <Test_x0020_Source>
+        <add key="Username" value="user" />
+        <add key="ClearTextPassword" value="%TestSourcePassword%" />
+    </Test_x0020_Source>
+</packageSourceCredentials>
+
+<packageSources>\nstuff\n<\/packageSources>\n
+<packageSourceCredentials >\n      <buildxl>\n          <add key=\"Username\" value=\"domino\" />\n          <add key=\"Password\" value=\"$(place.variable.name)\" />
+
+<configuration><packageSources><add key=\"dummyKey\" value=\"https://location.visualstudio.com/_packaging/dummyKey/nuget/v3/index.json\" /></packageSources><packageSourceCredentials><dummyKey><add key=\"Username\" value=\"dummyUserName\" /><add key=\"ClearTextPassword\" value=\"dummyPassword\" /></dummyKey></packageSourceCredentials></configuration>
```
jameswinkler, line 36: I still need to correct the XML parsing to be able to catch this example.

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_043.NuGetCredentials.ps1
```diff
@@ -0,0 +1,36 @@
+<packageSources>
```
jameswinkler, line 1: Update the usernames and passwords so they can be more easily identified in the sarif file.

Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -354,6 +354,12 @@
           "ContentsRegex": "$SEC101/042.DynatraceToken",
           "MessageArguments": { "secretKind": "Dynatrace Key" }
         },
+        {
+          "Id": "SEC101/043",
```
jameswinkler, line 358: Should we merge this with rule 31?

Src/Plugins/Security/Utilities/StaticStringHelper.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.Utilities
+{
+    public static class StaticStringHelper
```
michaelcfanning, line 6: >StaticStringHelper [](start = 24, length = 18)

Here's a rule for you: please don't create any static helper classes with a single method in it. Instead, create a single file named Utilities to hold diverse helpers until there's enough of them to warran grouping/refactoring.

michaelcfanning: Create a file named 'FilteringHelpers'. Move this code into it and delete this file. Move all DomainFilteringHelper code into and delete that file. Thanks. Once you have had the experience of watching years of file/class/etc. profileration lead to a code base that is unwieldy and hard to manage, you may begin to agree with me that this is a worthwhile concern. For now, I don't require you to be persuaded of that. :) But please do make the change, I appreciate it.

---
In reply to: [611237162](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611237162) [](ancestors = 611237162)

Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
michaelcfanning: It's a good question. They are distinct secrets in the NuGet eco-system and the help for each kind would be distinct. This is my usual test for whether a rule stands on its own, so let's keep these distinct for now.

---
In reply to: [610962329](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r610962329) [](ancestors = 610962329)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        protected override ValidationState IsValidStaticHelper(ref string matchedPattern,
+                                                               ref Dictionary<string, string> groups,
+                                                               ref string failureLevel,
+                                                               ref string message,
+                                                               out Fingerprint fingerprint)
+        {
+            fingerprint = default;
+
+            if (!groups.TryGetNonEmptyValue("secret", out string secret) ||
```
michaelcfanning, line 56: >groups.TryGetNonEmptyValue [](start = 17, length = 26)

Please follow the consistent order of elements followed by other checks. It's easy to remember, we waterfall from shortest args to longest, to help with readability.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+            }
+
+            if (StaticStringHelper.LikelyPowershellVariable(secret))
+            {
+                return ValidationState.NoMatch;
+            }
+
+            fingerprint = new Fingerprint
+            {
+                Host = host, // technically, this is XML which could contain multiple hosts.
```
michaelcfanning, line 70: >Host [](start = 16, length = 4)

please us the order that others checks go, against, we waterfall from shortest name to longest, so if, host, secret, and platform.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+                Platform = nameof(AssetPlatform.NuGet),
+            };
+
+            return ValidationState.Unknown;
+        }
+
+        protected override ValidationState IsValidDynamicHelper(ref Fingerprint fingerprint, ref string message, ref Dictionary<string, string> options)
+        {
+            string hostXmlAsString = fingerprint.Host;
+            string username = fingerprint.Id;
```
michaelcfanning, line 82: >string username [](start = 12, length = 15)

please extract parameters to variables named id, host, etc. order them from shortest to longest. you can review other rules to observe the consistent style we prefer.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+            string password = fingerprint.Secret;
+
+            XmlDocument hostXml = new XmlDocument();
```
michaelcfanning, line 85: >XmlDocument [](start = 12, length = 11)

this needs to be 'var'. do me a favor: consult with Eddy about getting your style settings corrected so that VS will tell you about the issue. this is one of those style nits that is very fatiguing to receive and to give in a code review context, in my opinion.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+                Host = host, // technically, this is XML which could contain multiple hosts.
+                Id = id,
+                Secret = secret,
+                Platform = nameof(AssetPlatform.NuGet),
+            };
+
+            return ValidationState.Unknown;
+        }
+
+        protected override ValidationState IsValidDynamicHelper(ref Fingerprint fingerprint, ref string message, ref Dictionary<string, string> options)
```
michaelcfanning, line 79: >Fingerprint [](start = 68, length = 11)

these arguments should be left-aligned across several lines, just as other methods above are.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+        {
+            string hostXmlAsString = fingerprint.Host;
+            string username = fingerprint.Id;
+            string password = fingerprint.Secret;
+
+            XmlDocument hostXml = new XmlDocument();
+            hostXml.LoadXml(hostXmlAsString);
+
+            // hostXml looks like "<packageSources>...</packageSources>
+            List<string> hosts = hostXml
```
michaelcfanning, line 89: >hostXml [](start = 33, length = 7)

can you make this code more readable? the individual lines ares quite long. why are you indenting further to the right, what's that intended to communicate? i'm open to your style choice but would like to hear your rationale on readability.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+                                    ?.ChildNodes[0] // <packageSources>...
+                                        ?.ChildNodes.Cast<XmlNode>().Where(x => x.Name.Equals("add", System.StringComparison.OrdinalIgnoreCase)) // <add ... >  <clear/> (the first node)
+                                            .Select(x => x.Attributes["value"]?.Value ?? x.Attributes["Value"]?.Value).ToList(); // <add key="name of host" value="http://nugetfeedUrl.com" /> (we're looking for URL)
+
+            // Uusually there will probably only be a single host.
+            // In any case, attempt to pass the credentials to each host one by one.
+            using (HttpClient httpClientNoCredentials = CreateHttpClient())
```
michaelcfanning, line 96: >using [](start = 12, length = 5)

for using, please prefer the new C# syntax that doesn't require creating a new conditional block.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+            using (HttpClient httpClientWithCredentials = CreateHttpClient())
```
michaelcfanning, line 97: >httpClientWithCredentials [](start = 30, length = 25)

can you please examine all assignments to a var of this type from CreateHttpClient, here in in out other repo, to establish the consistent name we prefer for this pattern?

our goal is for all rules to look the same, as if authored by a single person, and to be as consistent as possible in style, names, structure. the reason we do this is for readability, but also to lower costs in the future when we decide on a change and attempt to apply mechanism transformations/other parsing of the code.

again, i'm sufficiently old to play the long game here. i don't like going back and making repeated passes over code that has been created organically, with diverse styles, at the point that we realize it's all getting tricky/inconsistent to maintain or update.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                byte[] byteArray = Encoding.ASCII.GetBytes($"{username}:{password}");
+                httpClientWithCredentials.DefaultRequestHeaders.Authorization =
+                    new AuthenticationHeaderValue("Basic", Convert.ToBase64String(byteArray));
+
+                foreach (string host in hosts)
```
michaelcfanning, line 103: >hosts [](start = 40, length = 5)

the fact that we have no mechanism for this check to process/return multiple results is an interesting extensibility problem we should keep in mind. FYI,@eddynaka. I'm not sure we'll be able to fix this in-place. Probably would need to be considered as a scenario for the v3 of this scanner, which will entirely decouple analysis into multiple phases.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                    using (HttpResponseMessage responseWithNoCredentials = httpClientNoCredentials
+                    .GetAsync(host, HttpCompletionOption.ResponseHeadersRead)
+                    .GetAwaiter()
+                    .GetResult())
+                    {
+                        switch (responseWithNoCredentials.StatusCode)
+                        {
+                            case HttpStatusCode.OK:
+                                // Credentials not needed, this method of verification is indeterminate.
```
michaelcfanning, line 118: >Credentials [](start = 35, length = 11)

this structual choice to instantiate two http clients and embed the actual analysis only in the case where the client w/no credentials is forbidden is an interesting idea but I'd like a cleaner expression of it. You can start by adding braces to all of these switch cases, which is the pattern elsewhere.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                                break;
+                            default:
+                                break;
+                        }
+                    }
+                }
+            }
+
+            return ValidationState.Unknown;
```
michaelcfanning, line 151: >ValidationState.Unknown [](start = 19, length = 23)

we don't do this at function exit. instead, all handling s/be in the default cases in the switch statements such that this specific line of code is unreachable. please note that you should be returning the unexpected response code received for any code paths where we are, in fact, surprised by the response code. see other rules for examples of this.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,154 @@
...
+                                // Credentials may resolve this, try again with them.
+                                using (HttpResponseMessage responseWithCredentials = httpClientWithCredentials
+                                .GetAsync(host, HttpCompletionOption.ResponseHeadersRead)
+                                .GetAwaiter()
+                                .GetResult())
+                                {
+                                    switch (responseWithCredentials.StatusCode)
+                                    {
+                                        case HttpStatusCode.OK:
+                                            // Credentials resolved the forbidden/unauthorized message we received.
```
michaelcfanning, line 131: >// Credentials resolved the forbidden/unauthorized message we received. [](start = 44, length = 71)

This isn't a particularly helpful comment. You should avoid comments that are a sort of english language description of what is obvious in the code itself. It is our goal for methods like 'ReturnAuthorizedAccess' to be sufficiently descriptive to make a comment like 'The credentials are valid, return Authorized', and similar.

My test for a code comment is whether the comment has given me new information, which is both accurate and useful.

michaelcfanning (dismissed):
:clock1:

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: I'm meeting with him in a minute about it.  Just to clarify, it's not that it doesn't notify me.  It notifies me *of the exact opposite*, telling me to use an explicit type, with a red mark.  I'm not sure if it blocks building or anything because, as you can tell, I've been doing whatever auto-correct says.  We'll try to figure it out in a bit.

---
In reply to: [611238763](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238763) [](ancestors = 611238763)

Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
jameswinkler: Rule 31 looked weird, it was one of the validators with a "refine" id in the regex and no dynamic validation, so I thought maybe it was incomplete/a stub.  Sounds like that's not the case, so taking no action!

---
In reply to: [611238190](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238190) [](ancestors = 611238190,610962329)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: I actually wanted to talk to @eddynaka and you about this.  First, I don't understand why we assign these values to variables when we don't modify them and we have random access to the dictionary in the fingerprint.  Was there a reason behind that?

Second, regarding the naming inside the validators; in my view, the fingerprint is deliberately abstract to handle as many options as possible.  But validators know internally what those abstract variables are, and it doesn't need to be so vague.  It makes sense to perform this translation from abstract to concrete if we're going to be creating separate variables for these values.  NuGetCredentialsValidator knows that "id" really means "username" and that "host" is actually a bunch of XML with hosts inside.  This is a good opportunity for self-documenting code, is it not?  We lose a lot of information if we keep the abstract names, don't we?

---
In reply to: [611238606](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238606) [](ancestors = 611238606)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: It's tough to make linq pretty.  I was intending to communicate depth of XML with the depth of the indentation.  So line 90 is the root node, line 91 is a collection of child nodes underneath that, and line 92 is a collection of attributes within those child nodes.  I'm deleting the new line on line 89 for now, do you have any thoughts how I can prettify this further?

---
In reply to: [611238908](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238908) [](ancestors = 611238908)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: I guess it's hard to tell what's obvious in the code.  Eddy mentioned that he didn't care for the nested switch statements so I wanted to be pretty verbose if I was going to contnue to use them.  Plus, even if it's likely obvious, I do think there is value in re-affirming that the reader's thought process is still aligned with the author's intent.  I also use nesting depth as a measure of complexity, and we're half way across the page at this point hahaha.  Don't you think some paranoia regarding clarity is warranted?

---
In reply to: [611240191](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611240191) [](ancestors = 611240191)

Src/Plugins/Security/Utilities/StaticStringHelper.cs
jameswinkler: Done.

---
In reply to: [611237758](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611237758) [](ancestors = 611237758,611237162)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Done.

---
In reply to: [611238349](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238349) [](ancestors = 611238349)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Done.

---
In reply to: [611238485](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238485) [](ancestors = 611238485)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Done.

---
In reply to: [611238820](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611238820) [](ancestors = 611238820)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Done.

---
In reply to: [611239012](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611239012) [](ancestors = 611239012)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Merged the two instances into one, renamed simply to "client" as in other validators.

---
In reply to: [611239235](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611239235) [](ancestors = 611239235)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Refactored, so...done, kinda.

---
In reply to: [611239974](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611239974) [](ancestors = 611239974)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
jameswinkler: Done.

---
In reply to: [611239775](https://github.com/microsoft/sarif-pattern-matcher/pull/372#discussion_r611239775) [](ancestors = 611239775)

Src/Plugins/Security/SEC103_017.PfxCryptographicKeyfileValidator.cs
```diff
@@ -11,29 +11,35 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
...
                 // These certificates are handled by the SecurePlaintextSecrets rules.
-                return ValidationState.NoMatch;
+                return ValidationResult.NoMatch;
             }
 
-            return CertificateHelper.TryLoadCertificate(matchedPattern,
-                                                        ref fingerprint,
-                                                        ref message,
-                                                        ref resultLevelKind);
+            Fingerprint fingerprint = default;
```
michaelcfanning, line 28: >Fingerprint fingerprint = default; [](http://example.com/codeflow?start=12&length=34)

the initializations are no longer required and are just taking up space, can we remove them all?

Src/Plugins/Security/SEC101_040.ShopifySharedSecretValidator.cs
```diff
@@ -17,41 +17,36 @@ static ShopifySharedSecretValidator()
...
+                                                                             ref Dictionary<string, string> groups,
+                                                                             ref string message)
         {
-            fingerprint = default;
-            resultLevelKind = default;
-
             if (!groups.TryGetNonEmptyValue("refine", out string secret))
             {
-                return ValidationState.NoMatch;
+                return ValidationResult.NoMatch;
```
michaelcfanning, line 36: >ValidationResult [](http://example.com/codeflow?start=23&length=16)

this is clever but you have a problem, i.e., even though ValidationResult.NoMatch is read-only, it's still mutable! so if anyone makes a mistake and changes NoMatch to have an entirely different fingerprint, etc., that change will flow to everyone else.

There are many ways to resolve this. The easiest is to make ValidationResult a struct with no property setters.

A second possibility is to use an interface for the return value. Then you can build a IValidationResult implementing class that is immutable after construction.

Another idea is to use a builder pattern of some kind:

return ValidationResult.CreateNoMatch();

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -78,6 +78,7 @@
   $SEC101/040.ShopifySharedSecret=\b(?<refine>shpss_(?i)[0-9a-f]{32})
   $SEC101/041.RabbitMqConnectionString=(?i)(?<refine>amqps?:\/\/(?<id>[^:]+):(?<secret>[^@]+)@(?<host>[\w_\-\:]+)\/(?<resource>[\w]+))([^0-9a-z]|$)
   $SEC101/042.DynatraceToken=\b(?<refine>dt0(?i)[a-z]{1}[0-9]{2}\.[0-9A-Z]{24}\.[0-9A-Z]{64})
-  
+  $SEC101/043.NuGetPackageSourceCredentialsXml=(?i)(?<host><\s*packageSources\s*>(?s).{0,500}?(?-s)<\\?\/packageSources\s*>)(?s).{0,200}?(?-s)[^\/]packageSourceCredentials\s*>(?s).{0,200}?(?-s)<\s*add[\s\w\\"]key=\s*\\?"Username\\?"[\s\w\\"]value\s*=\s*\\?"(?<id>[^"]+)\\?"(?s).{0,200}?(?-s)<\s*add[\s\w\\"]key=\s*\\?"(ClearText)?Password\\?"[\s\w\\"]value\s*=\s*\\?"(?<secret>[^"]+)\\?"
+# Todo #43 apiKeys
```
michaelcfanning, line 82: ># Todo #43 apiKeys  [](http://example.com/codeflow?start=0&length=19)

I think this comment can be removed

Src/Plugins/Security/SEC101_040.ShopifySharedSecretValidator.cs
michaelcfanning: Follow on, because fingerprint is a reference class, the struct idea might not be sufficient to easily guarantee immutability. i.e., a struct that provides a reference type by a getter still leaves you vulnerable to changes to that reference type

Src/Plugins/Security/SEC103_017.PfxCryptographicKeyfileValidator.cs
eddynaka: actually, we do need this. because we need the fingerprint and resultlevelkind to be populated in the following method.

Src/Plugins/Security/SEC101_038.PostgreSqlConnectionStringValidator.cs
```diff
@@ -35,18 +35,14 @@ static PostgreSqlConnectionStringValidator()
...
         }
 
-        public static ValidationState IsValidStatic(ref string matchedPattern,
-                                                    ref Dictionary<string, string> groups,
-                                                    ref string message,
-                                                    out ResultLevelKind resultLevelKind,
-                                                    out Fingerprint fingerprint)
+        public static IEnumerable<ValidationResult> IsValidStatic(ref string matchedPattern,
+                                                                  ref Dictionary<string, string> groups,
+                                                                  ref string message)
```
michaelcfanning, line 40: >message [](http://example.com/codeflow?start=77&length=7)

message needs to be a part of validation result. i.e., it's not a given that there will be a single message only for each validation result. 

generally our pattern for the static validator: all input args are to process against, all return values are bundled in ValidationResult.

by the way! i don't know why we bothered to make Dictionary<string, string> a 'ref' arg! That doesn't make any sense. You can still mutate the groups data in arbitrary ways by adding keys, or clearing the dictionary. There's no need to replace the dictionary entirely.

Src/Plugins/Security/SEC101_040.ShopifySharedSecretValidator.cs
eddynaka: for now, i will go with the third option.
and if needed we can do another pr. this is getting too complex to manipulate :p

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,222 @@
...
+                }
+                catch
+                {
+                    // Still gotta return an empty list even if parsing failed
+                    return new List<string>();
+                }
+            }
+        }
+
+        protected override IEnumerable<ValidationResult> IsValidStaticHelper(ref string matchedPattern,
```
michaelcfanning, line 84: >IsValidStaticHelper [](http://example.com/codeflow?start=57&length=19)

This rule needs improvement! There are an arbitrary number of hosts in a config file, but you can also specify an arbitrary number of user credentials!

and so this helper should return host * secrets fingerprint candidates.

[https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file)

<packageSourceCredentials>
    <Contoso>
        <add key="Username" value="user@contoso.com" />
        <add key="Password" value="..." />
        <add key="ValidAuthenticationTypes" value="basic" />
    </Contoso>
    <Test_x0020_Source>
        <add key="Username" value="user" />
        <add key="ClearTextPassword" value="hal+9ooo_da!sY" />
        <add key="ValidAuthenticationTypes" value="basic, negotiate" />
    </Test_x0020_Source>
</packageSourceCredentials>

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -0,0 +1,222 @@
...
+        }
+
+        private static List<string> ExtractHostsHelper(string hostXmlAsString)
+        {
+            var hostXml = new XmlDocument();
+            hostXml.LoadXml(hostXmlAsString);
+
+            // First attempt the most common format for package sources: <packageSources><add key="..." value="{the thing we're interested in}" /></packageSources>
+            var returnList = hostXml?.ChildNodes[0] // <packageSources>...
+                            ?.ChildNodes.Cast<XmlNode>().Where(x => x.Name.Equals("add", StringComparison.OrdinalIgnoreCase)) // <add ... >  <clear/> (the first node)
```
michaelcfanning, line 206: >< [](http://example.com/codeflow?start=45&length=1)

this code is not readable.

Src/Sarif.PatternMatcher.Sdk/ValidationResult.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
+{
+    public class ValidationResult
+    {
+        public static readonly IEnumerable<ValidationResult> NoMatch = new[]
+        {
+            new ValidationResult
+            {
+                Fingerprint = default,
+                ResultLevelKind = default,
```
michaelcfanning, line 15: this is redundant, you don't ever need to explicitly set a property to its default value;

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: added as an issue to unblock the merge.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: changed a little bit, moved the comments above the linq query.

Src/Sarif.PatternMatcher.Sdk/ValidationResult.cs
eddynaka: fixed.

eddynaka (approved):
:shipit:

PR #376, opened 2021-04-13
Fingerprint should not require Provider

eddynaka:


michaelcfanning:
>         public string ResourceProvider { get; set; }

These shouldn't still be here if you're simply using type. I guess you are doing this now to make it the least intrusive change possible? i.e., so the right thing will happen in the fingerprint with the least amount of code disruption?

---
Refers to: Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs:119 in ef168ef. [](commit_id = ef168ef0f5a50318d71a1555424c5c6c858753c1, deletion_comment = False)

michaelcfanning (approved):
:shipit:

PR #380, opened 2021-04-13
Refactoring fingerprint

eddynaka:


Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -60,9 +61,10 @@ public static ValidationState IsValidDynamic(ref Fingerprint fingerprint, ref st
             fingerprint = new Fingerprint
             {
                 Host = host,
+                Scheme = scheme,
                 Secret = secret,
-                Resource = (groups.ContainsKey("resource") && groups["resource"] != "/")
-                    ? groups["resource"]
+                Path = (groups.ContainsKey("path") && groups["path"] != "/")
```
michaelcfanning, line 66: >= [](start = 21, length = 1)

I think our pattern for this would be to have an unchecked 

groups.TryGetNonEmptyValue("path", out string path);

then just assign path to this property in the fingerprint. Cleaner.

michaelcfanning: groups.TryGetNonEmptyValue("path", out string path);

// Do more processing of path argument here to put it into a null vs. non-null condition.

fingerprint = new Fingerprint
            {
                Host = host,
                Scheme = scheme,
                Secret = secret,
                Path = path;
            };




---
In reply to: [612765053](https://github.com/microsoft/sarif-pattern-matcher/pull/380#discussion_r612765053) [](ancestors = 612765053)

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -23,7 +23,7 @@ public abstract class EndToEndTests : FileDiffingUnitTests
         // Set this value to 'true' temporarily and rerun tests to update
         // test files in all \ExpectedOutputs\ locations. This value needs
         // to be set to 'false' against in order for suites to actually pass.
-        protected override bool RebaselineExpectedResults => false;
+        protected override bool RebaselineExpectedResults => true;
```
michaelcfanning, line 26: Oops! will need to revert. :)

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_001.HttpAuthorizationRequestHeader.sarif
```diff
@@ -74,9 +74,9 @@
             }
           ],
           "fingerprints": {
-            "AssetFingerprint/v1": "[host=https://example.com]",
-            "ValidationFingerprint/v1": "[host=https://example.com][secret=SomeAuthorizationKey1111111]",
-            "ValidationFingerprintHash/v1": "d14ab28fe321441f47b1793cbf4c38ca900e94869a5d065fee8ee4b6ecbe67ce"
+            "AssetFingerprint/v1": "[host=example.com][scheme=https]",
```
michaelcfanning, line 77: >https [](start = 62, length = 5)

I thought we were going to drop scheme if it is the default of https? what do you think? 

or do you think it is clearer to always have it?

michaelcfanning (approved):
:shipit:

PR #383, opened 2021-04-15
Improving mysql validator

eddynaka:


Src/Plugins/Tests.Security/Validators/SEC101_036.MySqlConnectionStringValidatorTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        public void MySqlConnectionStringValidator_Test()
+        {
+            string fingerprintText = "";
+            if (string.IsNullOrEmpty(fingerprintText))
+            {
+                return;
+            }
+
+            var fingerprint = new Fingerprint(fingerprintText);
+            string message = null;
```
michaelcfanning, line 24: I would have put this line above, in order to improve readability, small style point.

michaelcfanning (approved):
:shipit:

PR #385, opened 2021-04-15
Use file has as secret in file extension checks with no validator.

michaelcfanning:
@eddynaka, this change will use the file hash as the secret for any rules which don't have a static validator. This works for all the 'review sensitive files' checks. The change reveals a deficit for the 'review sensitive data' rules. These latter rules should create a fingerprint of the sensitive data. For example, in the email rule, the email address itself s/be persisted as the `secret` in a fingerprint constructed by a static validator.


Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -228,6 +225,18 @@ public static void DisableValidationCaching(MethodInfo shouldCacheMethodInfo, bo
 
             if (validationMethods == null)
             {
+                if (context.TargetUri.IsAbsoluteUri)
```
eddynaka, line 228: >context.TargetUri.IsAbsoluteUri [](http://example.com/codeflow?start=20&length=31)

this would only work if we run the tool in a machine. In our current scenario, the targeturi does not point to a physical location because it does not exist.

Maybe, we should not do this check and always apply the path hashing.

michaelcfanning: ok! well, this helper requires a file path with a file.

what we can do next is to check and see whether we have the file contents available and just hash that instead, if it's there?

shall we check this in and then maybe do that follow on work? want to take that improvement? :)

You could also update the review sensitive data check to provide static validators and populate a *real* secret. for extra points, we could enable a test that ensures this plug-in also always includes a static validator for any plug in rule. :)

eddynaka (approved):

PR #390, opened 2021-04-19
Changing resultLevelKind in dynamic validation

eddynaka:


Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -102,6 +105,7 @@ static HttpAuthorizationRequestHeaderValidator()
                     responseDummy.StatusCode == HttpStatusCode.NotFound ||
                     responseDummy.StatusCode == HttpStatusCode.NonAuthoritativeInformation)
                 {
+                    resultLevelKind.Level = FailureLevel.None;
```
michaelcfanning, line 108: >resultLevelKind.Level = FailureLevel.None; [](http://example.com/codeflow?start=20&length=42)

This is a redunandt initialization, the default is None.

Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
```diff
@@ -68,7 +68,10 @@ static GoogleApiKeyValidator()
                                                                  ref Dictionary<string, string> options,
                                                                  out ResultLevelKind resultLevelKind)
```
michaelcfanning, line 69: >out ResultLevelKind resultLevelKind [](http://example.com/codeflow?start=65&length=35)

Your change is not what I had in mind exactly, let's talk.

The problem with this pattern is that it's *very* easy to get wrong, if you neglect to set something in each return path. This is a serious issue for us, because the severity indicates what is and is not a live credential. And so it is better for the code if we set 'Error' in a single place when we see Authorized. Then it will never be wrong.

What I thought we should have is that the out argument is nullable. Only in cases when a rule wants to actually override the scanner's behavior would it be set.

If it were a ref arg, then this would also be minimally intrusive, you wouldn't be required to initialize it in every call.


out ResultKind? resultLevelKind

With this approach, then every rule that has the default behavior doesn't have to do anything at all. Only rules that do something *special* would sert this to a non-null value.

btw - all this is likely to continue to churn!! i.e., we will keep making changes.

A critical new change that we need is to bundle all the things that comprise an analysis result into a returnable array of objects,

'ValidationResult'

which would consist of ValidationState, message, the nullable ResultLevelKind and a fingerprint. the helpers would return an IEnumerable<ValidationResult>. this would allow each rule to return multiple things.

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -129,6 +134,7 @@ static HttpAuthorizationRequestHeaderValidator()
                         // If this happen, it means it does not matter if we add the authentication.
                         if (responseDummy.StatusCode == response.StatusCode)
                         {
+                            resultLevelKind.Level = FailureLevel.None;
```
michaelcfanning, line 137: >FailureLevel.None [](http://example.com/codeflow?start=52&length=17)

Same here and elsewhere, remove all explicit

PR #391, opened 2021-04-19
First version of SensitiveTerms spam plugin.

yongyan-gh:


Src/Plugins/SensitiveTerms/SEC106.SensitiveTerms.json
```diff
@@ -0,0 +1,138 @@
+{
+  "ValidatorsAssemblyName": "SensitiveTerms.dll",
+  "SharedStringsFileName": "SensitiveTerms.SharedStrings.txt",
+  "Definitions": [
+    {
+      "Id": "SEC106/001",
```
michaelcfanning, line 6: Do not set the ID here, it is unnecessary: each match expression already sets its id.

Src/Plugins/SensitiveTerms/SEC106.SensitiveTerms.json
```diff
@@ -0,0 +1,138 @@
...
+      "Name": "DoNotUseSensitiveTerms.Accessibility",
+      "Level": "Warning",
```
michaelcfanning, line 8: >Warning [](http://example.com/codeflow?start=16&length=7)

"Warning" is a default (or it should be) so no need to express it directly.

Src/Plugins/SensitiveTerms/SEC106.SensitiveTerms.json
```diff
@@ -0,0 +1,138 @@
...
+      "FileNameAllowRegex": "$SourceFiles",
```
michaelcfanning, line 9: >$SourceFiles [](http://example.com/codeflow?start=29&length=12)

Why are we restricting analysis only to source files? You should create a $TextFiles expression that contains all of source code but also things like txt, markdown, etc.

PR #398, opened 2021-04-21
Enabling Pass in SearchDefinition

eddynaka:


Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -158,6 +158,13 @@ internal static SearchDefinitions PushInheritedData(SearchDefinitions definition
                         matchExpression.Level = definition.Level;
                     }
 
+                    if (matchExpression.Kind == 0)
```
suvamM, line 161: Is it possible to use a enumerated data type instead of fixed integers such as 0? It's not immediately clear to me what is the interpretation when `matchExpression.Kind` is 0.

suvamM, line 161: I'm not very familiar with the code: but this is a bit unclear to me. We check the value of `matchExpression.Kind` against a fixed integer, and then assign to it an enumerated-type value. Why not assign the appropriate enumerated-type value in the first place? Please feel free to ignore this comment if it is out of context.

eddynaka: just change to the correct enumeration.

PR #400, opened 2021-04-21
Refactor match interface for named capturing groups

v-kevinvu:


Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -218,6 +221,11 @@ public void Regex2_MatchWithNamedGroups()
             Assert.Equal("a", submatchStrings[1]);
             Assert.Equal("b", submatchStrings[2]);
             Assert.Equal("c", submatchStrings[3]);
+            Assert.Equal(4, matchGroups.Count);
+            Assert.Equal(new MatchGroup(MatchGroupType.Full, "abc"), matchGroups[0]);
+            Assert.Equal(new MatchGroup(MatchGroupType.Named, "g1", "a"), matchGroups[1]);
+            Assert.Equal(new MatchGroup(MatchGroupType.Anonymous, "b"), matchGroups[2]);
+            Assert.Equal(new MatchGroup(MatchGroupType.Named, "g2", "c"), matchGroups[3]);
```
v-kevinvu, line 228: @eddynaka
Does this interface align with what you're expecting?

Src/RE2.Managed/Regex2.cs
```diff
@@ -217,7 +218,8 @@ public static void ClearRegexes()
             string text,
             out Dictionary<string, int> groupName2Index,
             out Dictionary<int, string> index2GroupName,
-            out List<string> submatchStrings)
+            out List<string> submatchStrings,
```
v-kevinvu, line 221: Will delete these once we decide on the interface.

Src/RE2.Managed/Regex2.cs
```diff
@@ -269,13 +272,33 @@ public static void ClearRegexes()
                         submatchStrings.Add(submatchString);
                     }
 
+                    // Build MatchGroup list
+                    matchGroups = new List<MatchGroup>();
```
eddynaka, line 276: >List [](http://example.com/codeflow?start=38&length=4)

change to dictionary<string, matchgroup>
for the annonymous, use 0->n
for the named, use groupname

this is the same way .NET Regex works

eddynaka: that would gives us O(1) access to the named groups

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -190,18 +190,21 @@ public void Regex2_MatchWithNamedGroups()
...
             Assert.True(isMatch);
             Assert.Empty(groupName2Index);
             Assert.Empty(index2GroupName);
             Assert.Single(submatchStrings);
             Assert.Equal("abc", submatchStrings[0]);
+            Assert.Single(matchGroups);
+            Assert.Equal(new MatchGroup(MatchGroupType.Full, "abc"), matchGroups[0]);
 
-            isMatch = Regex2.Matches(@"(?P<g1>a)(b)(?P<g2>c)", "abc", out groupName2Index, out index2GroupName, out submatchStrings);
+            isMatch = Regex2.Matches(@"(?P<g1>a)(b)(?P<g2>c)", "abc", out groupName2Index, out index2GroupName, out submatchStrings, out matchGroups);
```
eddynaka, line 207: >abc [](http://example.com/codeflow?start=64&length=3)

can you make the following test:
use "abc abc abc"

this should show 3 matches, and in each match we should see each group.

Check this output:
var matches = Regex.Matches("abc abc abc", @"(?<1>a)(b)(?<2>c)");

v-kevinvu: I've updated the test. Are these the results you are expecting?

Src/RE2.Managed/Regex2.cs
v-kevinvu: Please review updated interface.

michaelcfanning (approved):
:shipit:

eddynaka:
@michaelcfanning , i think this is not ready yet. This does not have the latest thing that we talked. Kevin is still working on this.

michaelcfanning:
Got it, sorry! Thanks for resetting to draft.

---
In reply to: [828689567](https://github.com/microsoft/sarif-pattern-matcher/pull/400#issuecomment-828689567) [](http://example.com/codeflow?ancestors=828689567,828689567)

v-kevinvu:
Sorry for the confusion. The implementation I'm working on with the changes Eddy mentioned is much different from this, so I am abandoning this PR.

---
In reply to: [828690958](https://github.com/microsoft/sarif-pattern-matcher/pull/400#issuecomment-828690958) [](http://example.com/codeflow?ancestors=828690958,828689567)

PR #406, opened 2021-04-24
Improving MySql validator

eddynaka:
## Changes
- with this change, we would also try to validate connections without database

### DIFF
--- a/Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
+++ b/Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
@@ -121,25 +121,29 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             string database = fingerprint.Resource;
 
             if (string.IsNullOrWhiteSpace(host) ||
-                string.IsNullOrWhiteSpace(database) ||
                 FilteringHelpers.LocalhostList.Contains(host))
             {
                 return ValidationState.Unknown;
             }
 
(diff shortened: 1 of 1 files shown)


Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
```diff
@@ -121,25 +121,29 @@ static MySqlConnectionStringValidator()
...
-
-            if (!string.IsNullOrWhiteSpace(port))
+            bool shouldRetry;
+            string connString;
+            if (!string.IsNullOrWhiteSpace(database))
             {
-                connString += $"Port={port}";
-            }
+                connString = $"Server={host}; Database={database}; Uid={account}; Pwd={password}; SslMode=Preferred;";
+                message = $"the '{account}' account was authenticated against database '{database}' hosted on '{host}'";
```
michaelcfanning, line 134: >the '{account}' account was authenticated against database '{database}' hosted on '{host}'"; [](http://example.com/codeflow?start=28&length=92)

Doesn't this message actually get returned when we *didn't* authenticate?

an attempt to authenticate the  '{account}' for database '{database}' hosted on '{host}' did not succeed";

eddynaka: no. we are using our helpers that override the message. so, this would only appear when we have an authorized connection.

michaelcfanning: ok!

michaelcfanning (approved):
:shipit:

PR #407, opened 2021-04-25
Improving nuget rule

eddynaka:


Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -211,5 +214,48 @@ private static List<string> ExtractHostsHelper(string hostXmlAsString)
...
+                        }
+                    }
+
+                    if (!string.IsNullOrEmpty(user) && !string.IsNullOrEmpty(password))
+                    {
+                        list.Add((user, password));
+                    }
+                }
+            }
+            catch (Exception)
```
michaelcfanning, line 253: >Exception [](http://example.com/codeflow?start=19&length=9)

You need to catch a more specific exception type here, e.g., XmlException may be appropriate. Otherwise you may swallow bugs that we don't have an easy way to detect.

I'd actually suggest wrapping the LoadXml call only in a try catch (again catching XmlException only).

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -73,19 +73,19 @@ internal static List<string> ExtractHosts(string hostXmlAsString)
...
-                !groups.TryGetNonEmptyValue("host", out string xmlHost) ||
-                !groups.TryGetNonEmptyValue("secret", out string secret))
+            if (!groups.TryGetNonEmptyValue("host", out string xmlHost) ||
+                !groups.TryGetNonEmptyValue("secret", out string xmlCredentials))
             {
                 return ValidationResult.CreateNoMatch();
             }
 
-            if (FilteringHelpers.LikelyPowershellVariable(secret))
+            if (FilteringHelpers.LikelyPowershellVariable(xmlCredentials))
```
michaelcfanning, line 82: >LikelyPowershellVariable [](http://example.com/codeflow?start=33&length=24)

Shouldn't this check occur against the contents of each XML element rather than the bundle of XML itself? 

i.e., will we actually ever enter this no match code path?

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -94,17 +94,20 @@ internal static List<string> ExtractHosts(string hostXmlAsString)
                     continue;
                 }
 
-                validationResults.Add(new ValidationResult
+                foreach ((string user, string password) in credentials)
```
michaelcfanning, line 97: >password [](http://example.com/codeflow?start=46&length=8)

maybe this is the place to do the likely powershell variable check?

but why would there be a powershell variable in XML content?? :)

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -211,5 +214,48 @@ private static List<string> ExtractHostsHelper(string hostXmlAsString)
...
+                        IEnumerable<XmlNode> current = item.Attributes.Cast<XmlNode>();
+                        if (current.Any(a => a.Value.Equals("username", StringComparison.OrdinalIgnoreCase)))
+                        {
+                            user = current.FirstOrDefault(a => a.Name.Equals("value", StringComparison.OrdinalIgnoreCase))?.Value;
+                        }
+
+                        if (current.Any(a => a.Value.Equals("password", StringComparison.OrdinalIgnoreCase) ||
+                                             a.Value.Equals("cleartextpassword", StringComparison.OrdinalIgnoreCase)))
+                        {
+                            password = current.FirstOrDefault(a => a.Name.Equals("value", StringComparison.OrdinalIgnoreCase))?.Value;
```
michaelcfanning, line 243: >password [](http://example.com/codeflow?start=28&length=8)

so 'cleartextpassword' is plainly bad, but is 'password' encrypted? if we have an encrypted password will the validator ever fire or are we just generating fingerprint candidates that will never validate?

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: removed this powershell condition.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: removed the powershell check.

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: thanks for the suggestion!

Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
eddynaka: In theory, ClearTextPassword would be the plaintext, but you could also save the same in the Password tag. Per my understanding:

when you use:
```
dotnet nuget add source [https://someServer/myTeam](https://someServer/myTeam) -n myTeam -u myUsername -p myPassword --store-password-in-clear-text
```

it would store as plaintext. If you don't use the store-password flag, it would encrypt it. But, you could manually
edit both and set the passwords. Still checking how to make an end-to-endd test

michaelcfanning (approved):
:shipit:

Issue #386, opened 2021-04-16
RE2 named capturing groups todolist


v-kevinvu:
* Integrate `GetRegexSetup` into `BuildRegex`.

eddynaka:
Already implemented x86 support moving the code to the props.

Now, we just have to check the first three bullets.

PR #411, opened 2021-04-28
Implement multi-matching for named groups

v-kevinvu:


Src/RE2.Native/RE2Native.cpp
```diff
@@ -242,3 +242,123 @@ extern "C" __declspec(dllexport) bool MatchesNamedGroups(
...
+	_Out_ GroupNameHeader** groupNameHeadersOut,
+	_Out_ uint8_t**         groupNamesBufferOut,
+	_Out_ int*              numGroupNamesOut,
+	_Out_ Submatch***       matchesOut,
+	_Out_ size_t*           numMatchesOut,
+	_Out_ int*              numSubmatchesOut,
+
+	_Out_ std::vector<GroupNameHeader>** groupNameHeadersCleanupPtrOut,
+	_Out_ std::vector<uint8_t>**         groupNamesBufferCleanupPtrOut,
+	_Out_ std::vector<Submatch*>**       matchesCleanupPtrOut
```
v-kevinvu, line 269: This should probably be combined into a single output struct.

Src/RE2.Managed/NativeMethods.cs
```diff
@@ -158,6 +191,21 @@ private static class NativeMethodsX64
...
+                StringUtf8 text,
+                GroupNameHeader** groupNameHeadersOut,
+                byte** groupNamesBufferOut,
+                int* numGroupNames,
+                Submatch*** matchesOut,
+                int* numMatchesOut,
+                int* numSubmatchesOut,
+                void** groupNameHeadersCleanupPtrOut,
+                void** groupNamesBufferCleanupPtrOut,
+                void** matchesCleanupPtrOut);
```
v-kevinvu, line 208: TODO: implement cleanup method

Src/RE2.Native/RE2Native.cpp
```diff
@@ -242,3 +242,123 @@ extern "C" __declspec(dllexport) bool MatchesNamedGroups(
...
+	Memory layout
+	match 0: submatch 0, sm 1, sm2
+	match 1: submatch 0, sm 1, sm2
+	match 2: submatch 0, sm 1, sm2
+
+	std::vector<[numSubmatches]Submatch>>
+*/
+// NOTE: The data std::vector is guaranteed to be contiguous by the C++ standard,
+//       so we can pass it to C# space for pointer-based copying.
+extern "C" __declspec(dllexport) void MatchesNamedGroups2(
```
v-kevinvu, line 256: >MatchesNamedGroups2 [](http://example.com/codeflow?start=38&length=19)

TODO: better naming

v-kevinvu:
Note that this uses a different interop approach from my previous implementation and the existing codebase.

Previously, the code in C# space would allocate and own memory which would be written by the RE2 (C++) space.

Now, it is reversed. The C++ space will allocate and own memory which will be passed back to the C# space. Then, C# space will copy that memory into C#-native structures. Finally, C# space will call back into C++ space to free the memory.

The advantages of this approach are:

1. Dynamic output size: We can extract an arbitrary count of matches, named groups, etc instead of having to predetermine an output size.
2. Code complexity: C# space does not need to allocate any memory to be written by C++ space, nor does it need to directly wrap C++ structures such as `std::vector` or `std::map`.

The disadvantages are:

1. Copying costs: We copy the output from C++ space into C# structures for easier use.
2. Code complexity: The C# space must make 2 interop calls. The first runs the regex engine and the second frees memory in C++ space.

Src/RE2.Managed/NativeMethods.cs
```diff
@@ -88,15 +88,36 @@ public static unsafe void GetNamedGroupsSetup(StringUtf8 pattern, int* numCaptur
...
+                    numMatchesOut,
+                    numSubmatchesOut,
+                    groupNameHeadersCleanupPtrOut,
+                    groupNamesBufferCleanupPtrOut,
+                    matchesCleanupPtrOut);
             }
             else
             {
-                return NativeMethodsX86.MatchesNamedGroups(pattern, text, groupNameHeaders, groupNamesBuffer, submatches);
+                throw new NotImplementedException();
```
v-kevinvu, line 120: >throw new NotImplementedException(); [](http://example.com/codeflow?start=16&length=36)

TODO: add boilerplate for this

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -184,40 +184,55 @@ public void Regex2_MatchAndClear()
         }
 
         [Fact]
-        public void Regex2_MatchWithNamedGroups()
+        public void Regex2_Groups()
```
v-kevinvu, line 187: >Regex2_Groups [](http://example.com/codeflow?start=20&length=13)

TODO: better naming

Src/RE2.Managed/Regex2.cs
```diff
@@ -187,111 +187,71 @@ public static void ClearRegexes()
...
-        /// index2GroupName = [ "abc", "a", "b", "c" ]
-        /// </code>
-        /// </example>
-        public static unsafe bool Matches(
-            string pattern,
-            string text,
-            out Dictionary<string, int> groupName2Index,
-            out Dictionary<int, string> index2GroupName,
-            out List<string> submatchStrings)
+        public static unsafe void MatchesNamedGroups2(string pattern, string text, out List<Dictionary<string, string>> matches)
```
v-kevinvu, line 190: >MatchesNamedGroups2 [](http://example.com/codeflow?start=34&length=19)

Add documentation

eddynaka: I think we could use Matches as name, since the overload would work

Src/RE2.Managed/NativeMethods.cs
eddynaka: why do we need this?
can't you call the nativemethodsx86...?

Src/RE2.Managed/NativeMethods.cs
```diff
@@ -88,15 +88,36 @@ public static unsafe void GetNamedGroupsSetup(StringUtf8 pattern, int* numCaptur
             }
         }
 
-        public static unsafe bool MatchesNamedGroups(StringUtf8 pattern, StringUtf8 text, GroupNameHeader* groupNameHeaders, byte* groupNamesBuffer, Submatch* submatches)
+        public static unsafe void MatchesNamedGroups2(
+                StringUtf8 pattern,
```
eddynaka, line 92: >StringUtf8 [](http://example.com/codeflow?start=16&length=10)

Can we replace with String8Interop?

v-kevinvu: Yes with some work, but I would like to push that to a later refactoring task.

Src/RE2.Managed/Regex2.cs
v-kevinvu: Fixed.

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
v-kevinvu: Fixed.

Src/RE2.Native/RE2Native.cpp
v-kevinvu: Fixed.

Src/RE2.Managed/NativeMethods.cs
v-kevinvu: @eddynaka 
I am having issues getting the unit tests to run for x86. I recommend we leave this unimplemented for now and push the x86 implementation and testing to a backlog item.

Src/RE2.Native/RE2Native.cpp
v-kevinvu: Done.

Src/RE2.Managed/NativeMethods.cs
v-kevinvu: Done.

Src/RE2.Managed/NativeMethods.cs
v-kevinvu: Added task here: [https://github.com/microsoft/sarif-pattern-matcher/issues/386](https://github.com/microsoft/sarif-pattern-matcher/issues/386)

Src/RE2.Managed/NativeMethods.cs
v-kevinvu: I added x86 support, but it is untested.

I added a test task here: [https://github.com/microsoft/sarif-pattern-matcher/issues/386](https://github.com/microsoft/sarif-pattern-matcher/issues/386)

Src/RE2.Managed/Regex2.cs
```diff
@@ -191,108 +191,86 @@ public static void ClearRegexes()
...
+        ///
         /// </code>
         /// </example>
-        public static unsafe bool Matches(
-            string pattern,
-            string text,
-            out Dictionary<string, int> groupName2Index,
-            out Dictionary<int, string> index2GroupName,
-            out List<string> submatchStrings)
+        public static unsafe void Matches(string pattern, string text, out List<Dictionary<string, string>> matches)
```
eddynaka, line 221: >void [](http://example.com/codeflow?start=29&length=4)

let's return bool, it would facilitate the conditions :)

Src/RE2.Managed/Regex2.cs
```diff
@@ -191,108 +191,86 @@ public static void ClearRegexes()
...
-                        Console.WriteLine(submatch.Index);
-                        string submatchString = Encoding.UTF8.GetString(textUtf8Bytes, submatch.Index, submatch.Length);
-                        submatchStrings.Add(submatchString);
-                    }
-
-                    return true;
-                }
-                else
+                // Build SubmatchIndex to GroupName map
+                var submatchIndex2GroupName = new Dictionary<int, string>();
```
eddynaka, line 237: >var submatchIndex2GroupName = new Dictionary<int, string>(); [](http://example.com/codeflow?start=16&length=60)

you could add the size to the constructor, to prevent resizing

Src/RE2.Managed/Regex2.cs
```diff
@@ -191,108 +191,86 @@ public static void ClearRegexes()
...
-            {
-                fixed (int* numCapturingGroupsPtr = &numCapturingGroups)
-                fixed (int* numNamedCapturingGroupsPtr = &numNamedCapturingGroups)
-                fixed (int* groupNamesBufferSizePtr = &groupNamesBufferSize)
+                // Build matches.
+                matches = new List<Dictionary<string, string>>();
+                for (int matchIndex = 0; matchIndex < output->NumMatches; matchIndex++)
                 {
-                    NativeMethods.GetNamedGroupsSetup(new StringUtf8(patternUtf8BytesPtr, pattern.Length), numCapturingGroupsPtr, numNamedCapturingGroupsPtr, groupNamesBufferSizePtr);
+                    var newSubmatch = new Dictionary<string, string>();
```
eddynaka, line 249: >Dictionary<string, string>(); [](http://example.com/codeflow?start=42&length=29)

same here.

Src/RE2.Managed/Regex2.cs
v-kevinvu: Done.

Src/RE2.Managed/Regex2.cs
v-kevinvu: Done.

Src/RE2.Managed/Regex2.cs
```diff
@@ -191,108 +191,86 @@ public static void ClearRegexes()
...
-        {
-            byte[] patternUtf8Bytes = Encoding.UTF8.GetBytes(pattern);
 
-            fixed (byte* patternUtf8BytesPtr = patternUtf8Bytes)
-            {
-                fixed (int* numCapturingGroupsPtr = &numCapturingGroups)
-                fixed (int* numNamedCapturingGroupsPtr = &numNamedCapturingGroups)
-                fixed (int* groupNamesBufferSizePtr = &groupNamesBufferSize)
+                // Build matches.
+                matches = new List<Dictionary<string, string>>();
```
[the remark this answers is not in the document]
v-kevinvu: Done.

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -184,40 +184,87 @@ public void Regex2_MatchAndClear()
...
+            Assert.True(matches[0].ContainsKey("a"));
+            Assert.Equal("a", matches[0]["a"]);
+            Assert.True(matches[0].ContainsKey("bb"));
+            Assert.Equal("b", matches[0]["bb"]);
+            Assert.True(matches[0].ContainsKey("ccc"));
+            Assert.Equal("c", matches[0]["ccc"]);
+        }
+
+        [Fact]
+        public void Regex2_CaptureGroups_NonOverlapping()
```
eddynaka, line 249: >Regex2_CaptureGroups_NonOverlapping [](http://example.com/codeflow?start=20&length=35)

can you create a test where we dont find a match?
just to check if the match.count>0 would work fine in this case.

v-kevinvu: Yup, I added the test case `Regex2_CaptureGroups_NoMatch`.

eddynaka (approved):

PR #414, opened 2021-05-01
Replacing .NET regex for RE2 regex

eddynaka:


Src/RE2.Managed/RE2Regex.cs
```diff
@@ -15,7 +14,9 @@ public class RE2Regex : IRegex
         public static IRegex Instance = new RE2Regex();
         public static TimeSpan DefaultTimeout = TimeSpan.FromMilliseconds(int.MaxValue - 1);
 
-        private RE2Regex() { }
+        private RE2Regex()
+        {
+        }
```
v-kevinvu, line 19: Is this needed?

Src/Sarif.PatternMatcher/CachedDotNetRegex.cs
```diff
@@ -27,7 +27,9 @@ static CachedDotNetRegex()
             RegexCache = new ConcurrentDictionary<Tuple<string, RegexOptions>, Regex>();
         }
 
-        private CachedDotNetRegex() { }
+        private CachedDotNetRegex()
+        {
+        }
```
v-kevinvu, line 32: Is this needed?

Src/Sarif.PatternMatcher/DotNetRegex.cs
```diff
@@ -19,7 +19,9 @@ public class DotNetRegex : IRegex
 
         internal static TimeSpan DefaultTimeout = TimeSpan.FromMilliseconds(int.MaxValue - 1);
 
-        private DotNetRegex() { }
+        private DotNetRegex()
+        {
+        }
```
v-kevinvu, line 24: Is this needed?

v-kevinvu (dismissed):
:shipit:

Src/RE2.Managed/RE2Regex.cs
eddynaka: my format automatically makes this change. will revert later.

Src/RE2.Managed/Regex2.cs
```diff
@@ -273,14 +273,16 @@ public static unsafe bool Matches(string pattern, string text, out List<Dictiona
                                 submatchString = Encoding.UTF8.GetString(textUtf8Bytes, submatchTextStartIndex, submatchLength);
                             }
 
+                            var flexMatch = new FlexMatch() { Success = true, Index = submatchTextStartIndex, Length = submatchLength, Value = submatchString };
```
v-kevinvu, line 276: >submatchLength [](http://example.com/codeflow?start=119&length=14)

`submatchLength` != `submatchString.Length`

Use `submatchString.Length` here.

v-kevinvu (changes requested):
:clock1:

Src/RE2.Managed/Regex2.cs
v-kevinvu, line 276: >submatchTextStartIndex [](http://example.com/codeflow?start=86&length=22)

This is the start index in a byte buffer which represents a UTF-8 encoded string.

It is not necessarily the same as the start index of the Unicode character in a Unicode string.

PR #417, opened 2021-05-01
Fix handling of optional groups in RE2 wrapper

v-kevinvu:
RE2 allows for optional capturing groups.

The way this is handled in RE2 is unexpected (to me at least) and was not explicitly documented in the header API, which is why I missed it.
RE2 reports a submatch with index=0 if the submatch is not found.

I did not explicitly handle that case in my code. This PR adds that handling.

### DIFF
--- a/Src/RE2.Managed/Regex2.cs
+++ b/Src/RE2.Managed/Regex2.cs
@@ -197,6 +197,7 @@ namespace Microsoft.RE2.Managed
         /// <param name="pattern">Pattern to search for in RE2 syntax.</param>
         /// <param name="text">Text to search.</param>
         /// <param name="matches">A list of successive, non-overlapping matches.</param>
+        /// <returns>A bool indicating if 1 or more matches were found.</returns>
         ///
         /// <example>
         /// <code>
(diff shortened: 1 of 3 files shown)


eddynaka (approved):

Src/RE2.Managed/Regex2.cs
```diff
@@ -255,7 +257,15 @@ public static unsafe bool Matches(string pattern, string text, out List<Dictiona
                         int submatchTextStartIndex = submatchRe2.Index;
                         int submatchLength = submatchRe2.Length;
 
-                        string submatchString = Encoding.UTF8.GetString(textUtf8Bytes, submatchTextStartIndex, submatchLength);
+                        string submatchString;
+                        if ((submatchTextStartIndex == -1) && (submatchLength == -1))
```
michaelcfanning, line 261: >submatchTextStartIndex [](http://example.com/codeflow?start=29&length=22)

NIT: i would tend to initialize this local with string.empty and then use a single conditional to reassign the GetString call.

Small fix that reduces vertical lines in code without reallky impacting perf.

v-kevinvu: See this PR:
[https://github.com/microsoft/sarif-pattern-matcher/pull/418](https://github.com/microsoft/sarif-pattern-matcher/pull/418)

PR #418, opened 2021-05-02
Refactor optional group handling

v-kevinvu:


Src/RE2.Managed/Regex2.cs
```diff
@@ -260,16 +260,12 @@ public static unsafe bool Matches(string pattern, string text, out List<Dictiona
...
-                                submatchString = string.Empty;
-                            }
-                            else
+                            string submatchString = string.Empty;
+                            if (!submatchRe2.IsOptionalGroupAndNotFound())
                             {
+                                int submatchTextStartIndex = submatchRe2.Index;
+                                int submatchLength = submatchRe2.Length;
                                 submatchString = Encoding.UTF8.GetString(textUtf8Bytes, submatchTextStartIndex, submatchLength);
                             }
```
v-kevinvu, line 270: @michaelcfanning This is fewer lines, but it is more confusing to read to me. What do you think?

PR #424, opened 2021-05-04
Analyze database command

eddynaka:
With this command, we will be able to do this:
```bash
.\spam.exe analyze-database c:\Temp\ --connection "kusto" --data-type Kusto --target "query" --identity "identity column" -o kusto.sarif --dynamic-validation --level "Error;Warning;Note" --insert "RegionSnippets;ContextRegionSnippets" -d "SEC101.SecurePlaintextSecrets.json"
```


Src/Sarif.PatternMatcher.Cli/AnalyzeDatabaseCommand.cs
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        public int Run(AnalyzeDatabaseOptions options)
+        {
+            try
+            {
+                options.Threads = options.Threads > 0 ? options.Threads : Environment.ProcessorCount * 4;
+                dynamicValidation = options.DynamicValidation;
+
+                var channelOptions = new BoundedChannelOptions(int.MaxValue)
```
eddynaka, line 42: >int.MaxValue [](http://example.com/codeflow?start=63&length=12)

the speed of things being added >> processing rate. Based on that, we can not have a fixed small size.

Src/Sarif.PatternMatcher.Cli/AnalyzeDatabaseCommand.cs
```diff
@@ -0,0 +1,302 @@
...
+                    default:
+                    {
+                        throw new NotImplementedException();
+                    }
+                }
+
+                Task.WhenAll(workers).Wait();
+
+                aggregatingLogger.AnalysisStopped(RuntimeConditions.None);
+                aggregatingLogger?.Dispose();
```
eddynaka, line 91: >aggregatingLogger?.Dispose(); [](http://example.com/codeflow?start=16&length=29)

we need to dispose manually. Otherwise, .NET won't know that we are still using and will dispose after some time.

PR #433, opened 2021-05-07
 Convert UTF-8 match indices to UTF-16

v-kevinvu:


Src/RE2.Managed/Regex2.cs
```diff
@@ -219,7 +220,11 @@ public static void ClearRegexes()
         ///
         /// </code>
         /// </example>
-        public static unsafe bool Matches(string pattern, string text, out List<Dictionary<string, string>> matches)
+        public static unsafe bool Matches(
+            string pattern,
+            string text,
+            out List<Dictionary<string, string>> matches,
+            out List<Dictionary<string, (int Start, int Length)>> matchIndices)
```
eddynaka, line 227: >int Start, int Length [](http://example.com/codeflow?start=41&length=21)

instead of doing this, pls, use the FlexMatch (check my other pr)

v-kevinvu: Done.

Src/RE2.Managed/Regex2.cs
```diff
@@ -219,7 +220,11 @@ public static void ClearRegexes()
         ///
         /// </code>
         /// </example>
-        public static unsafe bool Matches(string pattern, string text, out List<Dictionary<string, string>> matches)
+        public static unsafe bool Matches(
+            string pattern,
+            string text,
+            out List<Dictionary<string, string>> matches,
+            out List<Dictionary<string, FlexMatch>> matchIndices)
```
eddynaka, line 227: you dont need two structures if im not mistaken.
you could use flexmatch as well. flexmatch contains the property to hold the value.

v-kevinvu: Yes, I will remove the redundant structure.

v-kevinvu: Done.

PR #442, opened 2021-05-12
Replace file name in output with a truncated secret.

michaelcfanning:
@cfaucon, this is an important change for any consumers that attempt to replace message string arguments in-place. :)

@eddynaka

Where previously, we repeated the file location of the find:

> e:\repros\test.txt(2,16-77): error SEC101/006: 'test.txt' contains a valid legacy format GitHub personal access token (the compromised GitHub account is '[someaccount](https://github.com/someaccount)')

We now report a truncated form of the exposed secret:

> e:\repros\test.txt(2,16-77): error SEC101/006: 'ab41f9…' is a valid legacy format GitHub personal access token (the compromised GitHub account is '[someaccount](https://github.com/someaccount)')

### DIFF
--- a/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_016.StripeApiKey.sarif
+++ b/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_016.StripeApiKey.sarif
@@ -23,7 +23,7 @@
                   "text": "'{0}' was not evaluated for check '{1}' because the analysis is not relevant for the following reason: {2}."
                 },
                 "Default": {
-                  "text": "'{0}' contains {1}{2}{3}{4}{5}."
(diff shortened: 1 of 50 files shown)


Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_016.StripeApiKey.sarif
```diff
@@ -44,7 +44,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "SEC101_016.StripeApiKey.txt",
+              "sk_tes…",
```
michaelcfanning, line 47: >sk_tes [](http://example.com/codeflow?start=15&length=6)

note that some secret kinds have prefixes that already consume all the available truncation space without providing any per-secret data. We could consider allowing rules to provide a different truncation value in the future (longer than the default of 6 chars).

eddynaka (approved):
:shipit:

PR #445, opened 2021-05-15
Adding retry and max-memory options

eddynaka:


Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -51,5 +51,16 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
+        [Option(
+            "retry",
+            HelpText = "Enable retry connection if enabled.")]
+        public bool Retry { get; set; }
+
+        [Option(
+            "max-memory",
+            HelpText = "The maximum memory that can be used for RE2.",
+            Default = -1)]
+        public long MaxMemory { get; internal set; }
```
michaelcfanning, line 64: >MaxMemory [](http://example.com/codeflow?start=20&length=9)

You did not make it clear whether you are specifying KB or bytes. It is a good practice to always make this clear. Suggest this:

--max-memory-in-kb

Remember when I gave you this note and you changed the other name in this file to FileSizeInKilobytes? :)

While you're at it, let's change that other argument name to:

--file-size-in-kb

michaelcfanning (changes requested):
:clock1:

Src/Sarif.PatternMatcher/AnalyzeOptions.cs
eddynaka: fixed :)

michaelcfanning (approved):
:shipit:

PR #446, opened 2021-05-15
New intrafile analysis

michaelcfanning:
Very intrusive change imminent. This is preparatory work that cleans up some incidental things and drops in the primary change: passing FlexMatch values to validators (which contain offset data within the primary match). Now, validators can provide an arbitrary FlexMatch in order to override the region-selecting behavior. This works well for the new intrafile analysis, where a validator receives a number of components gathered across the file. The validator validates and then selects one (typically the secret itself) that should be used as the primary code region of interest.

This draft is intended to allow me a close look at some incidental things that got cleaned up along the way, mostly improvements to add refinements and fingerprints to plug-ins.

Key design changes coming next:

* We should use sets not lists for candidate fingerprints.
* We should be able to associated metadata with a regex that describes whether it is optional or required for a match and whether it should be cached.


Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -42,25 +43,26 @@ static HttpAuthorizationRequestHeaderValidator()
         }
 
         protected override IEnumerable<ValidationResult> IsValidStaticHelper(ref string matchedPattern,
-                                                                             Dictionary<string, string> groups)
+                                                                             Dictionary<string, FlexMatch> groups)
```
michaelcfanning, line 46: >FlexMatch [](http://example.com/codeflow?start=96&length=9)

The essential change. Now, data flows into the validator that captures the specific offset of every group within the larger file. A validator can choose one of these in order to focus the primary region of interest.

Src/Plugins/Security/SEC101_001.HttpAuthorizationRequestHeaderValidator.cs
```diff
@@ -42,25 +43,26 @@ static HttpAuthorizationRequestHeaderValidator()
...
             {
+                RegionFlexMatch = secret,
                 Fingerprint = new Fingerprint
                 {
-                    Host = host,
-                    Path = path,
-                    Scheme = scheme,
-                    Secret = secret,
+                    Host = host.Value,
+                    Path = path?.Value,
```
michaelcfanning, line 63: >path?.Value [](http://example.com/codeflow?start=27&length=11)

Here's an important outcome of this change, if a value, such as 'path' in this case, is optional, the flex match could itself be null. It's too bad that we've introduced this consideration into the design. We could try to do something clever in an implicit cast operator to try to overcome it, though, not strictly sure on feasibility without looking at it.

Src/Plugins/Security/SEC101_003.GoogleApiKeyValidator.cs
```diff
@@ -42,13 +43,28 @@ static GoogleApiKeyValidator()
         }
 
         protected override IEnumerable<ValidationResult> IsValidStaticHelper(ref string matchedPattern,
-                                                                             Dictionary<string, string> groups)
+                                                                             Dictionary<string, FlexMatch> groups)
         {
+            if (!groups.TryGetNonEmptyValue("secret", out FlexMatch secret))
```
michaelcfanning, line 48: >TryGetNonEmptyValue [](http://example.com/codeflow?start=24&length=19)

As part of this change, I continue to eliminate idionsyncratic behaviors, such as treating the initial match as a secret. Instead, we now make this explicit in the regex itself. This consistency will continue to help with mechanical/other transformations of our plugins as we update the model.

Src/Sarif.PatternMatcher/ExtensionMethods.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using Microsoft.RE2.Managed;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
+{
+    public static class DictionaryExtensions
+    {
+        public static Dictionary<string, string> ToStringDictionary(this IDictionary<string, FlexMatch> dictionary)
+        {
+            var stringDictionary = new Dictionary<string, string>(dictionary.Count);
```
eddynaka, line 14: >dictionary.Count [](http://example.com/codeflow?start=66&length=16)

Added this as a minor perf improvement. Also, added tests.

Src/Sarif.PatternMatcher/ExtensionMethods.cs
```diff
@@ -0,0 +1,29 @@
...
+            {
+                stringDictionary[kv.Key] = kv.Value.Value.String;
+            }
+
+            return stringDictionary;
+        }
+
+        public static Dictionary<string, FlexMatch> Copy(this IDictionary<string, FlexMatch> dictionary)
+        {
+            return new Dictionary<string, FlexMatch>(dictionary);
```
eddynaka, line 26: >dictionary [](http://example.com/codeflow?start=53&length=10)

Changed to use the default constructor. Also, added some tests.

Src/Plugins/Security/SEC102.ReviewPotentiallySensitiveData.json
```diff
@@ -7,19 +7,20 @@
...
+          "ContentsRegex": "\\b(?i)(?P<refine>[0-9a-z._%+-]*[0-9a-z]+@[0-9a-z]+[0-9a-z.-]*\\.[0-9a-z]{2,})",
+          "Message": "'{0:id}@{1:host}' is an apparent {2:dataKind}.",
           "MessageArguments": { "dataKind": "email address" },
           "Notes": [ "This will not catch every possible email address.  According to https://www.regular-expressions.info/email.html it is not reasonable to expect to do so.  The formal RFC definition of an email address includes things which some systems can't support, such as double quotes, spaces, and other special characters.  What we have here should capture about 99% of email addresses in use today." ]
         },
         {
           "Id": "SEC102/002",
           "Name": "ReviewPotentiallySensitiveData/SocialSecurityNumber",
-          "ContentsRegex": "\\b[0-9]{3}-[0-9]{2}-[0-9]{4}\\b",
+          "ContentsRegex": "\\b(?P<refine>(?P<secret>[0-9]{3}-[0-9]{2}-[0-9]{4}))\\b",
```
michaelcfanning, line 23: >refine [](http://example.com/codeflow?start=35&length=6)

'refine' is sill available for region refinement in cases where a rule is entirely JSON based: no C# runtime validator, either static or dynamic.

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102_001.EmailAddress_with_email_addresses.sarif
```diff
@@ -23,7 +23,7 @@
                   "text": "'{0}' was not evaluated for check '{1}' because the analysis is not relevant for the following reason: {2}."
                 },
                 "Default": {
-                  "text": "'{0}' contains an apparent {1}."
```
michaelcfanning, line 26: >}' [](http://example.com/codeflow?start=30&length=2)

No more file paths in this plug in, same change with others.

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102_002.SocialSecurityNumber_with_social_security_numbers.sarif
```diff
@@ -71,10 +71,10 @@
           ],
           "fingerprints": {
             "AssetFingerprint/v1": "",
-            "ValidationFingerprint/v1": "[secret=626B314E07BDE35B105543A8A75E2D257C9D4B2EB75B8BC7166489B6B01D19D3]",
-            "ValidationFingerprintHash/v1": "7550a76d0dcf856cd9325ff0f8dd10dfe306d35aa0f0973726381cbf0ae70a16"
+            "ValidationFingerprint/v1": "[secret=123-45-6789]",
```
michaelcfanning, line 74: >123 [](http://example.com/codeflow?start=49&length=3)

another improvement, burn the secret into the fingerprint rather than the file hash!

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_001.HttpAuthorizationRequestHeader.sarif
```diff
@@ -60,14 +60,14 @@
...
+                  "startColumn": 43,
                   "endLine": 4,
                   "endColumn": 70,
-                  "charOffset": 51,
-                  "charLength": 137,
+                  "charOffset": 161,
+                  "charLength": 27,
                   "snippet": {
-                    "text": "https://example.com\");\r\nvar request = new RestRequest(Method.GET);\r\nrequest.AddHeader(\"Authorization\", \"Basic SomeAuthorizationKey1111111"
+                    "text": "SomeAuthorizationKey1111111"
```
michaelcfanning, line 70: >SomeAuthorizationKey1111111 [](http://example.com/codeflow?start=29&length=27)

Now this plug in generally sets the selection region to the secret itself. This has implications for validation, as you know, this region is insufficient to revalidate the finding, except for some classes of token/key/pat.

Src/Plugins/Tests.Security/TestData/UseSecureApi/ExpectedOutputs/SEC104.Memory.Allocation_alloca.sarif
```diff
@@ -73,11 +73,10 @@
             }
           ],
           "fingerprints": {
-            "AssetFingerprint/v1": "",
-            "ValidationFingerprint/v1": "[secret=D62A0E587857E47344ED096CC889828D53B1AB3EA83E005D277AC18A2AB80B4F]",
-            "ValidationFingerprintHash/v1": "b344d33a3519f593c86ad5274421711b348eb2b250a69bdb4b6d3600bc069563"
-          },
-          "rank": 54.99
+            "AssetFingerprint/v1": "[id=_alloca]",
```
michaelcfanning, line 76: >_alloca [](http://example.com/codeflow?start=40&length=7)

use secure api rule sets the relevant api to 'id'. this data isn't considered a secret...

Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102_002.SocialSecurityNumber_with_social_security_numbers.sarif
```diff
@@ -44,7 +44,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "SEC102_002.SocialSecurityNumber_with_social_security_numbers.txt",
+              "123-45…",
```
michaelcfanning, line 47: >123 [](http://example.com/codeflow?start=15&length=3)

We use truncated secret here as in the plaintext secrets rule.

Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -210,7 +289,13 @@ public static void DisableValidationCaching(MethodInfo shouldCacheMethodInfo, bo
                     ValidationState = ValidationState.ValidatorNotFound,
                 };
 
-                if (context.TargetUri.IsAbsoluteUri)
+                validationResult.Fingerprint = SearchSkimmer.CreateFingerprintFromMatch(groups);
```
michaelcfanning, line 261: >CreateFingerprintFromMatch [](http://example.com/codeflow?start=61&length=26)

we automatically generate a fingerprint now for JSON only rules with no validators. If they use group names that match fingerprint property names, these will flow directly to a fingerprint! Cool feature that enables the JSON only scenario....

eddynaka (approved):
:shipit:

PR #447, opened 2021-05-16
Improve match efficiency. Provide two specific examples.

michaelcfanning:
@eddynaka.

* This change resolves a basic inefficiency in what was checked in: we now preserve a single copy of any duplicated secret components in a file. 
* Renamed some things for clarity.
* Implemented two rules for intrafile analysis.


Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
```diff
@@ -126,13 +126,13 @@ private ValidationState RetrieveInformation<T>(string url, string id, ref string
 
             try
             {
-                using HttpResponseMessage httpResponse = httpClient.GetAsync(url).GetAwaiter().GetResult();
+                using HttpResponseMessage response = httpClient.GetAsync(url).GetAwaiter().GetResult();
```
michaelcfanning, line 129: >response [](http://example.com/codeflow?start=42&length=8)

let's continue to converge on consistent patterns for variable renaming etc. Also, very important: we need to rework our HTTP client creation/usage. The current model is not correct, per information provided by @cfaucon recently. I'll open an issue.

Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
```diff
@@ -147,14 +147,16 @@ private ValidationState RetrieveInformation<T>(string url, string id, ref string
                     }
 
                     case System.Net.HttpStatusCode.BadRequest:
+                    case System.Net.HttpStatusCode.InternalServerError:
```
michaelcfanning, line 150: >InternalServerError [](http://example.com/codeflow?start=51&length=19)

I see a confusing phenomenon, some ids return bad request, others return internal server error. I don't see anything obvious that would distinguish between these conditions, so we'll merge them.

Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
```diff
@@ -147,14 +147,16 @@ private ValidationState RetrieveInformation<T>(string url, string id, ref string
...
                     {
                         return ReturnUnauthorizedAccess(ref message, asset: id);
                     }
 
                     default:
                     {
-                        message = $"Unexpected response status code: '{httpResponse.StatusCode}'";
-                        return ReturnUnknownAuthorization(ref message, asset: id);
+                        message = CreateUnexpectedResponseCodeMessage(response.StatusCode, asset: id);
```
michaelcfanning, line 157: >CreateUnexpectedResponseCodeMessage [](http://example.com/codeflow?start=34&length=35)

TODO. Shouldn't we have a ReturnUnexpectedResponseCodeMessage handler? i.e., is there a reason why we wouldn't standardize the return value here, to ValidationState.Unknown?

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,15 +23,23 @@
...
 
-  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?P<scheme>http|ftp|https):\/\/(?P<host>[\w_.-]{1,200})(?P<path>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?P<secret>[^'"><;\s]{1,500})
+# Intrafile analysis
+  $SEC101/004.FacebookAppCredentialsId=(?si)facebook.{1,50}\b(?P<id>[0-9]{15})([^0-9]|$)
+  $SEC101/004.FacebookAppCredentialsSecret=(?si)facebook.{1,50}\b(?P<secret>[0-9a-f]{32})([^0-9]|$)
+
+  $SEC101/008.AwsCredentialsId=(?s)\b(?P<id>(A3T[0-9A-Z]|AKIA|AGPA|AROA|AIPA|ANPA|ANVA|ASIA)[0-9A-Z]{16})([^0-9A-Z]|$)
+  $SEC101/008.AwsCredentialsSecret=(?i)\b(?P<secret>[0-9a-z\/+]{40})([^0-9a-z\/+]|$)
+
+# Legacy analysis model : no dynamic validator
```
michaelcfanning, line 33: >Legacy [](http://example.com/codeflow?start=2&length=6)

Let's explicitly isolate and track regexes with no dynamic validator...

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,15 +23,23 @@
 
 # $FacebookAppCredentials=(?si)(((facebook|fb)(.{0,20})?[`'"\s>](?P<id>[0-9]{13,17})[^0-9].{1,500}?)|((facebook|fb)(.{0,20})?[`'"\s>](?P<secret>[0-9a-f]{32})([^0-9a-f]?)).{1,500}?){2}
 
-  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?P<scheme>http|ftp|https):\/\/(?P<host>[\w_.-]{1,200})(?P<path>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?P<secret>[^'"><;\s]{1,500})
+# Intrafile analysis
+  $SEC101/004.FacebookAppCredentialsId=(?si)facebook.{1,50}\b(?P<id>[0-9]{15})([^0-9]|$)
```
michaelcfanning, line 27: >FacebookAppCredentialsId [](http://example.com/codeflow?start=14&length=24)

Please review these reauthored regexes carefully to form an opinion on them. We should be *very* consistent in the way we author them. Note that one specific of the facebook rewrite is that we require an explicit prefix of some kind with the term 'facebook' in it before we attempt to retrieve what is presumably a secret. We do this because the pattern for the facebook secret is so generic.

michaelcfanning, line 27: >([^0-9]|$) [](http://example.com/codeflow?start=78&length=10)

please note this pattern for detecting the end of a match. we should be consistent about this or reject it as a pattern. :)

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,15 +23,23 @@
...
+  $SEC101/004.FacebookAppCredentialsSecret=(?si)facebook.{1,50}\b(?P<secret>[0-9a-f]{32})([^0-9]|$)
+
+  $SEC101/008.AwsCredentialsId=(?s)\b(?P<id>(A3T[0-9A-Z]|AKIA|AGPA|AROA|AIPA|ANPA|ANVA|ASIA)[0-9A-Z]{16})([^0-9A-Z]|$)
+  $SEC101/008.AwsCredentialsSecret=(?i)\b(?P<secret>[0-9a-z\/+]{40})([^0-9a-z\/+]|$)
+
+# Legacy analysis model : no dynamic validator
   $SEC101/002.GoogleOAuthCredentials=(?s)\b(?P<id>[0-9]{12}\-[0-9a-z]{32})\.apps\.googleusercontent\.com.{1,500}?\b(?i)(?P<secret>[0-9a-z\-]{24})
```
michaelcfanning, line 34: >\b [](http://example.com/codeflow?start=41&length=2)

we should identify a good prefix pattern to use here. \b will also match the beginning of a line and is a concise way to go. we could consider a more precise match, which requires a start line or a character which is not part of the match set (this is what we do for the end of the match). \b could be sufficient, though, and it's more readable. Thoughts?

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,15 +23,23 @@
 
 # $FacebookAppCredentials=(?si)(((facebook|fb)(.{0,20})?[`'"\s>](?P<id>[0-9]{13,17})[^0-9].{1,500}?)|((facebook|fb)(.{0,20})?[`'"\s>](?P<secret>[0-9a-f]{32})([^0-9a-f]?)).{1,500}?){2}
 
-  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?P<scheme>http|ftp|https):\/\/(?P<host>[\w_.-]{1,200})(?P<path>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?P<secret>[^'"><;\s]{1,500})
+# Intrafile analysis
+  $SEC101/004.FacebookAppCredentialsId=(?si)facebook.{1,50}\b(?P<id>[0-9]{15})([^0-9]|$)
+  $SEC101/004.FacebookAppCredentialsSecret=(?si)facebook.{1,50}\b(?P<secret>[0-9a-f]{32})([^0-9]|$)
+
+  $SEC101/008.AwsCredentialsId=(?s)\b(?P<id>(A3T[0-9A-Z]|AKIA|AGPA|AROA|AIPA|ANPA|ANVA|ASIA)[0-9A-Z]{16})([^0-9A-Z]|$)
```
michaelcfanning, line 30: >\b [](http://example.com/codeflow?start=35&length=2)

we should identify a good prefix pattern to use here. \b will also match the beginning of a line and is a concise way to go. we could consider a more precise match, which requires a start line or a character which is not part of the match set (this is what we do for the end of the match). \b could be sufficient, though, and it's more readable. Thoughts?

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -23,15 +23,23 @@
 
 # $FacebookAppCredentials=(?si)(((facebook|fb)(.{0,20})?[`'"\s>](?P<id>[0-9]{13,17})[^0-9].{1,500}?)|((facebook|fb)(.{0,20})?[`'"\s>](?P<secret>[0-9a-f]{32})([^0-9a-f]?)).{1,500}?){2}
 
-  $SEC101/001.HttpAuthorizationRequestHeader=(?i)(?P<scheme>http|ftp|https):\/\/(?P<host>[\w_.-]{1,200})(?P<path>[\w.,@?^=%&:\/~+#-]*[\w@?^=%&\/~+#-])?(.|\n){0,100}?authorization[,\[:= "']+(basic)[\s\r\n]{0,10}(?P<secret>[^'"><;\s]{1,500})
+# Intrafile analysis
```
michaelcfanning, line 26: >Intrafile [](http://example.com/codeflow?start=2&length=9)

Note, the intrafile analysis no longer uses the <refine> convention. We retrain this currently for JSON regex only scenarios, to help improve region selection.

michaelcfanning:
>         protected override IEnumerable<ValidationResult> IsValidStaticHelper(ref string matchedPattern,

Note that in general, we don't require a change to the static helper when converting to intrafile analysis. But I think it's good to use this as yet another excuse to look at the implementation and try to tinker/improve.

NOTE: we can remove the 'matchedPattern' argument at any time at this point! the 'groups' argument "0" will always contain the full match. PLEASE DON'T MAKE THIS CHANGE YET. :) Later.

---
Refers to: Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs:44 in 449cc1e. [](commit_id = 449cc1ea5dbef16861292aed6e8171e37a386eac, deletion_comment = False)

michaelcfanning:
>             var validationResult = new ValidationResult

We now have a helper to create a fingerprint from a dictionary of flex matches (if it finds a group with the same name as a fingerprint property, it will populate it). After we get through current work, we should make a single push to improve/simplify this common return value pattern. But later! not yet, please. :)

---
Refers to: Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs:53 in 449cc1e. [](commit_id = 449cc1ea5dbef16861292aed6e8171e37a386eac, deletion_comment = False)

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_004.FacebookAppCredentials.sarif
```diff
@@ -103,11 +103,11 @@
                   "uriBaseId": "SRC_ROOT"
                 },
                 "region": {
-                  "startLine": 5,
-                  "startColumn": 14,
-                  "endLine": 5,
-                  "endColumn": 46,
-                  "charOffset": 149,
+                  "startLine": 2,
+                  "startColumn": 20,
```
michaelcfanning, line 107: >20 [](http://example.com/codeflow?start=33&length=2)

Note several interesting and subtle points here. First, this test file changed (but the AWS did not) because in this code, there is a single shared secret and four unique ids. And so, every find uses the first secret region when it is located! Clearly, moving forward, we should add the regions for the unique ids as a related location. That can come later. It makes an important and useful point: to prevent combinations from exploding in a test file, it is helpful to use the convention of creating unique secrets but use a consistent id. That will create an issue that on double-click takes you to the specific test case. I am not suggesting changing this test file, though. We will definitely want specific tests that are dedicated to exploring the 'shared secret, unique id' and other analysis patterns. Until then, we can keep tests as they are. That is, unless a test file makes the findings explode due to all the combinations. In this case, we may prefer to update the tests to use unique secrets with shared other data, host, id, etc. Keep this in mind as you move forward with other intrafile conversions.

Src/Sarif.PatternMatcher/RegexMetadata.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// Indicates that the regex retrieves an optionally
+        /// available value.
+        /// </summary>
+        Optional = 0x1,
+
+        /// <summary>
+        /// Indicates that the regex is potentially shared across
+        /// multiple rules (and therefore can be cached).
+        /// </summary>
+        Shared = 0x2,
```
michaelcfanning, line 28: >Shared [](http://example.com/codeflow?start=8&length=6)

Unused for now, but there is an intended future use we have in mind. Another interesting case is to mark a regex as one that, whenever it matches, should fire a warning but only if another rule hasn't already interacted with it. 

So, for example, many rules may consume a 256 bit symmetric key. If any rule seems to think a key of this kind 'belongs' to it, great. If no rule does, the scanner might emit a warning which basically notes 'this file has something that looks like a secret but no rule recognized it as relevant to it.'

Src/Sarif.PatternMatcher/FlexMatchValueComparer.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+using Microsoft.RE2.Managed;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
+{
+    internal class FlexMatchValueComparer : IEqualityComparer<FlexMatch>
```
michaelcfanning, line 10: >FlexMatchValueComparer [](http://example.com/codeflow?start=19&length=22)

This should have come in with some unit tests, feel free to add a couple, its not much work! :)

Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
eddynaka: I like the idea :)
would facilitate in a lot of places!

Src/Sarif.PatternMatcher/FlexMatchValueComparer.cs
eddynaka: I did two major changes:
1. moved this to RE2, so people can use it and this would not depend on SarifPatternMatcher
2. added tests

Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
eddynaka: Created the PR: [https://github.com/microsoft/sarif-pattern-matcher/pull/451](https://github.com/microsoft/sarif-pattern-matcher/pull/451)

Src/Plugins/Security/SEC101_004.FacebookAppCredentialsValidator.cs
eddynaka: Created the PR:
[https://github.com/microsoft/sarif-pattern-matcher/pull/452](https://github.com/microsoft/sarif-pattern-matcher/pull/452)

Src/Plugins/Security/Security.SharedStrings.txt
eddynaka: Agreed. simplifies the regex :)

eddynaka (approved):
:shipit:

PR #451, opened 2021-05-17
Do not dispose HttpClient

### LINKED ISSUE #448: Improve HTTP client caching/utilization
@cfaucon has explained to us that our current model is not ideal.

@eddynaka

eddynaka:
Fixes #448


Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -34,6 +34,9 @@ public abstract class ValidatorBase
             new Regex($@"The underlying connection was closed: Could not establish " +
                          "trust relationship for the SSL/TLS secure channel.", s_options);
 
+        private static readonly object sync = new object();
+        private static HttpClient httpClient;
```
michaelcfanning, line 38: >httpClient [](http://example.com/codeflow?start=34&length=10)

why aren't we caching a per-thread instance of http client? with this solution, we wouldn't require locking, as all per-thread usage would be synchronous

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -275,19 +278,28 @@ internal static string ParseValue(string value)
 
         protected HttpClient CreateHttpClient()
         {
-            var httpClientHandler = new HttpClientHandler()
+            if (httpClient == null)
             {
-                AllowAutoRedirect = true,
-                MaxAutomaticRedirections = 10,
-            };
+                lock (sync)
```
michaelcfanning, line 283: >sync [](http://example.com/codeflow?start=22&length=4)

can we find a solution that doesn't throttle requests to a single one at a time?

did you test the performance impact of this change? this is one of those that really should be tested.

michaelcfanning (changes requested):
:clock1:

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
eddynaka: the idea here was to prevent multiple instances even in multithreaded. so, would lock in the first request and all  the others would be the static version.

To generate a port exaustion, we would require tons of httpclients. So, movint to threadstatic would not require locking and would also fix the current issue.

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
eddynaka: Changed to threadstatic.

PR #452, opened 2021-05-17
Unhandled response code helper

### LINKED ISSUE #449: Shouldn't we have a ReturnUnexpectedResponseCode helper?
This would return a consistent message and return value for this scenario.

eddynaka:
Fixes #449


Src/Plugins/Security/SEC101_044.NpmCredentialsValidator.cs
```diff
@@ -148,8 +148,7 @@ static NpmCredentialsValidator()
 
                     default:
                     {
-                        message = CreateUnexpectedResponseCodeMessage(responseWithCredentials.StatusCode, host);
-                        return ValidationState.Unknown;
+                        return ReturnUnexpectedResponseCode(ref message, responseWithCredentials.StatusCode, asset: host);
```
michaelcfanning, line 151: >asset: host [](http://example.com/codeflow?start=109&length=11)

what does this message look like, does it contain the id?

please look at line 141 and 146, asset details typically contain the account information: that's the compromised thing, a specific nuget account, not the host in general.

our test story is weak because we do not have a good set of mocked test cases that generate every possible string. we should build this. the result should be a SARIF file that renders all combinations of our utility-generated messages. this would make it easy to review textual improvements. 

the ideal would be to see a message for every possible return condition for each rule. that would require mocking the HTTP client on a per-rule basis to raise all status code, generate exceptions, etc. This could be a long-term project to ask for Yong/Shaopeng's help on. We'd want a clean pattern, could just start with designing a comprehensive test for one rule and then going from there.

michaelcfanning (approved):
:shipit:

PR #462, opened 2021-05-21
Singleline regexes

michaelcfanning:
@eddynaka, finally, investing in test infrastructure! we should polish these patterns until they shine. :) and add more good tests.


eddynaka (approved):
:shipit:

PR #465, opened 2021-05-22
Removing unnamed groups

eddynaka:
image

Left side = Previous
Right side = After change

Removing unnamed groups would allocate less strings, ints, bools.
Without allocating those, we would call less times GC, with that GC would affect less the runtime of the analysis


michaelcfanning (approved):
:shipit:

PR #466, opened 2021-05-23
Author single-line, multicomponent analysis.

michaelcfanning:
@eddynaka, hey, look at this! an engine change that comes in first with testing. :)

as a next step, I will re-author a single-line, multicomponent rule, wanted you to get this first look.

### DIFF
--- a/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
+++ b/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
@@ -137,6 +137,16 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 
                 foreach (MatchExpression matchExpression in definition.MatchExpressions)
                 {
+                    if (matchExpression.SingleLineRegexes?.Count > 0)
+                    {
+                        for (int i = 0; i < matchExpression.SingleLineRegexes.Count; i++)
+                        {
+                            PushData(matchExpression.SingleLineRegexes[i],
+                                         definition.SharedStrings,
+                                         sharedStrings);
+                        }
+                    }
+
                     if (matchExpression.IntrafileRegexes?.Count > 0)
                     {
                         matchExpression.IntrafileRegexMetadata =
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -423,6 +423,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
             {
                 foreach (KeyValuePair<string, FlexMatch> keyValue in groups)
                 {
+                    string key = keyValue.Key;
+
                     // We only persist named groups, not groups specified by index.
                     if (int.TryParse(keyValue.Key, out int val)) { continue; }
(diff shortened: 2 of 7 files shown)


Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -137,6 +137,16 @@ internal static SearchDefinitions PushInheritedData(SearchDefinitions definition
 
                 foreach (MatchExpression matchExpression in definition.MatchExpressions)
                 {
+                    if (matchExpression.SingleLineRegexes?.Count > 0)
+                    {
+                        for (int i = 0; i < matchExpression.SingleLineRegexes.Count; i++)
```
eddynaka, line 142: nit: replace for:
for (int i = 0; i < matchExpression.SingleLineRegexes?.Count; i++)
{
     PushData
}

nit: verify the  alignment

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -438,36 +440,44 @@ private static void MergeDictionary(IList<Dictionary<string, FlexMatch>> mergeFr
...
             else
             {
-                // Both FileNameAllowRegex and ContentRegex are null or empty.
+                throw new InvalidOperationException("Malformed expression contains no regexes.");
             }
         }
 
+        private void RunMatchExpressionForSingleLineAndIntrafileRegexes(AnalyzeContext context, MatchExpression matchExpression)
+        {
+            RunMatchExpressionForIntrafileRegexes(context, matchExpression);
```
eddynaka, line 468: >RunMatchExpressionForIntrafileRegexes [](http://example.com/codeflow?start=12&length=37)

is this intentional? with this, we would be able to run Intrafile and SingleLine

michaelcfanning: Yes. What problem do you see? Currently, you can create a single rule that has both Intrafile and SingleFile regexes. Do you think we should not permit that??

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
michaelcfanning: I fixed the alignment. I don't understand your other note. :) Don't forget, we already know because of line 140 that this member is non-null, which is why the '?' isn't used on line 142.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
eddynaka: i thought we were doing something like one or the other. :)

but, doing both will permit a lot of combinations.
it was just to confirm

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
eddynaka: similar to a change that u made in SearchSkimmer:
`for (int i = 0; i < matchExpression.IntrafileRegexes?.Count; i++)`

that would prevent the if + for.
that could be applied here as well.

eddynaka (approved):
:shipit:

PR #470, opened 2021-05-30
MySQL single line refactor

eddynaka:


Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
```diff
@@ -54,49 +51,38 @@ public static IEnumerable<ValidationResult> IsValidStatic(Dictionary<string, Fle
...
-            string database = ParseExpression(RegexEngine, groups["0"], DatabaseRegex, ref unused);
+            groups.TryGetNonEmptyValue("port", out FlexMatch port);
+            groups.TryGetNonEmptyValue("resource", out FlexMatch resource);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
-
-            if (HostsToExclude.Contains(hostValue) ||
-                hostValue.IndexOf("postgres", StringComparison.OrdinalIgnoreCase) != -1)
+            if (hostValue.IndexOf("postgres", StringComparison.OrdinalIgnoreCase) != -1 ||
+                HostsToExclude.Any(hostToExclude => hostValue.IndexOf(hostToExclude, StringComparison.OrdinalIgnoreCase) != -1))
```
eddynaka, line 65: >HostsToExclude.Any(hostToExclude => hostValue.IndexOf(hostToExclude, StringComparison.OrdinalIgnoreCase) != -1) [](http://example.com/codeflow?start=16&length=111)

we weren't filtering correctly in our previous validations.

Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
```diff
@@ -54,49 +51,38 @@ public static IEnumerable<ValidationResult> IsValidStatic(Dictionary<string, Fle
         protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<string, FlexMatch> groups)
         {
             if (!groups.TryGetNonEmptyValue("id", out FlexMatch id) ||
+                !groups.TryGetNonEmptyValue("host", out FlexMatch host) ||
```
eddynaka, line 54: >FlexMatch host [](http://example.com/codeflow?start=56&length=14)

if we don't have a host, we will return NoMatch

Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
michaelcfanning: oh, because these are components of the host rather than the actual hosts. did you add tests for this?

Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -290,7 +290,13 @@
         {
           "Id": "SEC101/036",
           "Name": "DoNotExposePlaintextSecrets/MySqlConnectionString",
-          "ContentsRegex": "$SEC101/036.MySqlConnectionStringAdo",
+          "SingleLineRegexes": [
```
michaelcfanning, line 293: >SingleLineRegexes [](http://example.com/codeflow?start=11&length=17)

One point about this model, the *first* match is critical. if it doesn't match, nothing else happens. :) and so the first match must be something that is *required* and ideally is something that is not frequent, i.e., is sufficiently distinct to help keep candidate numbers down.

Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
```diff
@@ -17,9 +18,6 @@ public class MySqlConnectionStringValidator : ValidatorBase
         internal static MySqlConnectionStringValidator Instance;
         internal static IRegex RegexEngine;
```
michaelcfanning, line 19: >RegexEngine [](http://example.com/codeflow?start=31&length=11)

Do you need this anymore?

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_036.MySqlConnectionString.sarif
```diff
@@ -155,49 +155,6 @@
...
-            {
-              "physicalLocation": {
-                "artifactLocation": {
-                  "uri": "src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_036.MySqlConnectionString.ps1",
-                  "uriBaseId": "SRC_ROOT"
-                },
-                "region": {
-                  "startLine": 3,
-                  "startColumn": 122,
-                  "endLine": 3,
```
michaelcfanning, line 183: >": [](http://example.com/codeflow?start=26&length=2)

Nice! now we don't have two candidates, one with a port and one without! In the future, you could explain this code change to me. i.e., we should always try to explain to each other why things appeared or disappeared if it's not obvious in the baseline logs.

michaelcfanning (approved):
:shipit:

Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
eddynaka: no, we don't. removing it

Src/Plugins/Security/SEC101_036.MySqlConnectionStringValidator.cs
eddynaka: we have one dropped with this change, since we will filter correctly

PR #471, opened 2021-05-30
Improve SQL region selection.

michaelcfanning:
@eddynaka


Src/Plugins/Security/SEC101_037.SqlConnectionStringValidator.cs
```diff
@@ -85,7 +85,6 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
 
             var validationResult = new ValidationResult
             {
-                RegionFlexMatch = secret,
```
michaelcfanning, line 88: >RegionFlexMatch [](http://example.com/codeflow?start=16&length=15)

this is all that it takes for a single line regexes rule to get the specialized region selection in the scanner.

eddynaka (approved):
:shipit:

PR #474, opened 2021-05-30
Replacing Guid.NewGuid for ScanId

eddynaka:


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -1243,7 +1242,7 @@ private void AddFixToResult(FlexMatch flexMatch, SimpleFix simpleFix, Result res
         {
             int argsCount = namedArgumentToIndexMap.Count;
 
-            var arguments = new List<string>(new string[argsCount]);
```
eddynaka, line 1246: >new string[argsCount] [](http://example.com/codeflow?start=45&length=21)

this was creating a string array with argsCount position, that wasn't being used.
Passing the count prevents resizing.

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -15,6 +15,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public abstract class ValidatorBase
     {
+        public const string ScanIdentityId = "01350135-0135-4135-0135-013501350135";
```
michaelcfanning, line 18: >ScanIdentityId [](http://example.com/codeflow?start=28&length=14)

You should make this a property and initialize it with a random GUID value.

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -53,9 +55,32 @@ protected ValidatorBase()
...
+                }
+
+                return scanIdentityGuid;
+            }
+
+            set
+            {
+                if (!Guid.TryParse(value, out _))
+                {
+                    throw new ArgumentException($"The '{value}' is not a GUID.");
```
michaelcfanning, line 74: >The '{value}' is not a GUID. [](http://example.com/codeflow?start=50&length=28)

'{value}' is a not a valid GUID.

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -53,9 +55,32 @@ protected ValidatorBase()
             PerFileFingerprintCache = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
         }
 
+        protected virtual string ScanIdentityGuid
+        {
+            get
+            {
+                if (string.IsNullOrEmpty(scanIdentityGuid))
```
michaelcfanning, line 62: >IsNullOrEmpty [](http://example.com/codeflow?start=27&length=13)

according to this code, you can never set scanIdentityGuid to an empty value. i.e., a null check here would be just fine. So I'd be tempted to render this code as a two-liner:

scanIdentityGuid ??= Guid.NewGuid().ToString();
return scanIdentityGuid;

michaelcfanning (approved):
:shipit:

PR #484, opened 2021-06-10
Fixing shared properties

eddynaka:


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -591,14 +591,16 @@ private void RunMatchExpressionForSingleLineRegexes(AnalyzeContext context, Matc
             {
                 matchExpression.Properties ??= new Dictionary<string, string>();
 
+                var properties = new Dictionary<string, string>(matchExpression.Properties);
+
                 string filePath = context.TargetUri.GetFilePath();
-                matchExpression.Properties["scanTargetFullPath"] = filePath;
```
eddynaka, line 595: >matchExpression. [](http://example.com/codeflow?start=16&length=16)

when analyzing in multi-thread, there were times where two different threads were changing this. When this happen, the AddProperties that is called inside `Validate` method generates an exception.

The base properties should be shared if exists.

michaelcfanning: So, the general scan architecture design pattern for this would be to create a 'context' object per-thread. 

You can see how this would be a more efficient change, i.e., rather than copying a dictionary on every call to validate (which occurs frequently), instead, we'd copy the match expression properties a single time on thread creation, and simply update the 'scanTargetFullPath' property as we did before (because it is a thread-safe operation).

michaelcfanning:
>             _matchExpressions = definition.MatchExpressions;

so, this is breaking the pattern. what's supposed to happen is that shared information flows in the context object. i.e., the skimmers generally should be thread-safe.



---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:81 in 37f3b14. [](commit_id = 37f3b142cd24160fd3b85f7f8c3d3ba8948f16bd, deletion_comment = False)

michaelcfanning:
>             foreach (MatchExpression matchExpression in _matchExpressions)

this should be coming to us via the context object.

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:176 in 37f3b14. [](commit_id = 37f3b142cd24160fd3b85f7f8c3d3ba8948f16bd, deletion_comment = False)

PR #485, opened 2021-06-11
Adding deprecated name to sarif

eddynaka:


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -66,6 +67,14 @@ public class SearchSkimmer : Skimmer<AnalyzeContext>
                 string matchExpressionMessage = matchExpression.Message;
                 matchExpression.ArgumentNameToIndexMap = GenerateIndicesForNamedArguments(ref matchExpressionMessage);
 
+                if (!string.IsNullOrEmpty(matchExpression.DeprecatedName) && _deprecatedNames == null)
+                {
+                    _deprecatedNames = new List<string>
```
michaelcfanning, line 72: >_deprecatedNames [](http://example.com/codeflow?start=20&length=16)

This data and the match expressions should be persisted to the context object passed to Analyze. We can take this change for now, but should fix this moving forwarqd.

eddynaka: I don't think we should add this to the context. and the reason for that is:

today we have:
1 skimmer with everything and  1.000 contexts, with that, only one list of matchExpressions and one list of deprecatedNames.

if we change:
1 skimmer and 1.000 contexts, with that,  each context would have the list of matchExpressions and deprecatedNames.

our memory consumption would increase a lot...

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/SearchSkimmer.cs
eddynaka: I think things/properties that do not change should be  in the skimmer and things that might change should be in the context.

for those cases, none would be in the context.

michaelcfanning:
> using Microsoft.CodeAnalysis.Sarif.Driver;

Can you add a unit test for this change? Thanks!

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:12 in ef7afa3. [](commit_id = ef7afa3e24b9d68e48e9f3fb9db224d328c137e4, deletion_comment = False)

michaelcfanning (changes requested):
:clock1:

eddynaka:
> using Microsoft.CodeAnalysis.Sarif.Driver;

just added :)

---
In reply to: [859953718](https://github.com/microsoft/sarif-pattern-matcher/pull/485#issuecomment-859953718) [](http://example.com/codeflow?ancestors=859953718,859953718)

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:12 in ef7afa3. [](commit_id = ef7afa3e24b9d68e48e9f3fb9db224d328c137e4, deletion_comment = False)

Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
```diff
@@ -294,6 +294,75 @@ public void AnalyzeCommand_PushInheritedDataTest()
             }
         }
 
+        [Fact]
+        public void AnalyzeCommand_WithDeprecatedName()
```
michaelcfanning, line 298: >AnalyzeCommand_WithDeprecatedName [](http://example.com/codeflow?start=20&length=33)

You added a test! Nice. Note, that we could have tested this capability by simply extending another test, to make things more efficient.

There is always a balance, do you repeat a lot of work, just to check a small condition. Or do you try to inject that testing into some other test. It depends on how serious/significant the capability you are testing.

For this one, I probably would have just put the deprecated name in some other test and made sure that it flowed as appropriate.

Now a very detailed test specific to deprecated names probably would make sense for a test that was targeted at a lower level, the skimmer class, say, or tests against a specific method somewhere.

michaelcfanning (approved):
:shipit:

PR #487, opened 2021-06-12
Re-mapping level to note/none

eddynaka:


michaelcfanning:
I don't think I agree with these changes. Secrets that are validated provide lots of good data on issue, such as the presence of an expired secret. Why would we drop that information entirely, as if it were a false positive? Can you make the case for me?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -908,7 +908,7 @@ private void RunMatchExpressionForFileNameRegex(AnalyzeContext context, MatchExp
 
                             case ValidationState.Expired:
                             {
-                                level = FailureLevel.Note;
+                                level = FailureLevel.None;
```
michaelcfanning, line 911: >None [](http://example.com/codeflow?start=53&length=4)

This loss of information concerns me. If we have something that conclusively was a secret, we shouldn't make it disappear entirely, like a 'no match'. We should produce an engineering note on our strategy (i.e., a one-page document).

Let's hold off on this until we have a written philosophy for handling results. Would you like to take a try at authoring something?

PR #488, opened 2021-06-15
Provide alternate JSON fingerprint

michaelcfanning:
The first incremental change towards providing full support for a JSON fingerprint. @eddynaka 

@cfaucon heads up. :)

Below is an example of the two formats, our current and the JSON form. The JSON form adds two additional characters per-property. The benefit of this form is that these fingerprints can easily be deserialized (and queried) by our data platform.

Future changes include 1) allowing fingerprints to be constructed from this new format, 2) persisting the new JSON fingerprints to SARIF (as a 'v2' value, we should continue to emit the 'v1' unless/until we switch over entirely to the new format).

~~~
[host=hostValue][id=idValue][secret=secretValue]
{"host":"hostValue","id":"idValue","secret":"secretValue"}


michaelcfanning:
>                 string expectedFingerprint = fingerprint.ToString();

Note that this round-tripping doesn't operate against the JSON form yet.

---
Refers to: Src/Test.UnitTests.Sarif.PatternMatcher.Sdk/FingerprintTests.cs:188 in 44bc512. [](commit_id = 44bc512ecef1fe34e866ed8e8763177f0ce5d660, deletion_comment = False)

eddynaka (approved):
:shipit:

PR #489, opened 2021-06-17
Fixing AssetFingerprint when used in ValidatingVisitor

eddynaka:


Src/Test.UnitTests.Sarif.PatternMatcher.Sdk/FingerprintTests.cs
```diff
@@ -386,6 +386,23 @@ public void Fingerprint_ShouldParseNormally()
...
+        public void Fingerprint_ShouldMergeNormally()
+        {
+            string previousFingerprint = "[host=host][part=part]";
+            var fingerprint = new Fingerprint
+            {
+                Id = "id",
+                Secret = "secret"
+            };
+
+            fingerprint.Merge(previousFingerprint);
```
michaelcfanning, line 399: Your test should also verify that no other values are unexpectedly non-null. :)

michaelcfanning (approved):
:shipit:

Src/Test.UnitTests.Sarif.PatternMatcher.Sdk/FingerprintTests.cs
eddynaka: thanks for the review!
with the new change, i'm checking all properties besides the one's that I already checked.

PR #490, opened 2021-06-18
Adding flag to persist path in asset fingerprint

eddynaka:


Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -26,27 +27,21 @@ public struct Fingerprint
...
+        private static readonly ReadOnlyCollection<string> s_emptyDenyList = new ReadOnlyCollection<string>(Array.Empty<string>());
         private static readonly byte[] HashKeyBytes = Encoding.UTF8.GetBytes(HashKey);
 
-        private static readonly HashSet<string> s_assetOnlyKeys =
-            new HashSet<string>(new string[]
-            {
-                PartKeyName,
-                PlatformKeyName,
-            });
+        private static readonly ReadOnlyCollection<string> s_assetOnlyKeys =
```
eddynaka, line 36: >ReadOnlyCollection [](http://example.com/codeflow?start=32&length=18)

changing to readonlycollection  to prevent changes

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -26,25 +27,17 @@ public struct Fingerprint
 
         private const char RightBracketReplacement = '\t';
         private const string HashKey = "7B2FD4B8B55B49428DBFB22C9E61D817";
-
         private const string Base64EncodingSymbolSet =
             "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 
-        private static readonly HashSet<string> s_emptyDenyList = new HashSet<string>();
+        private static readonly ReadOnlyCollection<string> s_emptyDenyList = new ReadOnlyCollection<string>(Array.Empty<string>());
```
michaelcfanning, line 33: >ReadOnlyCollection [](http://example.com/codeflow?start=32&length=18)

Suggest using ImmutableHash<string> instead.

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -106,12 +101,14 @@ private enum ParseState
 
         public string Secret { get; set; }
 
-        public string Resource { get; set; }
+        public bool IgnorePath { get; set; }
```
michaelcfanning, line 104: >IgnorePath [](http://example.com/codeflow?start=20&length=10)

oh, you need this because this is a struct and it's required for all instantiation patterns?

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
eddynaka: I thought about it. but we would need to add a new package dependency. its not a bad thing, but since we have only a few elements, the performance is similar

Src/Test.UnitTests.Sarif.PatternMatcher.Sdk/FingerprintTests.cs
```diff
@@ -421,6 +421,37 @@ public void Fingerprint_ShouldMergeNormally()
...
+            const string path = "[path=path]";
+            var fingerprint = new Fingerprint($"{id}{path}");
+            fingerprint.GetAssetFingerprintText().Should().Be($"{id}{path}");
+            fingerprint.GetValidationFingerprintText().Should().Contain(path);
+
+            fingerprint.IgnorePath = true;
+            fingerprint.GetAssetFingerprintText().Should().Be(id);
+            fingerprint.GetValidationFingerprintText().Should().Contain(path);
+
+            fingerprint = new Fingerprint
```
michaelcfanning, line 437: >Fingerprint [](http://example.com/codeflow?start=30&length=11)

please add a secret to these tests.

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
eddynaka: Not only that. if we add a field and you  use:

new fingerprint
{
 a = b
}

this won't trigger the constructor and, we can not have a constructor without initializing all parameters and without parameters

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
michaelcfanning: that's true. but now ToString can accept a list with duplicates.

it's fine, i think you made the right call for now, for the reasons you mentioned.

michaelcfanning (approved):
:shipit:

PR #492, opened 2021-06-22
Enable json fingerprint

eddynaka:


Src/Plugins/Tests.SalModernization/TestData/UpdateSalToCurrentVersion/ExpectedOutputs/SEC105_001.AnalysisAssume.sarif
```diff
@@ -292,7 +301,10 @@
           "fingerprints": {
             "AssetFingerprint/v1": "",
             "ValidationFingerprint/v1": "[secret=4D4967B311CCE348B4DC5945D87CCF8484E0FF1223A889C3D6B2A38B543F38E9]",
-            "ValidationFingerprintHash/v1": "11626548a2bba258a57c3ddd73dc6980361c0e2f0e34895104be73678ce1640b"
+            "ValidationFingerprintHash/v1": "11626548a2bba258a57c3ddd73dc6980361c0e2f0e34895104be73678ce1640b",
+            "AssetFingerprint/v2": "{}",
+            "ValidationFingerprint/v2": "{\"secret\":\"4D4967B311CCE348B4DC5945D87CCF8484E0FF1223A889C3D6B2A38B543F38E9\"}",
+            "ValidationFingerprintHash/v2": "9dc9fc7e9bdbb4af988847caf672e8aab78bff499e10c033c5024b805d2ac6d3"
```
michaelcfanning, line 307: >9dc9fc7e9bdbb4af988847caf672e8aab78bff499e10c033c5024b805d2ac6d3 [](http://example.com/codeflow?start=45&length=64)

btw - Chris and I had discussed having the validation fingerprint hash v2 *stop* hashing the property names.

so, host=a,secret=b

would hash a and b only.

this would give us a hash but also allow the property names to be changed, i.e., a bit more resilient.

what we'd want is a new ToString() method, I think, that sorts the properties in order and simply appends all the properties values together.

it seems like there could be a statistically unlikely chance of this operation resulting in a collision, though. and currently, of course, our property names have been very stable.

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,9 +19,13 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
     public class SearchSkimmer : Skimmer<AnalyzeContext>
     {
         public const string AssetFingerprint = "AssetFingerprint/v1";
-        public const string GlobalFingerprint = "GlobalFingerprint/v1";
         public const string ValidationFingerprint = "ValidationFingerprint/v1";
         public const string ValidationFingerprintHash = "ValidationFingerprintHash/v1";
+
+        public const string AssetFingerprintV2 = "AssetFingerprint/v2";
+        public const string ValidationFingerprintV2 = "ValidationFingerprint/v2";
+        public const string ValidationFingerprintHashV2 = "ValidationFingerprintHash/v2";
```
michaelcfanning, line 27: >ValidationFingerprintHash [](http://example.com/codeflow?start=59&length=25)

Ok, so thinking this through some more. First, our validation hashes aren't necessary to emit, strictly speaking. The reason is that they are simply a SHA256 hash of other data that's already in the log file. Also, the new v2 hash doesn't afford any new value, strictly speaking, i.e., the v1 hash is still perfectly sufficient for its purpose. The reason that we're replacing it is because of encoding problems parsing the unhashed version. But the hashed version works fine.

And so, what we could do here is to either not emit the v2 hash for now (i.e., continue to use the v1). Note that anyone is still free to hash the v2 validation fingerprint.

Alternately, we could go implement that new hash that omits the property names. Now *this* hash is not something that could be computed from other log data and this value *would* provide additional value, in that it would subsequently be resilient to any property name changes. If/when we have this, we could emit is as a v2 value.

Thoughts?

Src/Plugins/Tests.SalModernization/TestData/UpdateSalToCurrentVersion/ExpectedOutputs/SEC105_001.AnalysisAssume.sarif
eddynaka: I agree and disagree :D

so, let's think in the below example:
host=a,secret=b, where "a" and "b" = hash value

Now, we changed from host to host1 and secret to secret2:
host1=a,secret2=b. So, properties name changed but values are the same.

Even with that, since this is a fingerprint, the previous result and the current result wouldn't match.

So:
- I agree that would be easier to understand and reliable
- I disagree, because if we rename the property, we would not have the same thing without adding some extra logic. the sarif itself wouldn't  be similar.

If we want to do something like that I would suggest just to use the values instead:
concat hostvalue + secretValue = c hash. We  would be able to change properties and the sarif would still be valid between renames.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
eddynaka: I think we can remove the hash-v2 for now and re-think the use of values only to generate the hash as u described above and i described in the other comment :)

PR #496, opened 2021-06-30
Improving MySql and Postgres validator

eddynaka:


Src/Plugins/Security/SEC101_036.MySqlCredentialsValidator.cs
```diff
@@ -23,6 +23,7 @@ public class MySqlCredentialsValidator : ValidatorBase
             "database.windows.net",
             "database.chinacloudapi.cn",
             "postgres.database.azure.com",
+            "database.secure.windows.net",
```
eddynaka, line 26: >database.secure.windows.net [](http://example.com/codeflow?start=13&length=27)

this is Azure SQL only

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
```diff
@@ -23,6 +23,7 @@ public class PostgreSqlCredentialsValidator : ValidatorBase
             "localhost",
             "database.windows.net",
             "database.chinacloudapi.cn",
+            "database.secure.windows.net",
```
eddynaka, line 26: >database.secure.windows.net [](http://example.com/codeflow?start=13&length=27)

this is Azure SQL only

PR #497, opened 2021-07-02
HelpUri should be configurable

eddynaka:


Bpendragon (approved):
All looks good to me.

Src/Sarif.PatternMatcher/MatchExpression.cs
```diff
@@ -21,6 +21,8 @@ public class MatchExpression
 
         public string DeprecatedName { get; set; }
 
+        public string HelpUri { get; set; }
```
michaelcfanning, line 24: >HelpUri [](http://example.com/codeflow?start=22&length=7)

This is fine for now. Sometimes people provide a base URL and then a relative rule topic URL. This can reduce space and also allow for multiple versions of docs to be hosted.

BUT DON'T CHANGE YOUR CODE. :) We can make this improvement later, if it looks like a good idea. What you've done is great for now! :)

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/MatchExpression.cs
eddynaka: interesting!

PR #504, opened 2021-07-14
ValidationResult should always point to secret region

eddynaka:


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -82,7 +82,7 @@
   $SEC101/013/PemCer.CryptographicPrivateKey=(?si)-{5}BEGIN CERTIFICATE-{5}\n?(?P<secret>[^:]*?)-{5}END CERTIFICATE-{5}
   $SEC101/013/Putty.CryptographicPrivateKey=(?is)PuTTY-User-Key-File-2.+?Private-Lines:\s*[0-9]+\s*(?P<secret>.+?)Private-MAC:\s[0-9a-z]+(?:[^0-9a-d]|$)
   $SEC101/013/Pem.CryptographicPrivateKey=(?si)-{5}BEGIN (?:DSA|EC|OPENSSH|PGP|RSA|SSH2 ENCRYPTED)?\s*PRIVATE (?:KEY BLOCK|KEY)-{5}.*?(?:VERSION: [^\n]+\n)*\n?(?P<secret>[^:]*?)-{5}END (?:DSA|EC|OPENSSH|PGP|RSA|SSH2 ENCRYPTED)?\s*PRIVATE (?:KEY BLOCK|KEY)-{5}
-  $SEC101/015.AkamaiCredentials=(?si)https:\/\/(?P<resource>[\w\-\.]+)\.akamaiapis\.net.{0,150}(?:(?:client_token.{0,10}\b(?P<id>akab[\w\-]+).{0,50})|(?:access_token.{0,10}\b(?P<id>akab[\w\-]+).{0,200})|(?:(?:client_secret).{0,10}\b(?P<secret>[0-9a-z\/\+]{43}=))){3}
+  $SEC101/015.AkamaiCredentials=(?si)https:\/\/(?P<host>[\w\-\.]+)\.akamaiapis\.net.{0,150}(?:(?:client_token.{0,10}\b(?P<id>akab[\w\-]+).{0,50})|(?:access_token.{0,10}\b(?P<resource>akab[\w\-]+).{0,200})|(?:(?:client_secret).{0,10}\b(?P<secret>[0-9a-z\/\+]{43}=))){3}
```
eddynaka, line 85: >host [](http://example.com/codeflow?start=51&length=4)

with the new test, I observed that we had a wrong pattern that was not implemented correctly.

Src/Plugins/Tests.Security/SEC101.SecurePlaintextSecretsTests.cs
```diff
@@ -19,5 +19,9 @@ public SecurePlaintextSecretsTests(ITestOutputHelper outputHelper) : base(output
         [Fact]
         public void SecurePlaintextSecrets_EndToEndFunctionalTests()
             => RunAllTests();
+
+        [Fact]
+        public void SecurePlaintextSecrets_ValidateOutputFiles()
+            => ValidateOutputFiles();
```
eddynaka, line 25: >ValidateOutputFiles [](http://example.com/codeflow?start=15&length=19)

This new method will validate our expected sarif files and search for issues, for example, an empty sarif.

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -138,5 +139,43 @@ protected void RunAllTests()
...
+                if (sarifLog.Runs[0].Results == null || sarifLog.Runs[0].Results.Count == 0)
+                {
+                    sb.AppendLine($"Results from file '{fileName}' should not be empty.");
+                    continue;
+                }
+
+                //foreach (Result result in sarifLog.Runs[0].Results)
+                //{
+                //    var fingerprint = new Fingerprint(result.Fingerprints[SearchSkimmer.ValidationFingerprintV2]);
+                //    if (!result.Locations[0].PhysicalLocation.Region.Snippet.Text.StartsWith(fingerprint.Secret))
```
eddynaka, line 169: >fingerprint.Secret [](http://example.com/codeflow?start=95&length=18)

I'm still trying to figure out how to test a requirement where the secret must appear in the region.

There is one scenario that this test does not work:
1. we receive a base64 string -> we make the conversion
2. when we have a rule that brings the entire context to be analyzed in our validator

michaelcfanning: Why isn't this invariant in the actual code somewhere? Verifying this in functional tests provides limited value.

Honestly, what we should probably do is stop requiring plug-ins to populate the region override. In cases when a rule resurns a fingerprint, if 'Secret' is populated, we should consider having the tool itself set the region to that data.

The scanner itself will know if it is base64-decoding or not and could skip this step.

michaelcfanning: btw - even if we decide to have 'test only' validation, instead of building it into a post-processing step, you could add conditionally executing code deep in the scanner itself. 

Sometimes I create a static property 'TestingInProgress' that I use to conditionally control additional verifications.

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Tests.Security/SEC104.UseSecureApiTests.cs
```diff
@@ -19,5 +19,9 @@ public UseSecureApiTests(ITestOutputHelper outputHelper) : base(outputHelper)
         [Fact]
         public void UseSecureApi_EndToEndFunctionalTests()
             => RunAllTests();
+
+        [Fact]
+        public void UseSecureApi_ValidateOutputFiles()
+            => ValidateOutputFiles();
```
michaelcfanning, line 25: >ValidateOutputFiles [](http://example.com/codeflow?start=15&length=19)

How are you going to guarantee the order of test execution? i.e., guarantee that all the output files have been generated?

Shouldn't we consider running validation, if it even occurs after that fact, in RunAllTests when we produce the log file?

Src/Plugins/Tests.Security/SEC101.SecurePlaintextSecretsTests.cs
eddynaka: removed the test since i added as a following verification after the RunAllTests method.

Src/Plugins/Tests.Security/SEC104.UseSecureApiTests.cs
eddynaka: u are right.
just changed the logic.
thanks for reviewing.

Src/Plugins/Tests.Security/EndToEndTests.cs
eddynaka: changed the logic in searchskimmer to automatically use secret match if that exists.

But, if any validator still needs to overwrite, it would still work.

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -137,6 +138,35 @@ protected void RunAllTests()
             }
 
             RunTest(inputFiles, expectedOutputResourceMap, enforceNotificationsFree: true);
+
+            ValidateOutputFiles();
+        }
+
+        private void ValidateOutputFiles()
```
michaelcfanning, line 145: >ValidateOutputFiles [](http://example.com/codeflow?start=21&length=19)

This doesn't look like a high value test to me.

Also, if you drop this in, we won't be able to use this mechanism to create test cases that are designed specifically to *not* fire results. 

I suggest we drop it unless you feel strongly about the value.

eddynaka: just removed.

PR #505, opened 2021-07-20
Fix MySQL to properly handle Azure hosts.

Bpendragon:
Makes a fix so that Azure MySQL DB matches that won't be well-formed are not attempted dynamically.

### DIFF
--- a/Src/Plugins/Security/SEC101_036.MySqlCredentialsValidator.cs
+++ b/Src/Plugins/Security/SEC101_036.MySqlCredentialsValidator.cs
@@ -29,6 +29,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         private static readonly HashSet<string> AzureHosts = new HashSet<string>
         {
             "mysqldb.chinacloudapi.cn",
+            "mysql.database.azure.com",
         };
 
         static MySqlCredentialsValidator()
@@ -66,11 +67,16 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             groups.TryGetNonEmptyValue("resource", out FlexMatch resource);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
-            if (AzureHosts.Any(azHosts => hostValue.IndexOf(azHosts, StringComparison.OrdinalIgnoreCase) != -1))
+            string idValue = id.Value;
+
+            // Azure Hosts must be in the form <username>@<hostname> if the username does not contain a host name, we can't connect.
+            if (AzureHosts.Any(azHosts => hostValue.IndexOf(azHosts, StringComparison.OrdinalIgnoreCase) != -1) &&
+                              !idValue.Contains("@"))
             {
-                // continue so we can do azure check?
+                return ValidationResult.CreateNoMatch();
             }
-            else if (hostValue.IndexOf("postgres", StringComparison.OrdinalIgnoreCase) != -1 ||
+
+            if (hostValue.IndexOf("postgres", StringComparison.OrdinalIgnoreCase) != -1 ||
                 HostsToExclude.Any(hostToExclude => hostValue.IndexOf(hostToExclude, StringComparison.OrdinalIgnoreCase) != -1))
             {
                 return ValidationResult.CreateNoMatch();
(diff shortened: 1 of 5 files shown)


Src/Plugins/Security/SEC101_036.MySqlCredentialsValidator.cs
```diff
@@ -61,6 +67,15 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
             groups.TryGetNonEmptyValue("resource", out FlexMatch resource);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
+            string idValue = id.Value;
+
+            // Azure Hosts must be in the form <username>@<hostname> if the username does not contain a host name, we can't connect.
```
eddynaka, line 72: >Azure Hosts [](http://example.com/codeflow?start=15&length=11)

it's not that azurehost must be. but the userid.
also, let's break the comment in two lines

Src/Sarif.PatternMatcher/SharedUtilities.cs
```diff
@@ -9,9 +9,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public static class SharedUtilities
     {
-        public static void PopulateAssetFingerprint(string host, ref Fingerprint fingerprint)
+        public static void PopulateAssetFingerprint(string host, ref Fingerprint fingerprint, List<string> azureHosts = null)
```
eddynaka, line 12: >PopulateAssetFingerprint [](http://example.com/codeflow?start=27&length=24)

can u add some tests to this function?

eddynaka: i would also create a new method instead of this with:
list, host, out resource.

the list would not be null and it would be used  to the second method.

The second method would also receive:
list, host, out resource.

and this current method would hard-code the list.

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/Inputs/SEC101_036.MySqlCredentials.ps1
```diff
@@ -4,6 +4,9 @@ Port=3306; Server=some-database-name.mysql.database.azure.com; Database=catalog_
 Server=some-database-name.mysql.database.azure.com; Database=catalog_db; Uid=username@some-database-name; Pwd=password_3; SslMode=Preferred;Port=3306; 
 Server=some-database-name.mysql.database.azure.com;Database=catalog_db; SslMode=Preferred; Pwd=password_4; Uid=username@some-database-name;
 
+# This is a malformed and invalid ADO MySQL String. UiD should be in the form of <username>@<database-name>. This should be missed.
```
eddynaka, line 7: >This [](http://example.com/codeflow?start=2&length=4)

I would suggest adding in the end. Simple reason: we would not see a change in the sarif

Src/Test.UnitTests.Sarif.PatternMatcher/SharedUtilitiesTests.cs
```diff
@@ -32,5 +34,35 @@ public void SharedUtilities_PopulateAssetFingerprint()
...
+
+            foreach (string azureHost in azureHosts)
+            {
+                SharedUtilities.PopulateAssetFingerprint(otherAzureHosts, azureHost, ref fingerprint);
+                fingerprint.Part.Should().Be("servers");
+                fingerprint.Platform.Should().Be(nameof(AssetPlatform.Azure));
+            }
+
+            SharedUtilities.PopulateAssetFingerprint(otherAzureHosts, "some-unknown-host", ref fingerprint);
+            fingerprint.Part.Should().Be("servers");
```
eddynaka, line 64: >fingerprint.Part.Should(). [](http://example.com/codeflow?start=12&length=26)

this is wrong.
move this before the foreach and you will see that part will be empty.

also, if  one of the tests start  failing  you would have to debug and see which one is the wrong one.

take a look at our other tests and see  our pattern

eddynaka (approved):
:shipit:

Src/Test.UnitTests.Sarif.PatternMatcher/SharedUtilitiesTests.cs
eddynaka: I made a few changes:
1. created a testCase dynamic with input and expected output
2. we compare everything and report in the end
3. we are always showing which test failed.

PR #506, opened 2021-07-21
Removing false-positives from NugetCredential validator

eddynaka:


Src/Plugins/Security/SEC101_043.NuGetCredentialsValidator.cs
```diff
@@ -109,71 +111,56 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
-            string password = fingerprint.Secret;
+            string secret = fingerprint.Secret;
 
             HttpClient client = CreateHttpClient();
 
             try
             {
-                using HttpResponseMessage responseWithNoCredentials = client
-                    .GetAsync(host, HttpCompletionOption.ResponseHeadersRead)
+                ValidationState validationState = ValidateWithEmptyAndRandomKey(ref message,
```
eddynaka, line 122: >ValidateWithEmptyAndRandomKey [](http://example.com/codeflow?start=50&length=29)

this method will:
1. make a request without a key: if it receives 200, for example, it will drop the result
2. make a request with a fixed key: if it receives 200, for example, it will drop the result

eddynaka: This will make the validator slower, but it will reduce false-positives

PR #507, opened 2021-07-21
Fix PostgreSQL to properly handle Azure hosts.

Bpendragon:


Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
```diff
@@ -61,6 +67,16 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
             groups.TryGetNonEmptyValue("resource", out FlexMatch resource);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
+            string idValue = id.Value;
+
+            // Username must be in the form <username>@<hostname> to communicate with Azure.
+            // If the username does not contain a host name, we can't connect.
+            if (AzureHosts.Any(azHosts => hostValue.IndexOf(azHosts, StringComparison.OrdinalIgnoreCase) != -1) &&
```
michaelcfanning, line 74: >Any [](http://example.com/codeflow?start=27&length=3)

You're side-stepping the value of having a set here, which is the ability to call Contains on it for look-up in constant time. Also, you can initialize your set with a string comparer that makes the Contains look-up case-insensitive. You may be required to refactor your initialization strategy though, since this will entail calling a parameterized ctor.

Btw - since you are only tracking a single host, an alternate approach is simply to persist this data as a const string (and to upgrade to a hashset later, if/when additional hosts require it).

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
```diff
@@ -70,14 +86,14 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
-                Id = id.Value,
+                Id = idValue,
                 Host = hostValue,
                 Port = port?.Value,
                 Secret = secret.Value,
                 Resource = resource?.Value,
             };
 
-            SharedUtilities.PopulateAssetFingerprint(hostValue, ref fingerprint);
+            SharedUtilities.PopulateAssetFingerprint(AzureHosts.ToList(), hostValue, ref fingerprint);
```
michaelcfanning, line 96: >AzureHosts [](http://example.com/codeflow?start=53&length=10)

This confuses me, I didn't think this helper accepts a list/enumerable, but only a single value. Is this code compiling?

In any case, we wouldn't want to pass all possible values from AzureHosts, we should be passing the specific host for this finding, which is in `hostValue`. So i think you should revert the change on this line. ?

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
```diff
@@ -61,6 +67,16 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
             groups.TryGetNonEmptyValue("resource", out FlexMatch resource);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
+            string idValue = id.Value;
```
michaelcfanning, line 70: >idValue [](http://example.com/codeflow?start=19&length=7)

So you've created a variable here that simply caches id.Value. Why? :) I'm not sure there is a substantive improvement in readability that warrants adding an additional line of code. 

i.e., all this does is transform every 'id.Value' expression to 'idValue'. How does this help? Why did you choose this property to cache and not, say, secret.Value?

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
Bpendragon: Makes sense. I was trying to keep the file self-consistent. Will swap it to `.contains`

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
Bpendragon: An overload for `populateassettfingerprint` was added in #505 that takes in an additional list of hosts. (additional tests also included in that PR)

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
Bpendragon: This was cached because I needed to use the value twice in the code (lines 75 and 89, previously only needed in the latter location), so I cached it to save the round trip on reading the object.

Src/Plugins/Security/SEC101_038.PostgreSqlCredentialsValidator.cs
```diff
@@ -61,6 +64,16 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+
+            // Username must be in the form <username>@<hostname> to communicate with Azure.
+            // If the username does not contain a host name, we can't connect.
+            if (hostValue.Contains(AzureHost) &&
+                              !idValue.Contains("@"))
```
eddynaka, line 72: here u can keep everything in one line.
Our idea is:
- the length should not be too long
- the length should be ordered  by length (check like 56-58).

Bpendragon: fixed

eddynaka (approved):

PR #512, opened 2021-07-24
NuGet refinement

michaelcfanning:
This change refines the NuGet API key match to reflect the patterns encoded into the Base32 expression (due to encoding a GUID).

Every 128-bit GUID drops 6 bits in order to express certain constant values (such as one that documents the GUID version). As a result of this lack of randomness, there are some restrictions on the key expression when base32-encoded. The new pattern accounts for these.

@eddynaka


Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -61,7 +61,8 @@
   $SEC101/028.PlaintextPassword=(?i)ConvertTo-SecureString\s*(?:-[S]tring)?\s*["'](?P<secret>[^"')(?:]*)["']
   $SEC101/030.GoogleServiceAccountKeyConsoleFormat="private_key":\s*"-----BEGIN PRIVATE KEY-----\\n(?s)(?P<secret>[^"]+)(?-s)\\n-----END PRIVATE KEY-----\\n(?s).{1,200}(?-s)"client_id"\s*:\s*"(?P<id>[\w.-]+)"
   $SEC101/030.GoogleServiceAccountKeyConsoleRestFormat="name": "projects\/[\w-]+\/serviceAccounts\/[\w@.-]+\/keys\/(?P<secret>[^"]+)",
-  $SEC101/031.NuGetApiKey=\b(?P<secret>oy2[0-9a-z]{43})(?:^[0-9a-z]|$)
+  $SEC101/031.NuGetApiKey=\b(?P<secret>oy2[0-9a-z]{16}[aq][0-9a-z]{11}[eu]{1}[bdfhjlnprtvxz357][a-p][0-9a-z]{12})(?:^[0-9a-z]|$)
```
eddynaka, line 64: >{16}[aq][0-9a-z]{11}[eu]{1}[bdfhjlnprtvxz357][a-p][0-9a-z]{12} [](http://example.com/codeflow?start=50&length=62)

It looks good.
We will need to create a new test, since the previous one will be dropped (it is invalid)

eddynaka (approved):
:shipit:

PR #515, opened 2021-07-29
Add Square Credentials Dynamic Validator

Bpendragon:
* Adds a Dynamic Stripe Validator
* Adds a DynamicStripeValidatopr_Test class as well.

### DIFF
--- a/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
+++ b/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text.Json;
+
 using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk;
 using Microsoft.RE2.Managed;
 
--- a/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
+++ b/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
@@ -105,11 +105,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                         if (content.Contains(codeNotFoundMessage))
                         {
                             // Credential was valid, code was not.
-                            return ValidationState.Authorized;
+                            return ReturnAuthorizedAccess(ref message, id);
                         }
 
                         // Credential not valid.
-                        return ValidationState.Unauthorized;
+                        return ReturnUnauthorizedAccess(ref message, id);
                     }
 
                     default:
--- a/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
+++ b/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
@@ -69,19 +69,21 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             string id = fingerprint.Id;
             string secret = fingerprint.Secret;
 
+            const string codeForRequest = "123";
             const string uri = "https://connect.squareup.com/oauth2/token";
-            const string CodeNotFoundMessage = "Authorization code not found";
-            const string CodeForRequest = "123";
+            const string codeNotFoundMessage = "Authorization code not found";
 
             try
             {
                 HttpClient client = CreateHttpClient();
 
-                var dict = new Dictionary<string, string>();
-                dict.Add("grant_type", "authorization_code");
-                dict.Add("code", CodeForRequest); // Needs some value for call to work.
-                dict.Add("client_id", id);
-                dict.Add("client_secret", secret);
+                var dict = new Dictionary<string, string>()
+                {
+                    { "client_id", id },
+                    { "code", codeForRequest },
+                    { "client_secret", secret },
+                    { "grant_type", "authorization_code" },
+                };
 
                 using var request = new HttpRequestMessage(HttpMethod.Post, uri);
                 request.Content = new FormUrlEncodedContent(dict);
@@ -91,18 +93,22 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                     .GetAwaiter()
                     .GetResult();
 
-                string res = response.Content.ReadAsStringAsync().Result; // Actually runs synchronously when "Result" is called.
-
                 switch (response.StatusCode)
                 {
                     case HttpStatusCode.OK:
                     {
-                        return ValidationState.Authorized; // This should theoretically never be hit, a "code" of 123 is much too short. But if we do get a 200, it's a valid Credential.
+                        // This should theoretically never be hit, a "code" of 123 is much too short.
+                        // But if we do get a 200, it's a valid Credential.
+                        return ValidationState.Authorized;
                     }
 
                     case HttpStatusCode.Unauthorized:
                     {
-                        if (res.Contains(CodeNotFoundMessage))
+                        string content = response.Content
+                                                 .ReadAsStringAsync()
+                                                 .GetAwaiter()
+                                                 .GetResult();
+                        if (content.Contains(codeNotFoundMessage))
                         {
                             // Credential was valid, code was not.
                             return ValidationState.Authorized;
@@ -111,6 +117,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                         // Credential not valid.
                         return ValidationState.Unauthorized;
                     }
+
                     default:
                     {
                         return ReturnUnexpectedResponseCode(ref message, response.StatusCode);
@@ -121,7 +128,6 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             {
                 return ReturnUnhandledException(ref message, e);
             }
-
         }
     }
 }
--- a/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
+++ b/Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
@@ -1,8 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
-
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text.Json;
 using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk;
 using Microsoft.RE2.Managed;
 
@@ -20,6 +24,18 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         public static IEnumerable<ValidationResult> IsValidStatic(Dictionary<string, FlexMatch> groups)
         {
             return IsValidStatic(Instance, groups);
+        }
+
+        public static ValidationState IsValidDynamic(ref Fingerprint fingerprint,
+                                                     ref string message,
+                                                     Dictionary<string, string> options,
+                                                     ref ResultLevelKind resultLevelKind)
+        {
+            return IsValidDynamic(Instance,
+                                  ref fingerprint,
+                                  ref message,
+                                  options,
+                                  ref resultLevelKind);
         }
 
         protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<string, FlexMatch> groups)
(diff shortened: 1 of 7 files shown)


Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+                                                                ref string message,
+                                                                Dictionary<string, string> options,
+                                                                ref ResultLevelKind resultLevelKind)
+        {
+            string id = fingerprint.Id;
+            string secret = fingerprint.Secret;
+
+            const string uri = "https://connect.squareup.com/oauth2/token";
+            const string CodeNotFoundMessage = "Authorization code not found";
+            const string CodeForRequest = "123";
```
eddynaka, line 74: 1. order by string length
2. Code* starts with uppercase and  uri lowercase. Make everything lowercase, pls

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+
+            try
+            {
+                HttpClient client = CreateHttpClient();
+
+                var dict = new Dictionary<string, string>();
+                dict.Add("grant_type", "authorization_code");
+                dict.Add("code", CodeForRequest); // Needs some value for call to work.
+                dict.Add("client_id", id);
+                dict.Add("client_secret", secret);
```
eddynaka, line 84: you can auto-initialize and order by string length

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+
+                using var request = new HttpRequestMessage(HttpMethod.Post, uri);
+                request.Content = new FormUrlEncodedContent(dict);
+
+                using HttpResponseMessage response = client
+                    .SendAsync(request, HttpCompletionOption.ResponseHeadersRead)
+                    .GetAwaiter()
+                    .GetResult();
+
+                string res = response.Content.ReadAsStringAsync().Result; // Actually runs synchronously when "Result" is called.
```
eddynaka, line 94: >Result [](http://example.com/codeflow?start=66&length=6)

getawaiter().getresult() 

also, change to content instead of res

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+
+                switch (response.StatusCode)
+                {
+                    case HttpStatusCode.OK:
+                    {
+                        return ValidationState.Authorized; // This should theoretically never be hit, a "code" of 123 is much too short. But if we do get a 200, it's a valid Credential.
```
eddynaka, line 100: add the comments before the return and break in multiple lines

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+                    {
+                        if (res.Contains(CodeNotFoundMessage))
+                        {
+                            // Credential was valid, code was not.
+                            return ValidationState.Authorized;
+                        }
+
+                        // Credential not valid.
+                        return ValidationState.Unauthorized;
+                    }
```
eddynaka, line 113: add a new line here

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,68 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+                    {
+                        return ReturnUnexpectedResponseCode(ref message, response.StatusCode);
+                    }
+                }
+            }
+            catch (Exception e)
+            {
+                return ReturnUnhandledException(ref message, e);
+            }
+
```
eddynaka, line 124: remove extra new line

Src/Plugins/Tests.Security/Validators/SEC101_011 - SquareCredentialsValidatorTests.cs
```diff
@@ -0,0 +1,34 @@
...
+                return;
+            }
+
+            string message = null;
+            ResultLevelKind resultLevelKind = default;
+            var fingerprint = new Fingerprint(fingerprintText);
+            var keyValuePairs = new Dictionary<string, string>();
+
+            SquareCredentialsValidator.IsValidDynamic(ref fingerprint,
+                                                          ref message,
```
eddynaka, line 29: >                                                           [](http://example.com/codeflow?start=0&length=58)

align with the first argument

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
eddynaka, line 94: >res [](http://example.com/codeflow?start=23&length=3)

you can move this to the  unauthorized section to prevent  this being used in a 404 (for example)

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Fixed in iteration 3

Src/Plugins/Tests.Security/Validators/SEC101_011 - SquareCredentialsValidatorTests.cs
Bpendragon: Resolved in iteration 4

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,74 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+            string id = fingerprint.Id;
+            string secret = fingerprint.Secret;
+
+            const string codeForRequest = "123";
+            const string uri = "https://connect.squareup.com/oauth2/token";
+            const string codeNotFoundMessage = "Authorization code not found";
+
+            try
+            {
+                HttpClient client = CreateHttpClient();
```
michaelcfanning, line 78: >CreateHttpClient [](http://example.com/codeflow?start=36&length=16)

This code doesn't actually create a client, does it? i.e., we implemented a caching scheme underneath the covers?

If so, this is a bad name for an API. It should be CreateOrRetrieveCachedHttpClient or something...

eddynaka: let's rename in a new pr. but, yes, we should rename it :)

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,74 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+                    .GetResult();
+
+                switch (response.StatusCode)
+                {
+                    case HttpStatusCode.OK:
+                    {
+                        // This should theoretically never be hit, a "code" of 123 is much too short.
+                        // But if we do get a 200, it's a valid Credential.
+                        return ValidationState.Authorized;
+                    }
```
michaelcfanning, line 103: You need to have the courage of your comment's convictions here. If this code should never be hit, you should be throwing an exception or throwing an error condition. 

Since I agree with you this code shouldn't be hit, it seems like any execution through this code path, in fact, would be a bug and not flag an actual find.

In general, it doesn't help us to code defensively. Instead, aggressively fail unexpected code paths and our test process (if we are testing properly) should flush those out.

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,74 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+
+                    case HttpStatusCode.Unauthorized:
+                    {
+                        string content = response.Content
+                                                 .ReadAsStringAsync()
+                                                 .GetAwaiter()
+                                                 .GetResult();
+                        if (content.Contains(codeNotFoundMessage))
```
michaelcfanning, line 111: >codeNotFoundMessage [](http://example.com/codeflow?start=45&length=19)

please add a blank vertical line before this if condition.

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
michaelcfanning: My specific suggestion is to use unexpected response code for this code path. Because that's what it is, right?

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
```diff
@@ -43,5 +60,74 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
 
             return new[] { validationResult };
         }
+
+        protected override ValidationState IsValidDynamicHelper(ref Fingerprint fingerprint,
+                                                                ref string message,
+                                                                Dictionary<string, string> options,
+                                                                ref ResultLevelKind resultLevelKind)
+        {
+            string id = fingerprint.Id;
```
michaelcfanning, line 69: >fingerprint [](http://example.com/codeflow?start=24&length=11)

It looks like this code doesn't return any asset information, i.e., the identifier for the compromised thing. Please see other rules for examples of proper reporting. I think 'id' is the thing we should be surfacing that we are not.

Please make sure to review actual user-facing strings returned by this change.

@eddynaka, enough is enough: we need proper unit and functional testing that forces testing through the dynamic validation piece. :)

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
Bpendragon: Makes sense, just going to remove the block entirely and we can have it fall to 'Unexpected Response Code'

Src/Plugins/Security/SEC101_011.SquareCredentialsValidator.cs
michaelcfanning: We need to report the id for this code path but can delay getting the additional testing out there. But the very next thing we should do is implement an approach for this.

Basically, for every dynamic validator, we need a way to exercise/validate all possible code paths. This will be tricky/likely require a mocked HTTP client/other things to accomplish.

Bpendragon: Added asset information to retrun messages.

michaelcfanning: Can you provide an example of what the user-facing string looks like for authorized/unauthorized access? I think there's another overload for some helpers that allow you to provide a descriptor of the id, when helpful. to clarify the output. So looking at current example would help understand whether we need to do something like that.

Bpendragon: @michaelcfanning 

For successfully validating the response is 
> The compromised asset is 'a'

For a credential that doesn't validate it's
> The provided secret is not authorized to access 'a'.

In both cases 'a' is replaced with the `client_id` of the Square app that was detected. (which normally looks like `sq0idp-<alphanumeric-identifier>`

Src/Plugins/Tests.Security/Tests.Security.csproj
```diff
@@ -21,5 +25,9 @@
     <ProjectReference Include="..\..\sarif-sdk\src\Test.Utilities.Sarif\Test.Utilities.Sarif.csproj" />
   </ItemGroup>
 
-  <ProjectExtensions><VisualStudio><UserProperties /></VisualStudio></ProjectExtensions>
+  <ProjectExtensions>
```
eddynaka, line 28: >ProjectExtensions [](http://example.com/codeflow?start=3&length=17)

delete this

eddynaka (approved):
:shipit:

PR #516, opened 2021-07-30
Enabling GitHubAppCredentials dynamic validator

eddynaka:


Src/Plugins/Tests.Security/xunit.runner.json
```diff
@@ -0,0 +1,4 @@
+{
+  "$schema": "https://xunit.net/schema/current/xunit.runner.schema.json",
+  "maxParallelThreads": 1
```
eddynaka, line 3: >maxParallelThreads [](http://example.com/codeflow?start=3&length=18)

Since the tests that uses httpClient share a common variable, we will need to run once at a time to prevent getting undesired results

eddynaka: This is only needed during tests because we are injecting what we want to receive for each test

michaelcfanning:
We need a ReleaseNotes.md file at the root of the enlistment! It should make clear note of any changes that add to analysis or which drop false positives, etc.! there are many things we haven't documented properly, so we can do this in a follow-on change.

michaelcfanning (approved):
:shipit:

PR #518, opened 2021-08-01
Crc helpers

michaelcfanning:
Add a CRC algorithm + tests and switch ADO to use it.

@eddynaka


Src/.editorconfig
```diff
@@ -1,6 +1,15 @@
 # Remove the line below if you want to inherit .editorconfig settings from higher directories
 root = true
 
+# Ignore paths
+[/sarif-pattern-matcher/Src/sarif-sdk/src/sarif/Autogenerated/**]
```
michaelcfanning, line 5: >matcher [](http://example.com/codeflow?start=16&length=7)

I'm having trouble making sure this folder isn't analyzed in the IDE. Do you know whether this should be working?

Src/Plugins/Security/Security.csproj
```diff
@@ -5,7 +5,7 @@
   <Import Project="$(MSBuildThisFileDirectory)..\..\..\Targets\build.plugins.props" />
 
   <PropertyGroup Label="AssemblyAttributes">
-    <TargetFrameworks>netstandard2.0;net472</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;net472</TargetFrameworks>
```
michaelcfanning, line 8: >1 [](http://example.com/codeflow?start=35&length=1)

Do you know why this plug is inconsistent with other MS compilation settings?

eddynaka: we started with 2.0, but the other package got a higher dependency that required 2.1.

eddynaka: In the security.tests test, can you change to netstandard2.1? (the strings in the unit tests)
that will probably fix the build issue

Src/.editorconfig
eddynaka: what are you trying to trigger here?
for this to work we would need to change the sarif-sdk and place a file like this in the autogenerated folder.

michaelcfanning: The IDE is showing analysis errors for the Autogenerated folder.

Src/Plugins/Security/Security.csproj
eddynaka: fixed.

eddynaka (approved):
:shipit:

PR #519, opened 2021-08-01
Removing parallel tests constraints

eddynaka:


eddynaka:
I will think about it later.
still facing some concurrency issues in the tests.

PR #524, opened 2021-08-03
Create Mock Http tests for Slack Tokens

Bpendragon:
* Converts `SlackTokenValidator` from using `WebClient` to `HttpClient`
* Adds Mock Http calls to `SlackTokenValidatorTests` to properly test all code paths.

### DIFF
--- a/Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
+++ b/Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
@@ -91,10 +91,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                     return ReturnUnexpectedResponseCode(ref message, response.StatusCode);
                 }
 
-                string content = response.Content
-                                         .ReadAsStringAsync()
-                                         .GetAwaiter()
-                                         .GetResult();
+                string content = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
 
                 AuthTestResponse authResponse = JsonSerializer.Deserialize<AuthTestResponse>(content);
 
@@ -108,7 +105,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                 if (!string.IsNullOrEmpty(authResponse.Error))
                 {
                     message = $"An unexpected error was observed " +
-                              $"attempting to validate token: '{authResponse.Error}'";
+                              $"attempting to validate the token: '{authResponse.Error}'";
                     return ValidationState.Unknown;
                 }
 
--- a/Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
+++ b/Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
@@ -102,7 +102,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                 {
                     case "token_revoked":
                     case "account_inactive": { return ValidationState.Expired; }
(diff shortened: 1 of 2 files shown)


Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
```diff
@@ -66,40 +68,52 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
+                using HttpResponseMessage response = client
+                    .SendAsync(request, HttpCompletionOption.ResponseHeadersRead)
+                    .GetAwaiter()
+                    .GetResult();
 
-            if (json.Contains("invalid_auth"))
-            {
-                return ValidationState.Unauthorized;
-            }
+                string content = response.Content
```
eddynaka, line 89: >response [](http://example.com/codeflow?start=33&length=8)

should we check if statuscode == okk before checking the content?

Src/Plugins/Tests.Security/Validators/SEC101_005.SlackTokenValidatorTests.cs
```diff
@@ -30,5 +35,106 @@ public void SlackTokenValidator_Test()
...
+                                                    "\"error\": \"account_inactive\"}",
+                                                    Encoding.UTF8,
+                                                    "application/json").As<HttpContent>(),
+                    ExpectedValidationState = ValidationState.Expired,
+                    ExpectedMessage = string.Empty
+                },
+                new
+                {
+                    Title = "Testing Unknown Slack Error",
+                    HttpStatusCode = HttpStatusCode.OK,
```
eddynaka, line 98: >HttpStatusCode.OK [](http://example.com/codeflow?start=37&length=17)

let's add badrequest ->  httpcontent = null -> and save as unexpectedstatuscode

Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
Bpendragon: Not necessary. This (and most other) Slack endpoints only return a 200 status code. You may hit a 5XX if Slack itself is having issues, but if the service is up 200 is the only thing you'll get.

Bpendragon: Added a quick tet for non-200 return codes just in case we get a 500 for something.

Src/Plugins/Tests.Security/Validators/SEC101_005.SlackTokenValidatorTests.cs
Bpendragon: done.

Src/Plugins/Security/SEC101_005.SlackTokenValidator.cs
```diff
@@ -66,40 +68,57 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(Dictionary<
...
-                case "invalid_auth": { return ValidationState.Unauthorized; }
-            }
+                AuthTestResponse authResponse = JsonSerializer.Deserialize<AuthTestResponse>(content);
 
-            if (!string.IsNullOrEmpty(response.Error))
+                switch (authResponse.Error)
+                {
+                    case "token_revoked":
+                    case "account_inactive": { return ValidationState.Expired; }
+                    case "invalid_auth": { return ReturnUnauthorizedAccess(ref message, fingerprint.Secret); }
```
eddynaka, line 105: >ReturnUnauthorizedAccess [](http://example.com/codeflow?start=50&length=24)

dont do this. use validationState.unauthorized. If you do this, you are going to save the secret in the message.

eddynaka (approved):
:shipit:

PR #526, opened 2021-08-05
Making HttpClient static again when not using in tests

eddynaka:
@michaelcfanning , this is the change we talked yesterday at night.
We will have two httpclients, one that is static and one private.
The static will be normally used.
The private will **only** be used during tests.

### DIFF
--- a/Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
@@ -325,7 +325,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
             httpClient = client;
         }
 
-        protected HttpClient CreateOrUseCachedHttpClient()
+        protected HttpClient CreateOrRetrieveCachedHttpClient()
         {
             // The httpClient is the property that will be used for tests only.
             if (httpClient != null)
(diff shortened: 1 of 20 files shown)


Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -324,24 +327,30 @@ internal void SetHttpClient(HttpClient client)
 
         protected HttpClient CreateOrUseCachedHttpClient()
         {
-            if (httpClient == null)
+            // The httpClient is the property that will be used for tests only.
+            if (httpClient != null)
+            {
```
michaelcfanning, line 332: shouldn't we be using this. as a prefix here? to make it clear we're referencing instance data on the class?

i like this pattern, *or* we could use m_ as a prefix to make that distinction clear. what do we do elsewhere?

Src/Sarif.PatternMatcher.Sdk/ValidatorBase.cs
```diff
@@ -324,24 +327,30 @@ internal void SetHttpClient(HttpClient client)
 
         protected HttpClient CreateOrUseCachedHttpClient()
```
michaelcfanning, line 328: >CreateOrUseCachedHttpClient [](http://example.com/codeflow?start=29&length=27)

Change this name to CreateOrRetrieveCachedHttpClient

michaelcfanning (approved):
:shipit:

PR #530, opened 2021-08-10
Exposing method to clear cache

eddynaka:


Src/Plugins/Security/SEC101_046.DiscordApiCredentialsValidator.cs
```diff
@@ -24,6 +22,11 @@ static DiscordApiCredentialsValidator()
             Instance = new DiscordApiCredentialsValidator();
         }
 
+        public static void Clear()
```
michaelcfanning, line 25: >Clear [](http://example.com/codeflow?start=27&length=5)

Hello! This change is not what we discussed. You have added a new method. What I suggested what that you obtain a reference to the Instance property and null that out. Also, I suggested that we delete the static ctor and initialize this data in the IsValid helpers.

Src/Sarif.PatternMatcher/ValidationMethods.cs
```diff
@@ -7,6 +7,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public class ValidationMethods
     {
+        public MethodInfo Clear { get; set; }
```
michaelcfanning, line 10: >MethodInfo [](http://example.com/codeflow?start=15&length=10)

Change this to a PropertyInfo named 'Instance'.

Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -417,6 +417,8 @@ internal static FlexMatch ConstructRegionFromGroups(IDictionary<string, FlexMatc
                             isValidDynamic = null;
                         }
 
+                        MethodInfo clear = type.GetMethod("Clear");
```
michaelcfanning, line 420: >GetMethod [](http://example.com/codeflow?start=48&length=9)

type.GetProperty("Instance"). You might need the non public binding flags...

michaelcfanning, line 420: >GetMethod [](http://example.com/codeflow?start=48&length=9)

Looks like you didn't complete the end to end of this change? For this sort of thing, an API change, I'd suggest doing that. What I'd do is plumb the complete API change through with a single validator, then if I liked it, would apply the change mechanically to everything else.

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/SEC101_046.DiscordApiCredentialsValidator.cs
eddynaka: yeah, i remember what we discussed...but, below my reasoning to not apply:

```
I was implementing the idea of static instance to be settable to null using reflection.

The issue: imagine a multi-threaded application, each thread can be in a different validation moment (let's just imagine that we have only one validator).

In one of the threads i decide to set to null.

This can generate multiple null references, because we are using the instance.dictionary or instance.hashseet or instance.validator
```

eddynaka:
Closing this, since we came up with a better idea

PR #539, opened 2021-08-15
Removing redundant TryGetNonEmptyValue when key is required

eddynaka:
## Changes

* [x] Added unit tests

### DIFF
--- a/Src/Plugins/Tests.Security/Validators/SEC101_102.AdoPatValidatorTests.cs
+++ b/Src/Plugins/Tests.Security/Validators/SEC101_102.AdoPatValidatorTests.cs
@@ -15,7 +15,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 {
     public class AdoPatValidatorTests
     {
-        internal class TestCase
+        private struct TestCase
         {
             public string Title;
             public string Input;
@@ -24,7 +24,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
             public string FailureLevel;
         }
 
-        internal static TestCase[] s_coreTestCases = new[]
+        private static readonly TestCase[] s_coreTestCases = new[]
         {
             new TestCase
             {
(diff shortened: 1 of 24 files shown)


michaelcfanning:
Eddy, where is the release note describing this change? No more changes of this kind without documenting them in the release notes, please. :)

Note that this change is actually key to document as we will now aggressively raise exceptions in cases where we previously did not!

michaelcfanning:
>                         Verify(fingerprintText == $"[pat/vs={testCase.Input}]", title, failedTestCases);

this shouldn't work! this is a very old fingerprint. does this test need updating?

---
Refers to: Src/Plugins/Tests.Security/Validators/SEC101_102.AdoPatValidatorTests.cs:76 in c0b866e. [](commit_id = c0b866e3e624e83167fa3b6bbaa5a292475672eb, deletion_comment = False)

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Tests.Security/Validators/SEC101_102.AdoPatValidatorTests.cs
```diff
@@ -53,16 +53,17 @@ public void AzureDevOpsPersonalAccessTokenValidator_IsValidBasic()
                 bool performDynamicValidation = testCase.PerformDynamicValidation;
                 string failureLevel = testCase.FailureLevel;
                 string fingerprintText = null;
-                var groups = new Dictionary<string, FlexMatch>();
+                var groups = new Dictionary<string, FlexMatch>
```
michaelcfanning, line 56: >groups [](http://example.com/codeflow?start=20&length=6)

why didn't we notice this problem before? looks like we're perhaps missing a key unit test somewhere?

eddynaka: This was a strange test.
I removed a few things and started validating the expected count and state.

eddynaka:
>                         Verify(fingerprintText == $"[pat/vs={testCase.Input}]", title, failedTestCases);

This isn't needed. just removed.

---
In reply to: [899106459](https://github.com/microsoft/sarif-pattern-matcher/pull/539#issuecomment-899106459) [](http://example.com/codeflow?ancestors=899106459,899106459)

---
Refers to: Src/Plugins/Tests.Security/Validators/SEC101_102.AdoPatValidatorTests.cs:76 in c0b866e. [](commit_id = c0b866e3e624e83167fa3b6bbaa5a292475672eb, deletion_comment = False)

eddynaka:
Added a new entry

---
In reply to: [899106233](https://github.com/microsoft/sarif-pattern-matcher/pull/539#issuecomment-899106233) [](http://example.com/codeflow?ancestors=899106233,899106233)

Src/ReleaseHistory.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+- Plugin Improvement: Required properties will throw exception if it does not exist. [#539](https://github.com/microsoft/sarif-pattern-matcher/pull/539)
```
michaelcfanning, line 5: >it does [](http://example.com/codeflow?start=66&length=7)

'if they do not exist'.

michaelcfanning (approved):
:shipit:

PR #548, opened 2021-09-03
Improving RabbitMQ regex (removing new lines and spaces) from secret

eddynaka:
## Changes

* [ ] Added unit tests

### DIFF
--- a/Src/Plugins/Security/Security.SharedStrings.txt
+++ b/Src/Plugins/Security/Security.SharedStrings.txt
@@ -123,7 +123,7 @@
   $SEC101/038.PostgreSqlCredentialsAdoSecret=(?i)(?:password|pwd)\s*=\s*(?P<secret>[^,;"'<\s\n]{8,128})(?:[,;"'<\s\n]|$)
   $SEC101/038.PostgreSqlCredentialsAdoResource=(?i)(?:database|db|dbname)\s*=\s*(?P<resource>[^,;"'<\s\n]+)(?:[,;"'<\s\n]|$)
 
-  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:]+):(?P<secret>[^@\s\n]+)@(?P<host>[\w_\-\:]+)\/(?P<resource>[\w]+)(?:[^0-9a-z]|$)
+  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:]+):(?P<secret>[^@\s]+)@(?P<host>[\w_\-\:]+)\/(?P<resource>[\w]+)(?:[^0-9a-z]|$)
   $SEC101/043.NuGetPackageSourceCredentialsXml=(?i)(?P<host><\s*packageSources\s*>(?s).{0,500}?(?-s)<\\?\/packageSources\s*>)(?s).{0,200}?(?-s)[^\/](?P<secret><\s*packageSourceCredentials\s*>(?s).{0,500}?(?-s)<\\?\/packageSourceCredentials\s*>)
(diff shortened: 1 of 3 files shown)


Src/ReleaseHistory.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+- FPC: Improving RabbitMQ regex (removing new lines and spaces) from secret.
```
michaelcfanning, line 5: >FPC [](http://example.com/codeflow?start=2&length=3)

Where's the explanatory table for this code? Can you put a copy of it at the top of the table?

michaelcfanning (changes requested):
:clock1:

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -123,7 +123,7 @@
   $SEC101/038.PostgreSqlCredentialsAdoSecret=(?i)(?:password|pwd)\s*=\s*(?P<secret>[^,;"'<\s\n]{8,128})(?:[,;"'<\s\n]|$)
   $SEC101/038.PostgreSqlCredentialsAdoResource=(?i)(?:database|db|dbname)\s*=\s*(?P<resource>[^,;"'<\s\n]+)(?:[,;"'<\s\n]|$)
 
-  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:]+):(?P<secret>[^@]+)@(?P<host>[\w_\-\:]+)\/(?P<resource>[\w]+)(?:[^0-9a-z]|$)
+  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:]+):(?P<secret>[^@\s\n]+)@(?P<host>[\w_\-\:]+)\/(?P<resource>[\w]+)(?:[^0-9a-z]|$)
```
michaelcfanning, line 126: >RabbitMqCredentials [](http://example.com/codeflow?start=14&length=19)

Where's your test? Please don't submit changes to any patterns without a test! :) That is not our standard for the project.

michaelcfanning, line 126: >secret [](http://example.com/codeflow?start=67&length=6)

Can we please take a change to the fingerprint class where it throws any exception if anyone introduces newline characters into it? generally, that shouldn't happen, right? this will raise visibility on this class of defect.

remember, spot-fixing a single bug is good. taking a change that allows you to identify every instance of this bug now and moving forward is much better.

michaelcfanning, line 126: >\n [](http://example.com/codeflow?start=79&length=2)

\n is already implied by \s isn't it?

i think that's the case. if it *weren't*. i think your change would be incomplete without adding \r as well. 

But i think \s is sufficient

Src/ReleaseHistory.md
eddynaka: done!

Src/Plugins/Security/Security.SharedStrings.txt
eddynaka: just synced my environments and pushed two :)

eddynaka: I rather do this in a separate PR because that can break many other things.
Also, i was just thinking about that, like where should we throw the exception.

eddynaka: you are write. \s already includes \n.

eddynaka: I will also create a new new pr refactoring the rules that contains \n and \s.

michaelcfanning (approved):
:shipit:

PR #561, opened 2021-09-27
Updating MongoDB packages

eddynaka:
## Changes
- This will fix [CVE-2021-39208](https://dev.azure.com/mseng/1es/_componentGovernance/670/alert/70403?typeId=96972)

* [ ] Added unit tests

### DIFF
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -27,10 +27,11 @@
     <PackageReference Include="aliyun-net-sdk-core" Version="1.5.10" />
     <PackageReference Include="AWSSDK.IdentityManagement" Version="3.5.0.61" />
     <PackageReference Include="GoogleApi" Version="3.10.5" />
-    <PackageReference Include="MongoDB.Driver" Version="2.12.0" />
+    <PackageReference Include="MongoDB.Driver" Version="2.13.1" />
     <PackageReference Include="MySqlConnector" Version="1.2.1" />
     <PackageReference Include="Npgsql" Version="5.0.3" />
     <PackageReference Include="RabbitMQ.Client" Version="6.2.1" />
+    <PackageReference Include="SharpCompress" Version="0.29.0" />
     <PackageReference Include="System.Data.SqlClient" Version="4.7.0" />
     <PackageReference Include="System.Text.Json" Version="5.0.1" />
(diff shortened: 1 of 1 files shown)


Src/Plugins/Security/Security.csproj
```diff
@@ -27,10 +27,11 @@
     <PackageReference Include="aliyun-net-sdk-core" Version="1.5.10" />
     <PackageReference Include="AWSSDK.IdentityManagement" Version="3.5.0.61" />
     <PackageReference Include="GoogleApi" Version="3.10.5" />
-    <PackageReference Include="MongoDB.Driver" Version="2.12.0" />
+    <PackageReference Include="MongoDB.Driver" Version="2.13.1" />
     <PackageReference Include="MySqlConnector" Version="1.2.1" />
     <PackageReference Include="Npgsql" Version="5.0.3" />
     <PackageReference Include="RabbitMQ.Client" Version="6.2.1" />
+    <PackageReference Include="SharpCompress" Version="0.29.0" />
```
JamieMagee, line 34: You don't need to add it as a direct dependency if `MongoDB.Driver` is the only thing using it.

eddynaka: MongoDB.Driver is still using version 0.23, it's why I have to force update it.

JamieMagee: Okay, can make 2 requests:
1. Open an issue/pull request upstream to `MongoDB.Driver` so that other consumers of that package also get the benefit: https://github.com/mongodb/mongo-csharp-driver/blob/dbdac7d0228808a153278f265946c10ed0f6c1d9/src/MongoDB.Driver.Core/MongoDB.Driver.Core.csproj#L179
2. Move the `SharpCompress` dependency to a different `ItemGroup` so that when you come back to it, it's clear that this is only for component governance:

	```XML
	<ItemGroup Label="Component Governance fixes">
	    <!-- MongoDB.Driver -->
    	<PackageReference Include="SharpCompress" Version="0.29.0" />
	  </ItemGroup>
	```

JamieMagee (approved):

Src/Plugins/Security/Security.csproj
eddynaka: They already have an open issue: https://jira.mongodb.org/browse/CSHARP-3869

also, I applied your suggestion.
thanks!

PR #587, opened 2021-11-30
Adding telegram bot token validator

eddynaka:
## Changes
This is a new rule which will look for telegram bot tokens. If any match is found, the dynamic validation will be able to check if this is still live or not, updating the `finngerprint.id` with the username of the bot if available.

* [x] Added unit tests

### DIFF
--- a/Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
+++ b/Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
@@ -467,7 +467,8 @@
           "Id": "SEC101/049",
           "Name": "DoNotExposePlaintextSecrets/TelegramBotToken",
           "ContentsRegex": "$SEC101/049.TelegramBotToken",
-          "MessageArguments": { "secretKind": "Telegram bot token" }
+          "MessageArguments": { "secretKind": "Telegram bot token" },
+          "HelpUri": "https://core.telegram.org/bots#generating-an-authentication-token"
         },
         {
           "Id": "SEC101/102",
--- a/Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
+++ b/Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
@@ -464,6 +464,12 @@
           "MessageArguments": { "secretKind": "Crates API key" }
         },
         {
+          "Id": "SEC101/049",
+          "Name": "DoNotExposePlaintextSecrets/TelegramBotToken",
+          "ContentsRegex": "$SEC101/049.TelegramBotToken",
+          "MessageArguments": { "secretKind": "Telegram bot token" }
+        },
+        {
           "Id": "SEC101/102",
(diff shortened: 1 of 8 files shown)


Src/Plugins/Security/SEC101.SecurePlaintextSecrets.json
```diff
@@ -463,6 +463,13 @@
           "ContentsRegex": "$SEC101/047.CratesApiKey",
           "MessageArguments": { "secretKind": "Crates API key" }
         },
+        {
+          "Id": "SEC101/049",
```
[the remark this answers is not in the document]
eddynaka: it is the other open pr

Src/Plugins/Security/SEC101_049.TelegramBotTokenValidator.cs
```diff
@@ -0,0 +1,111 @@
...
+
+                        fingerprint.Id = resultRoot?.Result?.Username;
+                        message = $"The compromised Telegram bot account is '{fingerprint.Id}'.";
+                        return ValidationState.Authorized;
+                    }
+
+                    case HttpStatusCode.Forbidden:
+                    case HttpStatusCode.Unauthorized:
+                    {
+                        return ValidationState.Unauthorized;
```
michaelcfanning, line 72: >ValidationState.Unauthorized [](http://example.com/codeflow?start=31&length=28)

Isn't there a helper for returning this condition?

Src/Plugins/Security/SEC101_049.TelegramBotTokenValidator.cs
```diff
@@ -0,0 +1,111 @@
...
+                    .GetResult();
+
+                switch (response.StatusCode)
+                {
+                    case HttpStatusCode.OK:
+                    {
+                        string content = response.Content.ReadAsStringAsync().GetAwaiter().GetResult();
+                        ResultRoot resultRoot = JsonConvert.DeserializeObject<ResultRoot>(content);
+
+                        fingerprint.Id = resultRoot?.Result?.Username;
```
michaelcfanning, line 64: >fingerprint.Id [](http://example.com/codeflow?start=24&length=14)

Why are you updating the ID? by changing the fingerprint, you will break matches against the static analysis only phase, won't you?

Src/Plugins/Security/SEC101_049.TelegramBotTokenValidator.cs
eddynaka: we have, but we require an id, for example. in this case, this does not apply.

Src/Plugins/Security/SEC101_049.TelegramBotTokenValidator.cs
eddynaka: in the cosmos analysis, it would change only the asset fingerprint.

in the client analysis, yes, it would change both.

michaelcfanning (approved):
:shipit:

PR #589, opened 2022-01-04
Removing dummy test from HttpAuthorizationRequestHeaderValidatorTests

eddynaka:
## Changes

Removing dummy test that was actually trying to connect and validate `www.host.com`. Since we have mock tests, this isn`t needed.

* [ ] Added unit tests

### DIFF
--- a/Src/Plugins/Tests.Security/Validators/SEC101_001.HttpAuthorizationRequestHeaderValidatorTests.cs
+++ b/Src/Plugins/Tests.Security/Validators/SEC101_001.HttpAuthorizationRequestHeaderValidatorTests.cs
@@ -26,25 +26,6 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.Validator
         private const string TestKey = "somekey";
         private const string TestHost = "www.host.com";
         private const string TestResource = "/some-path";
-        private const ValidationState ExpectedValidationState = ValidationState.NoMatch;
-
-        [Fact]
-        public void HttpAuthorizationRequestHeaderValidator_Test()
-        {
-            string fingerprintText = string.Format("[host={0}][resource={1}][scheme={2}][secret={3}]", TestHost, TestResource, TestScheme, TestKey);
-
-            string message = null;
-            ResultLevelKind resultLevelKind = default;
-            var fingerprint = new Fingerprint(fingerprintText);
-            var keyValuePairs = new Dictionary<string, string>();
-
-            var httpAuthorizationRequestHeaderValidator = new HttpAuthorizationRequestHeaderValidator();
-            ValidationState actualValidationState = httpAuthorizationRequestHeaderValidator.IsValidDynamic(ref fingerprint,
-                                                                                                           ref message,
-                                                                                                           keyValuePairs,
-                                                                                                           ref resultLevelKind);
-            Assert.Equal(ExpectedValidationState, actualValidationState);
-        }
 
         [Fact]
         public void HttpAuthorizationRequestHeaderValidator_MockHttpTests()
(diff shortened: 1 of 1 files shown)


Src/Plugins/Tests.Security/Validators/SEC101_001.HttpAuthorizationRequestHeaderValidatorTests.cs
```diff
@@ -26,25 +26,6 @@ public class HttpAuthorizationRequestHeaderValidatorTests
         private const string TestKey = "somekey";
         private const string TestHost = "www.host.com";
         private const string TestResource = "/some-path";
-        private const ValidationState ExpectedValidationState = ValidationState.NoMatch;
-
-        [Fact]
-        public void HttpAuthorizationRequestHeaderValidator_Test()
```
eddynaka, line 32: >HttpAuthorizationRequestHeaderValidator_Test [](http://example.com/codeflow?start=20&length=44)

This test was trying to connect at www.host.com and verifying if that validates with the key `somekey`. Since this isn't bringing any value, i'm removing it.

Src/Plugins/Tests.Security/Validators/SEC101_001.HttpAuthorizationRequestHeaderValidatorTests.cs
```diff
@@ -140,7 +121,7 @@ public void HttpAuthorizationRequestHeaderValidator_MockHttpTests()
 
             var sb = new StringBuilder();
             var mockHandler = new HttpMockHelper();
-            var httpAuthorizationRequestHeaderValidator = new HttpAuthorizationRequestHeaderValidator();
```
eddynaka, line 143: >httpAuthorizationRequestHeaderValidator [](http://example.com/codeflow?start=16&length=39)

renaming to simplify the indentation.

eddynaka:
I disabled the runner pipeline. Starting this year, the pipeline isn't running correctly for public and internal repositories.
I'm still investigating why it is happening.

Src/Plugins/Tests.Security/Validators/SEC101_001.HttpAuthorizationRequestHeaderValidatorTests.cs
marmegh: If we still don't undertsand the root cause, should we be deleting this entirely at this point? Or disable to attempt to unblock the pipeline?

eddynaka: since we were testing against a real site, it is expected to change from time to time. We don't need a test that we dont know for sure what is the expected result.

it's why i deleted it.

michaelcfanning: It's a good idea in general not to reach across the network with key unit tests, so I think this is probably a good idea. One hint, if you *do* need a host to test against, example.com can be useful. It is intended for allocating (non-functional) URLs, for illustrative purposes (i.e., to serve as examples).

michaelcfanning (approved):
:shipit:

marmegh (approved):
:shipit:

michaelcfanning:
Got it. We should open an issue somewhere on this. This is a high priority problem, we definitely don't want to lose the protection of our CI/CD pipeline. I will force merge this change but would prefer not to have to do that again. :)

---
In reply to: [1005150870](https://github.com/microsoft/sarif-pattern-matcher/pull/589#issuecomment-1005150870) [](http://example.com/codeflow?ancestors=1005150870,1005150870)

PR #595, opened 2022-01-27
Is valid static caching thread safety

michaelcfanning:
Our static analysis validation phase contains a per-rule caching mechanism that is intended to help prevent multiple scan results for the same unique secret firing against the same file. This pattern can occur due to poorly behaved regexes and so the rules, in this case, report the first occurrence only (and drop the remainder).

The backing data structure, a HashSet<string> isn't thread-safe, and so in complex analysis scenarios we see exceptions in this code path: an 'InvalidOperationException' (in cases where HashSet.Contains provoked the issue) or 'IndexOutOfRange' (in cases where HashSet,Add provoked the problem).

The solution is to convert to using a ConcurrentDictionary instance (as .NET has no built-in thread-safe HashSet collection). This is a bit idiosyncratic, but there's no strong driving need to create a custom type (or accept a new NuGet 3rd party dependency) to get a cleaner solution. Simply using ConcurrentDictionary as a set of key is sufficient for now.

To fix this bug, I first authored a multithreaded test to provoke the issue, which now passes. This test adds ~500 ms of execution on my machine (and fails ~50% of the time without the fix).

@EasyRhinoMSFT @marmegh @shaopeng-gh @yongyan-gh @eddynaka


Src/ReleaseHistory.md
```diff
@@ -18,6 +18,8 @@
 
 ## *Unreleased*
 
+- BUG: Resolve `InvalidOperationException` and `IndexOutOfRange` exceptions in `StaticValidatorBase.IsValidStatic` due to unsafe use of HashSet<string> class.
+  [#595](https://github.com/microsoft/sarif-pattern-matcher/pull/585)
```
shaopeng-gh, line 22: >585 [](http://example.com/codeflow?start=65&length=3)

wrong number

Src/Sarif.PatternMatcher.Sdk/IDictionary.cs
```diff
@@ -0,0 +1,9 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
+{
+    public interface IDictionary<T>
```
shaopeng-gh, line 6: >IDictionary [](http://example.com/codeflow?start=21&length=11)

is this file needed

Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs
```diff
@@ -23,13 +23,13 @@ public IEnumerable<ValidationResult> IsValidStatic(IDictionary<string, FlexMatch
...
 
-                if (PerFileFingerprintCache.Contains(key))
+                if (PerFileFingerprintCache.ContainsKey(key))
                 {
                     validationResult.ValidationState = ValidationState.NoMatch;
                     continue;
                 }
 
-                PerFileFingerprintCache.Add(key);
+                PerFileFingerprintCache.Add(key, (byte)0);
```
shaopeng-gh, line 32: >(byte)0 [](http://example.com/codeflow?start=49&length=7)

also can use default(byte)

Src/Test.UnitTests.Sarif.PatternMatcher.Sdk/StaticValidatorBaseTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            {
+                Interlocked.Increment(ref id);
+
+                for (int iterations = 0; iterations < 100; iterations++)
+                {
+                    var groups = new Dictionary<string, FlexMatch>
+                {
+                    { "scanTargetFullPath", new FlexMatch { Value = Guid.NewGuid().ToString()} },
+                    { "0", new FlexMatch { Value = Guid.NewGuid().ToString() } }
+                };
```
shaopeng-gh, line 46: is the alignment correct here or need to indent

shaopeng-gh (approved):
:shipit:

Src/ReleaseHistory.md
```diff
@@ -18,6 +18,8 @@
 
 ## *Unreleased*
```
shaopeng-gh, line 19: >Unreleased [](http://example.com/codeflow?start=4&length=10)

dotnet format task and build task is in failed status

marmegh: I have another PR open to fix that. It's caused by an ADO update.

PR #596, opened 2022-01-27
Updating build pipeline for 2022 updates

marmegh:
## Changes
There were changes to the ADO build pipelines that resulted in the sarif-pattern-matcher builds to fail due to upgrading Windows and Visual Studio instances. This change allows our pipelines to access the new folder structure and explicitly defines which versions we are using.

Updating .yml file for Windows and Visual Studio 2022 changes. This includes, switching to explicitly use Windows-2022 instead of Windows-latest and updating to the Visual Studio 2022 VsDevCmd.bat location.

* [ ] Added unit tests

### DIFF
--- a/Src/RE2.Native/RE2.Native.x64.vcxproj
+++ b/Src/RE2.Native/RE2.Native.x64.vcxproj
@@ -25,7 +25,7 @@
     <ConfigurationType>DynamicLibrary</ConfigurationType>
     <TargetName>RE2.Native.x64</TargetName>
     <TargetExt>.dll</TargetExt>
-    <PlatformToolset>v142</PlatformToolset>
+    <PlatformToolset>v143</PlatformToolset>
     <CharacterSet>Unicode</CharacterSet>
     <OutDir>$(SolutionDir)..\Bld\bin\$(Configuration)\x64\</OutDir>
     <IntDir>$(SolutionDir)..\Bld\bin\int\x64\$(Configuration)\</IntDir>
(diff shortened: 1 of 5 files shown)


azure-pipelines.yml
```diff
@@ -2,13 +2,13 @@ pr:
...
 pool:
-  vmImage: 'windows-latest'
+  vmImage: 'windows-2022'
 
 steps:
 - task: BatchScript@1
   displayName: 'Run VsDevCmd.bat'
   inputs:
-    filename: 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\VsDevCmd.bat'
+    filename: 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\Common7\Tools\VsDevCmd.bat'
```
eddynaka, line 11: >2022 [](http://example.com/codeflow?start=56&length=4)

we could also update re2.native.sln to use the latest bits.

f you open the re2.native.sln, check the properties of each project and you will see that it is using VS2019 + a specific version.

Src/RE2.Native/RE2.Native.x64.vcxproj
```diff
@@ -64,4 +64,4 @@
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.targets" />
   <ImportGroup Label="ExtensionTargets">
   </ImportGroup>
-</Project>
+</Project>
```
marmegh, line 67: Visual Studio decided to make this file match the other RE2 projects and remove the last line. No other changes.

azure-pipelines.yml
marmegh: Tried, looks like this change does not currently work. Discussed with Eddy offline.

michaelcfanning (approved):
:shipit:

PR #605, opened 2022-03-22
Update CodeQL Action and removing PREFast

eddynaka:
## Changes
- Removing PREFast action: GH does not work when we generate results for a submodule
- Updating codeql action: executing codeql in all branches


michaelcfanning (approved):
:shipit:

PR #607, opened 2022-04-11
Using latest windows SDK to compile RE2 libraries

eddynaka:
## Changes

This change will allow us to use the latest windows SDK installed instead of hardcoding a specific version.

* [ ] Added unit tests


michaelcfanning:
Here are two questions I always ask: 1. Where are the release notes 2. where are the tests? I understand why tests don't appear here. :)

michaelcfanning (approved):

michaelcfanning:
What you need to do is to make the release pipeline use a well-known Windows SDK and we need to document that in the PR notes.

PR #609, opened 2022-04-11
Updating release history related to PR #607

eddynaka:
## Changes

This change will add an entry to the latest PR that got merged (#607).

When we release it, we will use the latest available in the windows-2022 VM:
https://github.com/actions/virtual-environments/blob/main/images/win/Windows2022-Readme.md#installed-windows-sdks


michaelcfanning (changes requested):
:clock1:

Src/ReleaseHistory.md
```diff
@@ -16,6 +16,10 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## Unreleased
+
+- RE2.Native will compile with the latest Windows SDK 10.0.* installed. [#607](https://github.com/microsoft/sarif-pattern-matcher/pull/607)
```
michaelcfanning, line 21: >Native [](http://example.com/codeflow?start=6&length=6)

We don't want this note. Customers should not receive an arbitrary Windows SDK, and we don't need to document a change that's relevant only to dev builds.

michaelcfanning:
Hello! What's happened here? We were going to pin our pipeline builds to a specific Win SDK and then document that version in our release notes. Is that still happening? :)

Src/ReleaseHistory.md
```diff
@@ -16,6 +16,10 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## Unreleased
+
+- RE2.Native will now compile in all environments with the latest Windows SDK 10.0.* installed. [#607](https://github.com/microsoft/sarif-pattern-matcher/pull/607). Our current release pipelines build NuGet packages with Windows SDK version 10.0.22000.
```
michaelcfanning, line 21: >10.0.22000 [](http://example.com/codeflow?start=241&length=10)

@eddynaka, we need to update this version w/what's actually used in our release/deployment pipeline build agents.

eddynaka: The version that you mentioned is the version that we will be using.
So, the text is accurate.

Thanks for updating it.

PR #610, opened 2022-04-11
Moving validators to separate folders

eddynaka:
## Changes

This PR will move all the validators to folders that represent each type.
No tests were added since the code is the same and I only updated the namespace.

**Users who have reference to the plugins rule will break.**

* [ ] Added unit tests


Src/ReleaseHistory.md
```diff
@@ -16,6 +16,10 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## Unreleased
+
+- RRR: Moving validators to separate folders and updating namespaces accordingly. [#610](https://github.com/microsoft/sarif-pattern-matcher/pull/610)
```
michaelcfanning, line 21: >Moving [](http://example.com/codeflow?start=7&length=6)

Q: does this change impact runtime behaviors of our product?
A: No.

So why did you add a release note? Please remove.

Remember this question: does this change impact runtime behaviors of our product. You really need to understand this concept in order to maintain this file properly, it is intended to help external people who might be impacted by our changes.

It is not a general log of code changes. You might have been thinking our validators are a public API: they are not.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_008.AwsCredentialsValidator.cs
```diff
@@ -10,7 +10,7 @@
 using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk;
 using Microsoft.RE2.Managed;
 
-namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.SecurePlaintextSecretsValidators
```
michaelcfanning, line 13: >SecurePlaintextSecretsValidators [](http://example.com/codeflow?start=71&length=32)

This namespace change isn't really serving us well (except to keep namespaces in line with directories).

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_050.IdentifiableNpmAuthorTokenValidator.cs
```diff
@@ -8,7 +8,7 @@
 using Microsoft.RE2.Managed;
 using Microsoft.Security.Utilities;
 
-namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.SecurePlaintextSecretsValidators
```
michaelcfanning, line 11: >SecurePlaintextSecretsValidators [](http://example.com/codeflow?start=71&length=32)

please revert these changes, just use the base namespace.

an overly granular namespace organization just complicates code and lengthens lines of source.

Src/Plugins/Tests.Security/ReflectionTests.cs
```diff
@@ -7,6 +7,8 @@
 using System.Linq;
 using System.Reflection;
 
+using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.SecurePlaintextSecretsValidators;
```
michaelcfanning, line 10: >SecurePlaintextSecretsValidators [](http://example.com/codeflow?start=67&length=32)

Please make the file move without changing any using statement or namespace.

Thanks!

I understand you're trying to keep everything in line with directory structure but we didn't have a problem with our namespaces before, really, which aren't entirely consistent with current directory structure.

The problem is that the JSON files are hard to locate buried in dozens of validators, that was the complaint I remarked.

I really do appreciate that you tried to make a more assertive fix here, but when I see changes like this specific one that add a new 'using' with no discernable increase in value, I just think we're cluttering code unnecessarily.

michaelcfanning (changes requested):
:clock1:

Src/ReleaseHistory.md
eddynaka: Since I'm moving back the namespace, it is reasonable to remove this.

But, the validators API are all public. You, as a user, would be able to instantiate the AdoPatValidator and call `IsValidStatic`.
[https://github.com/microsoft/sarif-pattern-matcher/blob/1cb648695e1f3ea04e4ea9732f3e6939b19a5b0f/Src/Plugins/Security/SEC101_102.AdoPatValidator.cs#L12](https://github.com/microsoft/sarif-pattern-matcher/blob/1cb648695e1f3ea04e4ea9732f3e6939b19a5b0f/Src/Plugins/Security/SEC101_102.AdoPatValidator.cs#L12)

[https://github.com/microsoft/sarif-pattern-matcher/blob/1cb648695e1f3ea04e4ea9732f3e6939b19a5b0f/Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs#L12](https://github.com/microsoft/sarif-pattern-matcher/blob/1cb648695e1f3ea04e4ea9732f3e6939b19a5b0f/Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs#L12)

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_008.AwsCredentialsValidator.cs
eddynaka: ok!
just updated.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_050.IdentifiableNpmAuthorTokenValidator.cs
eddynaka: just updated!

Src/Plugins/Tests.Security/ReflectionTests.cs
eddynaka: Ok!
Just updated!

Src/ReleaseHistory.md
eddynaka: Removed the line.

michaelcfanning (approved):
:shipit:

PR #612, opened 2022-04-26
Updating sarif.sdk submodules

eddynaka:
## Changes

This change will update the sarif sdk submodules to point to the latest released that we made (2.4.15).

Here you can find the difference between the submodules change:
https://github.com/microsoft/sarif-sdk/compare/b2c7e399cfaf1fde47b336bbb4976f723263efa3...9f0eed7549736b28d59a2e93f443ba47e3bd978e

* [ ] Added unit tests

### DIFF
--- a/Src/Plugins/Tests.Security/EndToEndTests.cs
+++ b/Src/Plugins/Tests.Security/EndToEndTests.cs
@@ -28,15 +28,15 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 
         protected abstract string RuleId { get; }
 
-        protected abstract string Service { get; }
-
         protected abstract string Framework { get; }
+
+        protected abstract string PluginName { get; }
 
         protected override string TestLogResourceNameRoot => $"Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.TestData.{TypeUnderTest}";
 
         protected override string TestBinaryTestDataDirectory => Path.Combine(ProductRootDirectory, "src", "Plugins", TestBinaryName, "TestData");
 
-        protected override string ProductTestDataDirectory => Path.Combine(TestBinaryTestDataDirectory, Service);
+        protected override string ProductTestDataDirectory => Path.Combine(TestBinaryTestDataDirectory, PluginName);
 
         protected override IDictionary<string, string> ConstructTestOutputsFromInputResources(IEnumerable<string> inputResourceNames, object parameter)
         {
--- a/Src/Plugins/Tests.Security/EndToEndTests.cs
+++ b/Src/Plugins/Tests.Security/EndToEndTests.cs
@@ -28,22 +28,28 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 
         protected abstract string RuleId { get; }
 
+        protected abstract string Service { get; }
+
         protected abstract string Framework { get; }
 
         protected override string TestLogResourceNameRoot => $"Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.TestData.{TypeUnderTest}";
 
-        protected override string ProductDirectory => Path.Combine(base.ProductDirectory, @"Plugins\Tests.Security");
+        protected override string TestBinaryTestDataDirectory => Path.Combine(ProductRootDirectory, "src", "Plugins", TestBinaryName, "TestData");
+
+        protected override string ProductTestDataDirectory => Path.Combine(TestBinaryTestDataDirectory, Service);
 
         protected override IDictionary<string, string> ConstructTestOutputsFromInputResources(IEnumerable<string> inputResourceNames, object parameter)
         {
             var results = new Dictionary<string, string>();
             var dict = new Dictionary<string, Task<string>>();
 
+            const string input = "Inputs.";
+
             foreach (string inputResourceName in inputResourceNames)
             {
-                string secret = inputResourceName.Substring("Inputs.".Length);
+                string name = inputResourceName.Substring(inputResourceName.IndexOf(input) + input.Length);
 
-                dict[secret] = Task.Factory.StartNew(() => ConstructTestOutputFromInputResource(inputResourceName, parameter));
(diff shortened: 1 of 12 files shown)


Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -28,22 +28,28 @@ protected EndToEndTests(ITestOutputHelper outputHelper) : base(outputHelper)
...
         {
             var results = new Dictionary<string, string>();
             var dict = new Dictionary<string, Task<string>>();
 
+            const string input = "Inputs.";
+
             foreach (string inputResourceName in inputResourceNames)
             {
-                string secret = inputResourceName.Substring("Inputs.".Length);
+                string name = inputResourceName.Substring(inputResourceName.IndexOf(input) + input.Length);
```
eddynaka, line 50: >inputResourceName [](http://example.com/codeflow?start=58&length=17)

Previously, the inputResourceName was composed of `Inputs.RuleId.FileName.extension`. With the sarif-sdk change, inputResourceName has the fully qualified name: `Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security. etc etc`.

So, I had to made a change to find the name of the file that we are analyzing.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -98,7 +98,7 @@ private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
             string searchDefinitionsPath = @$"c:\{Guid.NewGuid()}.json";
 
             var mockFileSystem = new Mock<IFileSystem>();
-
+            mockFileSystem.Setup(x => x.DirectoryExists(rootDirectory)).Returns(true);
```
eddynaka, line 101: >DirectoryExists [](http://example.com/codeflow?start=40&length=15)

we have a new verification checking if the directory existing in the sarif.sdk. We will need to mock to make it to work.

Src/ReleaseHistory.md
```diff
@@ -19,6 +19,7 @@
 ## Unreleased
 
 - Bump MongoDB.Driver from 2.13.1 to 2.15.0 and Microsoft.AspNetCore.Http from 2.1.0 to 2.2.0. [#608](https://github.com/microsoft/sarif-pattern-matcher/pull/608)
+- Bump SARIF.SDK from 2.4.13 to 2.4.15. [#612](https://github.com/microsoft/sarif-pattern-matcher/pull/612)
```
michaelcfanning, line 22: >SARIF.SDK [](http://example.com/codeflow?start=7&length=9)

Casing does not correspond to NuGet

michaelcfanning, line 22: >Bump SARIF.SDK from 2.4.13 to 2.4.15.  [](http://example.com/codeflow?start=2&length=38)

Bump Sarif.Sdk from 2.4.13 to 2.4.15 by updating submodule to commit XXXXXXXXXXXXXXX.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
michaelcfanning: Nice, this is from Mary's change, got it.

Src/Plugins/Tests.Security/EndToEndTests.cs
eddynaka: I took a deeper look at this.
The Sarif.Sdk changed the behavior in the following way:
- previously, we had:
```csharp
        private string ConstructFullInputResourceName(string resourceName)
            => "Inputs." + resourceName;
```

- currently, we have:
```csharp
        private string GetFullResourcePathForInputResource(string resourceName)
        {
            return $"{TestLogResourceNameRoot}.Inputs.{resourceName}";
        }
```

There isn't any API that converts the embedded resource path to the file name. Also, we don't see this issue in the sarif.sdk because almost all the tests uses the RunTest call passing a single file, which wouldn't generate this.

eddynaka: To reduce this hard-coded strings, I'm now using the parameter argument and pass the original list. It is not ideal, but that will prevent us from hard-coding any substring/indexof logic.

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -30,20 +30,25 @@ protected EndToEndTests(ITestOutputHelper outputHelper) : base(outputHelper)
 
         protected abstract string Framework { get; }
 
+        protected abstract string PluginName { get; }
+
         protected override string TestLogResourceNameRoot => $"Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security.TestData.{TypeUnderTest}";
 
-        protected override string ProductDirectory => Path.Combine(base.ProductDirectory, @"Plugins\Tests.Security");
+        protected override string TestBinaryTestDataDirectory => Path.Combine(ProductRootDirectory, "src", "Plugins", TestBinaryName, "TestData");
```
eddynaka, line 37: >TestBinaryTestDataDirectory [](http://example.com/codeflow?start=34&length=27)

We need to override this because the paths have another level called Plugins.

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -147,7 +151,7 @@ protected void RunAllTests()
                     Path.GetFileNameWithoutExtension(testFileName) + ".sarif";
             }
 
-            RunTest(inputFiles, expectedOutputResourceMap, enforceNotificationsFree: true);
+            RunTest(inputFiles, expectedOutputResourceMap, enforceNotificationsFree: true, parameter: inputFiles);
```
eddynaka, line 154: >parameter: inputFiles [](http://example.com/codeflow?start=91&length=21)

This will be used in ConstructTestOutputsFromInputResources to get the name of the file under analysis.

michaelcfanning (approved):

PR #611, opened 2022-04-20
Url detection

michaelcfanning:
## Changes

Add a simple URL analysis that's currently constrained to a small subset of code files. We could consider expanding this and adding a dynamic validator (that can do things like provide a broken link detection).

@suvamM

PR #613, opened 2022-04-27
Adding Url rule to ReleaseHistory

eddynaka:
This change will update the ReleaseHistory file, adding the Url rule that was developed in the PR #611


michaelcfanning (approved):
:shipit:

PR #615, opened 2022-06-13
Enable response file parsing in driver.

michaelcfanning:
This change enables response file processing as provided by the driver framework. This change additionally enables a behavior previously requested by the Guardian team: to provide a zero (success) result code when a user initiates any valid command. Currently, due to our specific use pattern for CommandLibrary, the tool returns a result code of '1' on any help command. 

After this change, any argument prefixed with the '@' will interpreted as a file name. That file, if it exists, will be loaded. Each line in that file will be expanded into an argument that will be injected into the command-line (for example all env vars will be expanded).

 spam.exe @MyResponseFile.rsp

The driver code for the argument expansion is in the Sarif Driver [EntryPointUtilities](https://github.com/microsoft/sarif-sdk/blob/fc9a9dfb865096b5aaa9fa3651854670940f7459/src/Sarif.Driver/Sdk/EntryPointUtilities.cs#L12) code.

@marmegh @shaopeng-gh @cfaucon @EasyRhinoMSFT @eddynaka

### DIFF
--- a/Src/ReleaseHistory.md
+++ b/Src/ReleaseHistory.md
@@ -16,6 +16,11 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## *v1.9.1*
+
+- FEATURE: Enable response file parsing provided by driver framework. Arguments (e.g., '@Commands.rsp') prefixed with a '@' character will be evaluated as a file path to a text file that contains commands to be injected on the command-line. 
+
+
 ## *v1.9.0*
 
 - Bump MongoDB.Driver from 2.13.1 to 2.15.0 and Microsoft.AspNetCore.Http from 2.1.0 to 2.2.0. [#608](https://github.com/microsoft/sarif-pattern-matcher/pull/608)
--- a/Src/Sarif.PatternMatcher.Cli/ImportAndAnalyzeOptions.cs
+++ b/Src/Sarif.PatternMatcher.Cli/ImportAndAnalyzeOptions.cs
@@ -5,13 +5,13 @@ using CommandLine;
 
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
 {
-    [Verb("import-analyze", HelpText = "Export rules metadata to a markdown file.")]
+    [Verb("import-analyze", HelpText = "Import from and analyze database contents.")]
     internal class ImportAndAnalyzeOptions : AnalyzeOptions
     {
         [Option(
             "host",
             HelpText = "The database host from which we will retrieve data to scan. " +
-                       "Create AppClientId, AppSecrent, and AuthorityId as environment variables.",
+                       "Create AppClientId, AppSecret, and AuthorityId as environment variables.",
             Required = true)]
         public string Host { get; set; }
 
--- a/Src/Sarif.PatternMatcher.Cli/Program.cs
+++ b/Src/Sarif.PatternMatcher.Cli/Program.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using System.Runtime.CompilerServices;
 using System.Text;
 
@@ -20,6 +22,15 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
         {
             Console.OutputEncoding = Encoding.UTF8;
 
+            args = EntryPointUtilities.GenerateArguments(args, FileSystem, new EnvironmentVariables());
+
+            bool isValidHelpCommand =
+                args.Length > 0 &&
+                args[0] == "help" &&
+                ((args.Length == 2 && IsValidVerbName(args[1])) || args.Length == 1);
+
+            bool isVersionCommand = args[0] == "version" && args.Length == 1;
+
             return Parser.Default.ParseArguments<
                 AnalyzeOptions,
                 AnalyzeDatabaseOptions,
@@ -34,7 +45,19 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
                 (ExportRulesMetatadaOptions options) => new ExportRulesMetatadaCommand().Run(options),
                 (ExportSearchDefinitionsOptions options) => new ExportSearchDefinitionsCommand().Run(options),
                 (ValidateOptions options) => new ValidateCommand().Run(options),
-                _ => CommandBase.FAILURE);
+                _ => isValidHelpCommand || isVersionCommand
+                        ? CommandBase.SUCCESS
+                        : CommandBase.FAILURE);
+        }
+
+        private static bool IsValidVerbName(string verb)
+        {
+            return
+                verb == "analyze" ||
+                verb == "analyze-database" ||
+                verb == "export-rules" ||
+                verb == "export-search-definitions" ||
+                verb == "import-analyze";
         }
     }
 }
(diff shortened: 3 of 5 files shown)


Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -125,7 +125,7 @@ private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
 
             string tempFileName = Path.GetTempFileName();
             string sarifLogFileName = $"{tempFileName}.sarif";
-            SarifLog sarifLog = null;
+            SarifLog sarifLog = null, sarifLogFromResponseFile = null;
```
eddynaka, line 128: >sarifLogFromResponseFile [](http://example.com/codeflow?start=38&length=24)

nit: remove this since you are not using it.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        private static IList<string> GetVerbNames()
+        {
+            var verbNames = new List<string>();
+
+            Assembly assembly = typeof(Program).Assembly;
+            foreach (Type type in assembly.GetTypes())
+            {
+                VerbAttribute verbAttribute = type.GetAttribute<VerbAttribute>();
```
eddynaka, line 115: >GetAttribute [](http://example.com/codeflow?start=51&length=12)

replace this by GetCustomAttribute and remove the castle.core.internal using

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -0,0 +1,127 @@
...
+            // Explicitly provoke a failure to build confidence in prior
+            // successful tests.
+            TestResponseFile(
+                new string[] { $"help {Guid.NewGuid()}" }, CommandBase.FAILURE);
+        }
+
+        [Fact]
+        public void Program_VerbListMatcheExpected()
+        {
+            GetVerbNames().Count.Should().Be(5);
```
eddynaka, line 73: >Count.Should().Be [](http://example.com/codeflow?start=27&length=17)

nit: Should().HaveCount

michaelcfanning: What's the reason you have in mind? The key quality concern here is to ensure the count is 5. The GetVerbNames helper is test code, so we can assume things about its behavior.

eddynaka: it is a different syntax of the same thing. comparing the direct number vs the collection.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            // We expect:
+            // 'analyze'
+            // 'analyze-database'
+            // import-analyze'
+            // 'export-rules'
+            // 'export-search-definitions'
```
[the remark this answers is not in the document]
michaelcfanning: Nuts, good question and I intended to add an explicit comment here and neglected to. hold on.

cfaucon (approved):
:shipit:

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -0,0 +1,125 @@
...
+    {
+        [Fact]
+        public void Program_ResponseFileWorks()
+        {
+            // Ensure that simple commands, such as help and version
+            // work properly (as evident by a success response code).
+            TestResponseFile(new string[] { "help" });
+            TestResponseFile(new string[] { "version" });
+
+            // Our mocking strategy assumes that a single mocked called to File.ReadAllLines
```
marmegh, line 30: >single mocked called [](http://example.com/codeflow?start=51&length=20)

nit: using called (verb) instead of call (noun).

Src/Sarif.PatternMatcher.Cli/Program.cs
```diff
@@ -20,6 +22,15 @@ internal static int Main(string[] args)
         {
             Console.OutputEncoding = Encoding.UTF8;
 
+            args = EntryPointUtilities.GenerateArguments(args, FileSystem, new EnvironmentVariables());
```
michaelcfanning, line 25: >GenerateArguments [](http://example.com/codeflow?start=39&length=17)

This is the essential change. This driver helper processes all args by expanding any of them that reference a response file.

Src/ReleaseHistory.md
```diff
@@ -16,6 +16,11 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## *v1.9.1*
+
+- FEATURE: Enable response file parsing provided by driver framework. Arguments (e.g., '@Commands.rsp') prefixed with a '@' character will be evaluated as a file path to a text file that contains commands to be injected on the command-line.
```
michaelcfanning, line 21: >Enable [](http://example.com/codeflow?start=11&length=6)

Q: Should we update our command-line docs with this useful new functionality? A: Yes.

Src/Sarif.PatternMatcher.Cli/ImportAndAnalyzeOptions.cs
```diff
@@ -5,13 +5,13 @@
 
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
 {
-    [Verb("import-analyze", HelpText = "Export rules metadata to a markdown file.")]
+    [Verb("import-analyze", HelpText = "Import from and analyze database contents.")]
```
michaelcfanning, line 8: >Import from and analyze database contents [](http://example.com/codeflow?start=40&length=41)

This description was wrong, a cut and paste error. I fixed it up.

Src/Sarif.PatternMatcher.Cli/Program.cs
```diff
@@ -20,6 +22,15 @@ internal static int Main(string[] args)
...
+
+            bool isValidHelpCommand =
+                args.Length > 0 &&
+                args[0] == "help" &&
+                ((args.Length == 2 && IsValidVerbName(args[1])) || args.Length == 1);
```
shaopeng-gh, line 30: how about this command:
spam analyze --help

eddynaka (approved):
:shipit:

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
shaopeng-gh: I think Eddy mean it is a preferred syntax for fluentassertions to use:
collection.Should().HaveCount(5);
v.s.
collection.Count.Should().Be(5);

Src/Sarif.PatternMatcher.Cli/Program.cs
shaopeng-gh, line 25: >GenerateArguments [](http://example.com/codeflow?start=39&length=17)

question for the logic,
if there is a ResponseFile arg, we expand and return all.
But if not, do we miss out all the normal args user input?
the args is replaced after this line execution and will be used below line "Parser.Default.ParseArguments"

shaopeng-gh (changes requested):
:clock1:

Src/Sarif.PatternMatcher.Cli/Program.cs
shaopeng-gh: I see GenerateArguments do keep the normal args and adds the ones from ResponseFile. Not an issue.

michaelcfanning: Yes! It's actually very flexible, you can arbitrarily insert one or more response files onto the same command-line:

myapp.exe arg1 @responsefile1.rsp anotherArg @responseFile2.rsp

PR #616, opened 2022-06-15
Add unique secret fingerprint and refresh fingerprint output to allow for proper versioning.

michaelcfanning:
Our supported standard for all versioned fingerprints is that they consist of a fingerprint name + a suffix that indicates the current form of the hash. Optionally a fingerprint could contain a previous version. E.g., 'MyHash/current' and 'MyHash/previous'. This strategy works as long as we can guarantee processing is chained between tool versions, i.e., that every change in a hash produces a log file that allows systems that cache information to link their old data to the current hash.

An even more detailed versioning scheme would actually include a version that would allow users to 'pin' their hash to a specific version. The downside of this approach is that the log file would need to persist all historical versions.

This current draft PR is opened for discussion. What it does:

* Add a new unique secret fingerprint, defined to be the minimal data that uniquely identifies a credential. In some cases, due to a high degree of confidence that a specific component is globally unique, that single token can comprise the unique secret fingerprint. In systems where users can propose user names and passwords that could be duplicative with other services or assets, we may be required to add more detail (and even a platform designation). This draft PR starts simply with emitting the 'secret' element as this fingerprint. We can do the work now to actually create a secret fingerprint for things like SQL Server credentials *or* we can start with the simple secret scenario knowing that we will break secret fingerprints in the future (an opportunity to test our versioning mechanisms).
* This change drops stale fingerprints for asset and validation fingerprints.
* Heads up: there's a case that we should be retaining the 'ValidationFingerprintHash/v1' data as 'UniqueSecretHash/previous', as this information was effectively the deduplicating hash in our previous scanner.

Note: the test files can be sanity checked in this draft PR - I wouldn't spend too much time on them since our design/implementation isn't fixed yet.

@suvamM, you are the primary reviewer on this. Let's talk tomorrow about this in-person.
@marmegh, please track this PR carefully as an onboarding dev into the project.
@eddynaka, as always we need your review and feedback, which will be supplemental to Suvam's sign-off.


Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -227,6 +226,29 @@ public string GetValidationFingerprintHashText(bool jsonFormat = false)
...
+                Secret = this.Secret,
+            }.ToString();
+        }
+
+        /// <summary>
+        /// Returns the hash of the minimal fingerprint that uniquely
+        /// identifies a credential at a specific point-in-time.
+        /// </summary>
+        /// <returns>The minimal fingerprint that uniquely identifies a credential.</returns>
+        public string GetUniqueSecretHash()
```
suvamM, line 247: @michaelcfanning Is this hash guaranteed to be equal to the hash generated by the legacy SPAM tool?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,12 +19,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
...
     public class SearchSkimmer : Skimmer<AnalyzeContext>
     {
-        public const string AssetFingerprint = "AssetFingerprint/v1";
-        public const string ValidationFingerprint = "ValidationFingerprint/v1";
-        public const string ValidationFingerprintHash = "ValidationFingerprintHash/v1";
+        public const string UniqueSecretHashCurrent = "UniqueSecretHash256/current";
 
-        public const string AssetFingerprintV2 = "AssetFingerprint/v2";
-        public const string ValidationFingerprintV2 = "ValidationFingerprint/v2";
+        public const string AssetFingerprintCurrent = "AssetFingerprint/current";
```
suvamM, line 24: @michaelcfanning I am now a little uncertain about this versioning using `current` and `previous`.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,12 +19,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
...
+        public const string ValidationFingerprintCurrent = "ValidationFingerprint/current";
+        public const string UniqueSecretFingerprintCurrent = "UniqueSecretFingerprint/current";
```
suvamM, line 26: @michaelcfanning I'm sorry, but I am still not totally clear about this change based on versioning. In my mind, there are two distinct problems to solve:

1. Design a hash based on just the secret text which correlates with the hash generated by the legacy SPAM tool.

2. Extend the existing fingerprinting mechanism to account for changes in the fingerprint itself. 

I am unclear if this PR addresses (1).

Src/Sarif.PatternMatcher.Sdk/FingerprintElements.cs
```diff
@@ -0,0 +1,25 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
+{
+    [Flags]
+    internal enum FingerprintElements
```
michaelcfanning, line 11: >FingerprintElements [](http://example.com/codeflow?start=18&length=19)

FYI, this code is being added preparatory to a future change. It is not consume currently, nor is it in a finished condition (requires doc comments, for one thing).

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,12 +19,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public class SearchSkimmer : Skimmer<AnalyzeContext>
     {
-        public const string AssetFingerprint = "AssetFingerprint/v1";
-        public const string ValidationFingerprint = "ValidationFingerprint/v1";
-        public const string ValidationFingerprintHash = "ValidationFingerprintHash/v1";
+        public const string SecretHashSha256 = "SecretHashSha256/current";
```
eddynaka, line 22: >SecretHashSha256 [](http://example.com/codeflow?start=28&length=16)

If this is a secret hash using sha256, then, we wouldn't require the `/current`, because sha256 won't change.

Unless we rename this to SecretHashCurrent, which in this case would be a sha256

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,12 +19,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
...
 
-        public const string AssetFingerprintV2 = "AssetFingerprint/v2";
-        public const string ValidationFingerprintV2 = "ValidationFingerprint/v2";
+        public const string AssetFingerprintCurrent = "AssetFingerprint/current";
```
eddynaka, line 24: >AssetFingerprintCurrent [](http://example.com/codeflow?start=28&length=23)

@cfaucon , be aware that we will have to update this once we bump the packages.

Src/Sarif.PatternMatcher/ValidatingVisitor.cs
```diff
@@ -86,8 +86,7 @@ public override Result VisitResult(Result node)
                 node.Message.Arguments[1] = validationPrefix;
                 node.Message.Arguments[4] = validationSuffix;
                 node.Message.Arguments[5] = SearchSkimmer.NormalizeValidatorMessage(message);
-                node.Fingerprints[SearchSkimmer.AssetFingerprint] = fingerprint.GetAssetFingerprintText();
-                node.Fingerprints[SearchSkimmer.AssetFingerprintV2] = fingerprint.GetAssetFingerprintText(jsonFormat: true);
+                node.Fingerprints[SearchSkimmer.AssetFingerprintCurrent] = fingerprint.GetAssetFingerprint();
```
eddynaka, line 89: we have to pass true in this case

Src/Sarif.PatternMatcher.Sdk/Fingerprint.cs
```diff
@@ -227,6 +226,29 @@ public string GetValidationFingerprintHashText(bool jsonFormat = false)
             return ComputeHash(validationFingerprint);
         }
 
+        /// <summary>
+        /// Returns a fingerprint comprised of the minimal information that
+        /// uniquely identifies a credential at a specific point-in-time.
+        /// </summary>
+        /// <returns>The minimal fingerprint that uniquely identifies a credential.</returns>
+        public string GetUniqueSecretFingerprint()
```
eddynaka, line 234: >Unique [](http://example.com/codeflow?start=25&length=6)

I dont think we need to add Unique here. From the fingerprint perspective, it is just the fingerprint with the secret.

eddynaka: The same applies for the next method.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -1174,28 +1173,27 @@ private Region ConstructRegion(AnalyzeContext context, FlexMatch regionFlexMatch
 
             return new Dictionary<string, string>()
             {
-                { AssetFingerprint, fingerprint.GetAssetFingerprintText() },
-                { ValidationFingerprint, fingerprint.GetValidationFingerprintText() },
-                { ValidationFingerprintHash, fingerprint.GetValidationFingerprintHashText() },
-                { AssetFingerprintV2, fingerprint.GetAssetFingerprintText(jsonFormat: true) },
```
eddynaka, line 1180: >jsonFormat: true [](http://example.com/codeflow?start=74&length=16)

you cannot remove this

eddynaka (changes requested):
:clock1:

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_006.GitHubPat.sarif
```diff
@@ -83,11 +83,10 @@
             }
           ],
           "fingerprints": {
-            "AssetFingerprint/v1": "[platform=GitHub]",
-            "ValidationFingerprint/v1": "[secret=ff34885a8624460a855540c6592698d2f1812843]",
-            "ValidationFingerprintHash/v1": "25799b447c11c0dec40e393f9b0349fbdf11bbb5bb8c2e6c698dbbd1bbe7cb7c",
-            "AssetFingerprint/v2": "{\"platform\":\"GitHub\"}",
-            "ValidationFingerprint/v2": "{\"secret\":\"ff34885a8624460a855540c6592698d2f1812843\"}"
+            "AssetFingerprint/current": "[platform=GitHub]",
```
eddynaka, line 86: >[platform=GitHub] [](http://example.com/codeflow?start=41&length=17)

this should be json.

this is also related to my other comment saying that you are missing an argument

eddynaka: the same applies for validationfingerprint

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_009.LinkedInCredentials.sarif
```diff
@@ -83,11 +83,10 @@
             }
           ],
           "fingerprints": {
-            "AssetFingerprint/v1": "[id=0000000aaaaaaa][platform=LinkedIn]",
-            "ValidationFingerprint/v1": "[id=0000000aaaaaaa][secret=00000000aaaaaaaa]",
-            "ValidationFingerprintHash/v1": "adf099a2a5d039c5923f6d8de7f6e96ee22bcb3b1674596df5ab22198a0bd8b4",
```
eddynaka, line 88: >adf099a2a5d039c5923f6d8de7f6e96ee22bcb3b1674596df5ab22198a0bd8b4 [](http://example.com/codeflow?start=45&length=64)

I was not expecting this.
This means that we won't be able to link new results with old results

eddynaka (approved):
:shipit:

PR #619, opened 2022-06-23
Update SARIF sdk.

michaelcfanning:
Update SARIF SDK submodule to get benefit of most recent changes.

@marmegh @EasyRhinoMSFT @yongyan-gh @shaopeng-gh


shaopeng-gh:
does the other submodule showing not update to date as well: Src/RE2.Native/re2

shaopeng-gh (approved):
:shipit:

eddynaka:
I think we should add a release notes about this because we are basically introducing the comparers and that might change some behaviors.

marmegh:
+1

---
In reply to: [1164981971](https://github.com/microsoft/sarif-pattern-matcher/pull/619#issuecomment-1164981971) [](http://example.com/codeflow?ancestors=1164981971,1164981971)

yongyan-gh:
> I think we should add a release notes about this because we are basically introducing the comparers and that might change some behaviors.

Checked this Sarif.Sdk update contains 2 commitments than current Sarif.Sdk version, I think the manual comparers already included in current Sarif.Sdk.

The auto-generated comparers are not in yet.

PR #621, opened 2022-06-24
Resolve unhandled exceptions when analyzing folders that contain large files

### LINKED ISSUE #617: The argument/property file-size-in-kb does not prevent file from being loaded when it is greater than the property
# The issue:

When you use the command line `.\spam.exe analyze c:\SOME_PATH\ -o c:\OUTPUT_PATH\file.sarif --recurse --force -d "SEC101.SecurePlaintextSecrets.json" --file-size-in-kb 10485 --insert "Hashes;RegionSnippets;ContextRegionSnippets;ComprehensiveRegionProperties"` you expect it to filter out files where size is greater than 10485 KB. But, it does not do that.

In my scenario, I have a folder with big files (2GB, 100MB, etc.) and, with that, it fails to hash the file at first and the program stops. When I disabled the `Hashes` flag, I observed the issue in the SearchSkimmer.

# Investigation:

In the multithreadedAnalyzeCommandBase, we are hashing the file independently of the size:
https://github.com/microsoft/sarif-sdk/blob/503fe2946bda129bf0900dfa20ffdfcbe88e0be9/src/Sarif.Driver/Sdk/MultithreadedAnalyzeCommandBase.cs#L480

At the skimmer, we load the file again:
https://github.com/microsoft/sarif-pattern-matcher/blob/23542daf1ac2d637e52bfd141dc8d9e4368d8988/Src/Sarif.PatternMatcher/SearchSkimmer.cs#L154-L159

And we discard after loading it:
https://github.com/microsoft/sarif-pattern-matcher/blob/23542daf1ac2d637e52bfd141dc8d9e4368d8988/Src/Sarif.PatternMatcher/SearchSkimmer.cs#L186

# Suggestion/Fix:

We should move the flag the the MultithreadedAnalyzeCommandBase/AnalyzeCommandBase and before hashing, use some verification to check if we should process that or not.
Also, the size verification, should occur before reading the entire file.

marmegh:
## Changes

Fixes #617 .

`OutOfMemory` and `NullReferenceException` exceptions are being thrown when analyzing folders containing large files. There is an analysis option that is supposed to enforce a file size limit, but this was not being applied before attempting to read the file.

To address this, the analysis option was moved to the base class and exclusion logic was similarly moved to the appropriate base class.

* [x] Added unit tests


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -155,6 +155,12 @@ public override void Analyze(AnalyzeContext context)
             {
                 try
                 {
+                    var fileInfo = new FileInfo(filePath);
+                    if (context.FileSizeInKilobytes != -1 && fileInfo.Length / 1024 > context.FileSizeInKilobytes)
```
eddynaka, line 159: >if (context.FileSizeInKilobytes != -1 && fileInfo.Length / 1024 > context.FileSizeInKilobytes) [](http://example.com/codeflow?start=20&length=94)

you would need a similar condition to the else clause.

The file is already in memory and you should return if the size of it is greater than the max allowed.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -17,6 +17,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
 {
     public class AnalyzeCommandTests
     {
+        private const string SmallTargetName = "smallTarget.txt";
+        private const string LargeTargetName = "largeTarget.txt";
```
eddynaka, line 21: do you really need this?

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +92,66 @@ public void AnalyzeCommand_IntrafileBasic()
...
+        }
+
         private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
         {
             string sarifOutput;
-            string rootDirectory = @"e:\repros";
-            string scanTargetName = $"test.txt";
-            string scanTargetPath = @$"{rootDirectory}\{scanTargetName}";
+            string rootDirectory = Directory.GetCurrentDirectory();
+            string scanTargetName = SmallTargetName;
```
eddynaka, line 153: >scanTargetName [](http://example.com/codeflow?start=19&length=14)

don't you need to make this accept the targetName? small and large?

marmegh: No, this is to preserve the original functionality of this test, which relied on mocks and will throw an unhandled exception without a valid file to check the size of.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
marmegh: Are you asking if we really need a different large target file? If so, yes. This allows us to confirm that when we are filtering out files that are too large to read, we are still able to process files that are under the threshold.

eddynaka: I asked this because I didn't see you loading the file in memory. So, I was wondering if you need this or not.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +92,66 @@ public void AnalyzeCommand_IntrafileBasic()
...
+            sarifLogWithLargeFileExcluded.Runs?[0].Results?.Count.Should().Be(2);
+            sarifLogWithLargeFileIncluded.Should().NotBeNull();
+            sarifLogWithLargeFileIncluded.Runs?[0].Results?.Count.Should().Be(4);
+            sarifLogWithLargeFileIncluded.Runs[0].Artifacts.Count.Should().Be(2);
+
+            Assert.True(string.Equals(sarifLogWithLargeFileIncluded.Runs[0].Artifacts[0].Location.Uri, smallTargetPath) ||
+                string.Equals(sarifLogWithLargeFileIncluded.Runs[0].Artifacts[0].Location.Uri, largeTargetPath));
+
+            Assert.True(string.Equals(sarifLogWithLargeFileIncluded.Runs[0].Artifacts[1].Location.Uri, smallTargetPath) ||
+                string.Equals(sarifLogWithLargeFileIncluded.Runs[0].Artifacts[1].Location.Uri, largeTargetPath));
```
marmegh, line 135: This is where we are using the path for the large file within the test.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -134,7 +191,82 @@ private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
...
+            string sarifLogFileName = $"{tempFileName}.sarif";
+            SarifLog sarifLog = null;
+
+            try
+            {
+                string[] args = new[]
+                {
+                    "analyze",
+                    largeTargetPath,
+                    smallTargetPath,
```
marmegh, line 266: This is where we are passing the paths to the analyze command.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
suvamM, line 159: @marmegh A comment here would help. What is the logic? I am guessing `context.FileSizeInKilobytes` is the size of the file under analysis. What does it mean for the value to be -1? What is `fileInfo.Length`?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -182,10 +188,12 @@ public override void Analyze(AnalyzeContext context)
                     throw;
                 }
             }
-
-            if (context.FileSizeInKilobytes != -1 && context.FileContents.String.Length / 1024 > context.FileSizeInKilobytes)
+            else
             {
-                return;
+                if (context.FileSizeInKilobytes != -1 && context.FileContents.String.Length / 1024 > context.FileSizeInKilobytes)
+                {
```
suvamM, line 194: @marmegh I guess a similar comment would help here.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +92,66 @@ public void AnalyzeCommand_IntrafileBasic()
...
+
+            SarifLog sarifLogWithLargeFileExcluded = RunAnalyzeCommandWithFileSizeInKilobytes(
+                definitionsText,
+                fileContents,
+                fileSizeInKilobytes: 1);
+
+            SarifLog sarifLogWithLargeFileIncluded = RunAnalyzeCommandWithFileSizeInKilobytes(
+                definitionsText,
+                fileContents,
+                fileSizeInKilobytes: 1024);
```
suvamM, line 119: @marmegh Perhaps a naive question, but what happens if I pass a value like 999999999999?

Src/ReleaseHistory.md
```diff
@@ -16,6 +16,10 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## Unreleased
+
+- BUG: Resolve unhandled `OutOfMemory` and `NullReferenceException` exceptions when analyzing folders that contain big files even when using --file-size-in-kb. [#621](https://github.com/microsoft/sarif-pattern-matcher/pull/621)
```
michaelcfanning, line 21: >Resolve [](http://example.com/codeflow?start=7&length=7)

This note is confusing to me. Is the issue only provoked when using --file-size-in-kb? If not, don't mention it. If it's only provoked using this argument, we don't need the term 'even'. I will review the change and see what I think. :) But this was a little clear on first look. Also, you should always enquote arg names, like `--file-size-in-kb`. I would also suggest using 'large' over 'big', the former term is more typical for API/docs when referring to file sizes.

Src/ReleaseHistory.md
```diff
@@ -16,6 +16,10 @@
 - UER => eliminate unhandled exceptions in rules
 - UEE => eliminate unhandled exceptions in engine
 
+## Unreleased
```
michaelcfanning, line 19: >Unreleased [](http://example.com/codeflow?start=3&length=10)

You updated the SDK but there's nothing in the release notes about it. Did we need to do that in order to accept a bug fix relevant to this tool? If so, we should mention that. If not, it's fine not to mention the SDK update).

Remember, 'was there a potentially customer visible impact to the change?'. If so, talk about it. If not, ok, silence is golden. :)

michaelcfanning:
> This is a small test file.

Did you really need to add a new file for this? Couldn't we have repurposed an existing test file somewhere?

Did you need to add a large file? Couldn't we have simply specified a small file size threshold?

Why are we using checked in files for this testing at all rather than simply mocking it?

---
Refers to: Src/Test.UnitTests.Sarif.PatternMatcher.Cli/SmallTarget.txt:1 in f58f0bb. [](commit_id = f58f0bb8d3d9bcc4a7fb7b4b34155a3880b7831e, deletion_comment = False)

Src/Test.UnitTests.Sarif.PatternMatcher/TestLogger.cs
```diff
@@ -37,6 +38,7 @@ public void LogConfigurationNotification(Notification notification)
 
         public void LogToolNotification(Notification notification)
         {
+            ToolNotification = notification;
```
michaelcfanning, line 41: >ToolNotification [](http://example.com/codeflow?start=12&length=16)

Please make this a list and follow the log pattern for results to initialize and add the notification.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
michaelcfanning, line 159: >fileInfo [](http://example.com/codeflow?start=61&length=8)

There are many corner cases you aren't covering, in part because you've taken the route of using simple checked in files for your tests. Instead, you could have created a unit test strategy that allows you to mock arbitrary file sizes so you can test them with arbitrary file size in kilobyte settings.

The trick, of course, is finding the right layer to wrap/inject your testing.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -155,6 +155,12 @@ public override void Analyze(AnalyzeContext context)
             {
                 try
                 {
+                    var fileInfo = new FileInfo(filePath);
```
michaelcfanning, line 158: >FileInfo [](http://example.com/codeflow?start=39&length=8)

Always be careful when introducing file operations in code! You can't mock properly by calling this code directly. 

You should be operating against a FileSystem instance. You will need to create a helper in the default IFileSystem implementation that accepts a file path and computes the length (i.e., convert this pattern to a simple method)

Src/Sarif.PatternMatcher/SearchSkimmer.cs
michaelcfanning: FileSizeInKilobytes is not a great name. It is really a limit or threshold. For the record, I probably gave this argument a bad name.

Here's a comment that I think addresses

// Ensure that the byte of the file does not exceed the limit set by the
// `file-size-in-kilobytes` argument (which will be -1 if not set, in which
// case all files should be analyzed no matter their size).

Src/ReleaseHistory.md
michaelcfanning, line 21: >Resolve [](http://example.com/codeflow?start=7&length=7)



Resolve `OutofMemoryException` and `NullReferenceException' failures resulting from a failure to honor file size scan limits set by `--file-size-in-kb` argument.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +92,66 @@ public void AnalyzeCommand_IntrafileBasic()
             region.StartColumn.Should().Be(2);
         }
 
+        [Fact]
+        public void AnalyzeCommand_FileSizeInKilobytes()
```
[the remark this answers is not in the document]
marmegh: This test is, but I'll add PR comments to clarify when I push the next update.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+                    fileSize = (long)ulong.MinValue,
+                    maxFileSize = int.MaxValue,
+                    expectedResult = 4
+                },
+                new {
+                    fileSize = randomFileSize,
+                    maxFileSize = int.MaxValue,
+                    expectedResult = 4
+                },
+            };
```
marmegh, line 146: Arrange

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+
+            foreach (var testCase in testCases)
+            {
+                SarifLog logFile = RunAnalyzeCommandWithFileSizeLimits(
+                    maxFileSizeInKilobytes: testCase.maxFileSize,
+                    fileSizeInBytes: testCase.fileSize);
```
marmegh, line 152: Act

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+
+                logFile.Runs.Count.Should().Be(1);
+                logFile.Runs[0].Results.Count.Should().Be(testCase.expectedResult);
```
marmegh, line 155: Assert

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+            sarifLogWithLargeFileExcluded.Runs[0].Results[0].Locations.Should()
+                .BeEquivalentTo(sarifLogOnlySmallFile.Runs[0].Results[0].Locations);
+
+            sarifLogWithLargeFileExcluded.Runs[0].Results[1].Locations.Should()
+                .BeEquivalentTo(sarifLogOnlySmallFile.Runs[0].Results[1].Locations);
+
+            sarifLogWithLargeFileExcluded.Runs[0].Artifacts.Count.Should().Be(1);
+
+            Assert.True(string.Equals(sarifLogWithLargeFileExcluded.Runs[0].Artifacts[0].Location.Uri, smallTargetPath) &&
+                !string.Equals(sarifLogWithLargeFileExcluded.Runs[0].Artifacts[0].Location.Uri, largeTargetPath));
```
marmegh, line 208: Assert

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+                maxFileSizeInKilobytes: 1024,
+                fileSizeInBytes: long.MaxValue);
+
+            SarifLog sarifLogWithLargeFileIncluded = RunAnalyzeCommandWithFileSizeLimits(
+                maxFileSizeInKilobytes: 1024,
+                fileSizeInBytes: 0);
+
+            SarifLog sarifLogOnlySmallFile = RunAnalyzeCommand(
+                definitionsText,
+                fileContents);
```
marmegh, line 185: Act

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
marmegh: See the corresponding PR for the SARIF sdk [#2494](https://github.com/microsoft/sarif-sdk/pull/2494) for where this is now defined.

The actual input is text that the SARIF sdk attempts to convert to an int. When that fails, SPAM returns the following errors message:
`Option 'file-size-in-kb' is defined with a bad format.`
Followed by help content describing intended usage of each option.

Src/Test.UnitTests.Sarif.PatternMatcher/simpleExample.json
```diff
@@ -0,0 +1,967 @@
+{
+  "$schema": "https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json",
```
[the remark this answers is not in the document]
marmegh: No longer needed with the IFileSystem implementation.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
marmegh: We're now covering some of these with already established tests. VS is throwing an exception trying to run code coverage for SPAM. It looks like there may be a custom test setting that conflicts with a resource VS requires to run the analysis.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -89,12 +95,125 @@ public void AnalyzeCommand_IntrafileBasic()
...
+            string largeTargetPath = Path.Combine(rootDirectory, LargeTargetName);
+
+            string definitionsText = GetIntrafileRuleDefinition();
+
+            string fileContents = "unused leading space  \r\n" +
+                                  " secret1              \r\n" +
+                                  " host1                \r\n" +
+                                  " id1                  \r\n" +
+                                  " secret2              \r\n" +
+                                  "unused trailing space \r\n";
```
marmegh, line 173: Arrange

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -291,6 +291,18 @@ internal static void UpdateLevelKind(MatchExpression matchExpression)
         {
             AnalyzeContext context = base.CreateContext(options, logger, runtimeErrors, policy, filePath);
 
+            // Type or member is obsolete.
+#pragma warning disable CS0618
+            if (options.FileSizeInKilobytes != options.MaxFileSizeInKilobytes && options.FileSizeInKilobytes != default)
```
eddynaka, line 296: >FileSizeInKilobytes != default [](http://example.com/codeflow?start=89&length=30)

what would happen if:
FileSizeInKilobyte = 1
MaxFileSizeInKilobyte = 2

Is that correct?
should we generate an exception/validation error?

Src/Test.UnitTests.Sarif.PatternMatcher/SearchSkimmerTests.cs
```diff
@@ -409,12 +422,117 @@ public void SearchSkimmer_ShouldThrowWhenRuleDoesNotHaveAnyRegularExpression()
...
+
+                var context = new AnalyzeContext
+                {
+                    TargetUri = new Uri(filePath),
+                    Logger = logger,
+                    MaxFileSizeInKilobytes = testCase.maxFileSize
+                };
+
+                SearchSkimmer skimmer = CreateSkimmer(definition, fileSystem: mockFileSystem.Object);
+                Exception exception = Record.Exception(() => skimmer.Analyze(context));
```
eddynaka, line 495: >skimmer.Analyze [](http://example.com/codeflow?start=61&length=15)

how are we validating that we got in the code-path where the file was analyzed or not?

Src/Test.UnitTests.Sarif.PatternMatcher/SearchSkimmerTests.cs
```diff
@@ -409,12 +422,117 @@ public void SearchSkimmer_ShouldThrowWhenRuleDoesNotHaveAnyRegularExpression()
...
+            SearchSkimmer skimmer = CreateSkimmer(definition, fileSystem: mockFileSystem.Object);
             Exception exception = Record.Exception(() => skimmer.Analyze(context));
             exception.Should().NotBeNull();
             exception.GetType().Should().Be(typeof(InvalidOperationException));
         }
 
+        [Fact]
+        public void SearchSkimmer_ShouldNotEvaluateTooLargeFiles()
+        {
+            var random = new Random();
```
eddynaka, line 434: >random [](http://example.com/codeflow?start=16&length=6)

we don't need a random file.
just hard-code a reasonable size and use it

Src/Test.UnitTests.Sarif.PatternMatcher/SearchSkimmerTests.cs
```diff
@@ -409,12 +422,117 @@ public void SearchSkimmer_ShouldThrowWhenRuleDoesNotHaveAnyRegularExpression()
...
+            };
+
+            var searchSkimmer = new SearchSkimmer(regexEngine, null, null, searchDefinition);
+            var reportingDescriptor = searchSkimmer as ReportingDescriptor;
+
+            var logger = new TestLogger();
+
+            var context = new AnalyzeContext
+            {
+                TargetUri = new Uri($"file:///c:/{definition.Name}.{definition.FileNameAllowRegex}"),
```
eddynaka, line 522: >TargetUri [](http://example.com/codeflow?start=16&length=9)

be sure to mock this, otherwise it is really accessing the disk

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
marmegh: Also took advantage of Michael's recent update to Program.Main to validate the exact exception by retrieving the instantiated command.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -291,6 +293,35 @@ internal static void UpdateLevelKind(MatchExpression matchExpression)
...
+                        + $"`--file-size-in-kb {options.MaxFileSizeInKilobytes}` were used. "
+                        + "Please remove the obsolete option `--file-size-in-kb`.";
+
+                    var innerException = new ArgumentException(message);
+
+                    ThrowExitApplicationException(context: context, ExitReason.InvalidCommandLineOption, innerException);
+                }
+                else if (options.MaxFileSizeInKilobytes != DefaultMaxFileSizeInKilobytes)
+                {
+                    context.MaxFileSizeInKilobytes = options.FileSizeInKilobytes;
```
eddynaka, line 312: >FileSizeInKilobytes [](http://example.com/codeflow?start=61&length=19)

This should be MaxFileSize and the next check should be FileSize.

My reasoning: if maxFileSizeInKilobytes != default -> you should use it. Else the fileSize is different from default, so, you should use that.

marmegh: Added a test case to cover, too.

eddynaka (approved):
:shipit:

Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
```
eddynaka, line 4: >using System [](http://example.com/codeflow?start=0&length=12)

remove this if we are not using it.

Src/Sarif.PatternMatcher.Cli/Program.cs
```diff
@@ -87,5 +88,37 @@ private static bool IsValidVerbName(string verb)
...
+        {
+            bool hasObsoleteArgument = false;
+            bool hasCurrentFileSizeArg = false;
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                if (args[i] == "--file-size-in-kb")
+                {
+                    hasObsoleteArgument = true;
+                    args[i] = "--max-file-size-in-kb";
```
eddynaka, line 102: >max-file-size-in-kb [](http://example.com/codeflow?start=33&length=19)

we have some md files that shows how to use this. Can you also update to use the new argument, please?

marmegh: I only found one of these and updated accordingly.

PR #622, opened 2022-06-29
Move test API. Fix null dereference in command-line arg parsing.

michaelcfanning:
This work is preparatory to other blocked things. @cfaucon, @suvamM or @marmegh can you make it a priority?

@eddynaka, I know you're busy and your work isn't blocked by what I'm doing, so if you can jump in, great, otherwise don't worry.


Src/Plugins/Tests.Security/SEC101.SecurePlaintextSecretsTests.cs
```diff
@@ -21,5 +23,11 @@ public SecurePlaintextSecretsTests(ITestOutputHelper outputHelper) : base(output
         [Fact]
         public void SecurePlaintextSecrets_EndToEndFunctionalTests()
             => RunAllTests();
+
+        [Fact]
+        public void AllSecurityRules_ShouldHaveValidators()
```
michaelcfanning, line 28: >AllSecurityRules_ShouldHaveValidators [](http://example.com/codeflow?start=20&length=37)

All our end-to-end test files are basically stubs for invoking all tests driven by checked in files, so we should add other tests to it, rather than creating new test source files.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -616,8 +616,8 @@ private void RunMatchExpressionForSingleLineRegexes(AnalyzeContext context, Matc
             if (_validators != null && matchExpression.IsValidatorEnabled)
             {
                 string filePath = context.TargetUri.GetFilePath();
-
-                IEnumerable<ValidationResult> validationResults = _validators.Validate(reportingDescriptor.Name,
+                string ruleName = matchExpression.Name ?? reportingDescriptor.Name;
```
michaelcfanning, line 619: >ruleName [](http://example.com/codeflow?start=23&length=8)

It turns out that we're flowing the wrong identifiers around for a specific kind of utilization of our extensibility model (which is flexible). I have fixed this here but because it is currently an 'offroad' model and we are urgently driving some other work, I did not stop and create the unit tests this warrants. That will be done in a follow on change to a repo that includes this one as a submodule.

Src/Sarif.PatternMatcher.Cli/Program.cs
```diff
@@ -18,11 +18,23 @@ internal static class Program
...
+        internal static Exception RuntimeException;
+
         internal static int Main(string[] args)
         {
             Console.OutputEncoding = Encoding.UTF8;
 
-            args = EntryPointUtilities.GenerateArguments(args, FileSystem, new EnvironmentVariables());
+            try
+            {
+                args = EntryPointUtilities.GenerateArguments(args, FileSystem ?? new FileSystem(), new EnvironmentVariables());
```
michaelcfanning, line 30: >GenerateArguments [](http://example.com/codeflow?start=43&length=17)

This code path falls entirely out of our exception handling infrastructure. That's ok, in the sense that we would fail catastrophically in any case, but we need some machinery to capture the exception for test purposes.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -22,6 +23,26 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
 {
     public class ProgramTests
     {
+        [Fact]
+        public void Program_ResponseFileWorksWithNoExplicitlyAllocatedFileSystem()
```
michaelcfanning, line 27: >Program_ResponseFileWorksWithNoExplicitlyAllocatedFileSystem [](http://example.com/codeflow?start=20&length=60)

Funny corner case. This functionality is highly tested, but all tests pass a FileSystem mock object meaning that data was always non-null. In production, we missed a null deref when that mock object wasn't populated. :) 

This happens all the time, good testing still leaves little gaps, nooks and crannies in the actual production/runtime code paths...

Src/Sarif.PatternMatcher/SearchSkimmer.cs
marmegh: Can you add a TODO comment to track the follow up work is completed?

marmegh:
Completed a cursory review and planning to take another 10-15 minutes to compare using the CLI before and after + step through some of the test changes.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -22,6 +23,26 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
...
+            // successfully processed by the command-line argument parsers 
+            // and the response file was not found. This closes a very narrow
+            // code path in tests where an uninstantiated FileSystem object
+            // causes a null dereference. All other tests for this area provide
+            // a non-null file system object instance.
+            Program.RuntimeException.Should().NotBeNull();
+            Program.RuntimeException.GetType().Should().Be(typeof(FileNotFoundException));
+        }
+
+
```
eddynaka, line 45: remove this empty line.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
michaelcfanning: Created a work item for this and sent you a message. I don't want to add a TODO comment here because the code is presumed correct. :) The tests will actually be created in a separate workflow.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/ProgramTests.cs
```diff
@@ -22,6 +23,26 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
...
+            result.Should().Be(1);
+
+            // These conditions indicate that the response file argument was
+            // successfully processed by the command-line argument parsers 
+            // and the response file was not found. This closes a very narrow
+            // code path in tests where an uninstantiated FileSystem object
+            // causes a null dereference. All other tests for this area provide
+            // a non-null file system object instance.
+            Program.RuntimeException.Should().NotBeNull();
+            Program.RuntimeException.GetType().Should().Be(typeof(FileNotFoundException));
```
eddynaka, line 42: >FileNotFoundException [](http://example.com/codeflow?start=66&length=21)

not sure why but when we run this using the commandline, the tests don't pass.

Maybe debug vs release?

eddynaka: We saw that the real issue was how the order of things were happening. To enforce that the behavior will be always the same, michael pushed a change enforcing FileSystem = null.

eddynaka (approved):
:shipit:

PR #623, opened 2022-07-06
Make search definitions argument required.

michaelcfanning:
I previously added a new helper to the SARIF Driver SDK to report out when no plugins are configured. Today it occurred to me that the right fix is simply to make this command-line argument required. Oops. So let's review this change and then I'll go rip out that other helper again. 

The heart of this change is simply to add `Required = true` to the `OptionAttribute` backing the search definitions argument. I've added a unit test. This change includes a useful test method I intend to exercise later.

Also, I've incidentally updated SARIF-SDK submodule.

### DIFF
--- a/Src/Sarif.PatternMatcher.Cli/Program.cs
+++ b/Src/Sarif.PatternMatcher.Cli/Program.cs
@@ -65,9 +65,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
                         : CommandBase.FAILURE);
         }
 
-        private static int RunAnalyzeCommand(AnalyzeOptions options)
+        internal static int RunAnalyzeCommand(AnalyzeOptions options)
         {
-            InstantiatedAnalyzeCommand = new AnalyzeCommand();
+            InstantiatedAnalyzeCommand = new AnalyzeCommand(fileSystem: FileSystem);
             return InstantiatedAnalyzeCommand.Run(options);
         }
 
--- a/Src/Sarif.PatternMatcher.Cli/Program.cs
+++ b/Src/Sarif.PatternMatcher.Cli/Program.cs
@@ -21,6 +21,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
         [ThreadStatic]
         internal static Exception RuntimeException;
 
+        [ThreadStatic]
+        internal static AnalyzeCommand InstantiatedAnalyzeCommand;
+
         internal static int Main(string[] args)
         {
             Console.OutputEncoding = Encoding.UTF8;
@@ -53,13 +56,19 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
               .MapResult(
                 (AnalyzeDatabaseOptions options) => new AnalyzeDatabaseCommand().Run(options),
                 (ImportAndAnalyzeOptions options) => new ImportAndAnalyzeCommand().Run(options),
-                (AnalyzeOptions options) => new AnalyzeCommand(FileSystem).Run(options),
+                (AnalyzeOptions options) => RunAnalyzeCommand(options),
                 (ExportRulesMetatadaOptions options) => new ExportRulesMetatadaCommand().Run(options),
                 (ExportSearchDefinitionsOptions options) => new ExportSearchDefinitionsCommand().Run(options),
                 (ValidateOptions options) => new ValidateCommand().Run(options),
                 _ => isValidHelpCommand || isVersionCommand
                         ? CommandBase.SUCCESS
                         : CommandBase.FAILURE);
+        }
+
+        private static int RunAnalyzeCommand(AnalyzeOptions options)
+        {
+            InstantiatedAnalyzeCommand = new AnalyzeCommand();
+            return InstantiatedAnalyzeCommand.Run(options);
         }
 
         private static bool IsValidVerbName(string verb)
--- a/Src/Sarif.PatternMatcher.Cli/Program.cs
+++ b/Src/Sarif.PatternMatcher.Cli/Program.cs
@@ -65,6 +65,13 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
                         : CommandBase.FAILURE);
         }
 
+        internal static void ClearUnitTestData()
+        {
+            FileSystem = null;
+            RuntimeException = null;
+            InstantiatedAnalyzeCommand = null;
+        }
+
         internal static int RunAnalyzeCommand(AnalyzeOptions options)
         {
             InstantiatedAnalyzeCommand = new AnalyzeCommand(fileSystem: FileSystem);
--- a/Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
+++ b/Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
@@ -29,6 +29,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
                     $"-o", Guid.NewGuid().ToString(),
             };
 
+            Program.ClearUnitTestData();
             int result = Program.Main(args);
             result.Should().Be(CommandBase.FAILURE);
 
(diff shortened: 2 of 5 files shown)


Src/Sarif.PatternMatcher.Cli/Program.cs
```diff
@@ -53,7 +56,7 @@ internal static int Main(string[] args)
               .MapResult(
                 (AnalyzeDatabaseOptions options) => new AnalyzeDatabaseCommand().Run(options),
                 (ImportAndAnalyzeOptions options) => new ImportAndAnalyzeCommand().Run(options),
-                (AnalyzeOptions options) => new AnalyzeCommand(FileSystem).Run(options),
+                (AnalyzeOptions options) => RunAnalyzeCommand(options),
```
michaelcfanning, line 59: >RunAnalyzeCommand [](http://example.com/codeflow?start=44&length=17)

So here's a proposal for a test pattern as people add unit tests. The condition I need to check provokes a failure at the CommandLine library level (i.e., the analyze command is never invoked).

So I've created a pattern where on actually calling into the analyze command, we retain the command instance. For this case, a null value means we failed core command-line parsing, as expected. Moving forward, this pattern will actually enable unit-testing at a much higher level, you can simply pass test args to Main and do the work. This increase confidence in testing as more product code is actually utilized. Lots of helper code could wither away with this approach.

I'm putting out this single change to experiment/refine further. i.e., not over-investing in covering existing code. As usual, for addressing technical debt, my common approach is to work on things are they churn. Having landed on a good approach we could consider going back for scrubs, if helpful.

Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
```diff
@@ -23,6 +26,24 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public class AnalyzeCommandTests
     {
+        private static AnalyzeOptions CreateDefaultAnalyzeOptions()
```
michaelcfanning, line 29: >CreateDefaultAnalyzeOptions [](http://example.com/codeflow?start=38&length=27)

This is a useful helper that uses reflection against the product code to create an AnalyzeOptions instance that is populated with all product default values.

This will prevent errors due to dueling initialization code paths. This code isn't used in this test file, currently, but I intend to come back here over the next week and do other clean-up, so I'm adding it. A lot of code here can be simplified and improved.

michaelcfanning:
>             IEnumerable<Skimmer<AnalyzeContext>> applicableSkimmers = PatternMatcher.AnalyzeCommand.DetermineApplicabilityForTargetHelper(context, skimmers, disabledSkimmers);

Look at how horrible this test is. It proposes significant, entirely customized execution logic, simply because we didn't have a clean way previously to mock, retrieve state to validate post-processing, etc. My sense is that we've made enough progress here to do a better job and I'll look at this soon.

---
Refers to: Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs:132 in bf83cb5. [](commit_id = bf83cb59f13eecdfac545e8a9e7c0684d1c7fbb2, deletion_comment = False)

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -17,6 +19,26 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
 {
     public class AnalyzeCommandTests
     {
+        [Fact]
+        public void AnalyzeCommand_DefinitionsArgumentIsRequired()
```
marmegh, line 23: >public void AnalyzeCommand_DefinitionsArgumentIsRequired() [](http://example.com/codeflow?start=8&length=58)

This test isn't currently passing via the PR build pipeline or locally (using BuildAndTest.cmd).

Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
```diff
@@ -23,6 +26,24 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
...
+        {
+            var result = new AnalyzeOptions();
+            Type analyzeOptionsType = typeof(AnalyzeOptions);
+
+            foreach (PropertyInfo property in analyzeOptionsType.GetProperties())
+            {
+                var optionAttribute = (OptionAttribute)property.GetCustomAttribute(typeof(OptionAttribute));
+                if (optionAttribute == null || optionAttribute.Default == null)
+                { 
+                    continue;
```
marmegh, line 39: >  [](http://example.com/codeflow?start=29&length=1)

Trailing whitespace.

Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
```diff
@@ -23,6 +26,24 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
...
+        private static AnalyzeOptions CreateDefaultAnalyzeOptions()
+        {
+            var result = new AnalyzeOptions();
+            Type analyzeOptionsType = typeof(AnalyzeOptions);
+
+            foreach (PropertyInfo property in analyzeOptionsType.GetProperties())
+            {
+                var optionAttribute = (OptionAttribute)property.GetCustomAttribute(typeof(OptionAttribute));
+                if (optionAttribute == null || optionAttribute.Default == null)
+                {
```
marmegh, line 38: >  [](http://example.com/codeflow?start=17&length=1)

Trailing whitespace.

Src/Sarif.PatternMatcher.Cli/Program.cs
marmegh: Something else that's pretty great about this approach is you can check/validate the exit reason.

michaelcfanning: Yes, exactly, that's we 'poke' the command into static data, so you can check the RuntimeErrors, look at any unhandled exception that was raised, etc.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -17,6 +19,27 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
...
+            Program.ClearUnitTestData();
+            int result = Program.Main(args);
+            result.Should().Be(CommandBase.FAILURE);
+
+            // This validation is sufficient because the null check for an
+            // instantiated analyze command verifies that we failed the 
+            // CommandLine parsing code and error out before attempting
+            // analysis.
+            Program.InstantiatedAnalyzeCommand.Should().BeNull();
+        }
```
marmegh, line 41: Did this test fail before the change?

Using `spam analyze test.txt -o test2.txt` here are the before and after results:
* before: ```Analyzing...
SPAM : error ERR997.AllRulesExplicitlyDisabled

Analysis did not complete due to one or more unrecoverable execution conditions.
Unexpected fatal runtime condition(s) observed: NoRulesLoaded```
* after: ```
Analyzing...
SPAM : error ERR997.AllRulesExplicitlyDisabled : All rules were explicitly disabled so there is no work to do.

Analysis did not complete due to one or more unrecoverable execution conditions.
Unexpected fatal runtime condition(s) observed: NoRulesLoaded```

michaelcfanning: Can you run your test again after syncing? I really appreciate this in-depth examination, it's super helpful.

What you should see is that CommandLine lib parsing should error out, w/something like so.

I have to check in to unblock other work, but please do comment if I've missed something and I'll fold changes into other work lighting up today.

ERROR(S):
  Required option 'd, search-definitions' is missing.

marmegh: Well, this is interesting. Two scenarios get the expected result and two get the previous result. Unit test and running via debugger: exactly as expected. Use the command line (same debug output folder debugger runs off of or release): get the previous result.

PR #634, opened 2022-08-08
Update `Debugging Guide` for build setup

marmegh:
## Changes

Updated debugging guide to remove reference to Visual Studio 2019 and to add instructions to populate the debug build folder by running a debug build from within Visual Studio.

Screenshot from VS Code:
image

Release history and unit tests changes are not required as this is just a documentation update.


docs/DebuggingGuide.md
```diff
@@ -30,7 +30,13 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
     `git checkout -b informative-branch-name`
 
 ## Debugging
-0. Install the VS SARIF viewer for [VS 2019](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer) or [VS 2022](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer2022).
+0. Install the VS SARIF viewer for [VS 2022](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer2022).
```
marmegh, line 33: Removed the reference to VS 2019 per discussion during standup this morning.

docs/DebuggingGuide.md
```diff
@@ -30,7 +30,13 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
...
+   
+   Select `Debug` from the `Solution Configurations` menu:
+
+   ![Solution Configurations: Debug](DebugBuild.png)
+   
+   Then run `Build Solution` from the `Build` menu.
```
marmegh, line 39: @HulonJenkins, let me know if the image is helpful here. Also, if you want to check formatting for a documentation update (ie. when someone adds an image), it can be helpful to enlist in their branch and open in VS Code. Alternatively, you can ask for someone to add a screenshot of the result in the description.

docs/DebuggingGuide.md
```diff
@@ -21,7 +21,7 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
 
     `git clone https://github.com/YOURGITHUBACCOUNT/sarif-pattern-matcher`
 
-2. Run BuildAndTest.exe from the root of the enlistment. This will properly load all submodules and validate some machine configuration/settings. Complete any recommended installs/upgrades (e.g., to install a newer sdk via VS installer individual components):
+2. Run BuildAndTest.cmd from the root of the enlistment. This will properly load all submodules and validate some machine configuration/settings. Complete any recommended installs/upgrades (e.g., to install a newer sdk via VS installer individual components):
 
     `D:\src\sarif-pattern-matcher> BuildAndTest.cmd`
```
HulonJenkins, line 26: Worked for me up to this point! Also testing if I can write a comment on a PR

docs/DebuggingGuide.md
```diff
@@ -21,7 +21,7 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
 
     `git clone https://github.com/YOURGITHUBACCOUNT/sarif-pattern-matcher`
 
-2. Run BuildAndTest.exe from the root of the enlistment. This will properly load all submodules and validate some machine configuration/settings. Complete any recommended installs/upgrades (e.g., to install a newer sdk via VS installer individual components):
+2. Run BuildAndTest.cmd from the root of the enlistment. This will properly load all submodules and validate some machine configuration/settings. Complete any recommended installs/upgrades (e.g., to install a newer sdk via VS installer individual components):
```
michaelcfanning, line 24: >BuildAndTest [](http://example.com/codeflow?start=7&length=12)

Why are we building twice? i.e., once in #2 and again in #0? Maybe we can just indicate that the user should build again in the IDE.

docs/DebuggingGuide.md
```diff
@@ -30,7 +30,13 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
     `git checkout -b informative-branch-name`
 
 ## Debugging
-0. Install the VS SARIF viewer for [VS 2019](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer) or [VS 2022](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer2022).
+0. Install the VS SARIF viewer for [VS 2022](https://marketplace.visualstudio.com/items?itemName=WDGIS.MicrosoftSarifViewer2022).
+   
+   Select `Debug` from the `Solution Configurations` menu:
```
michaelcfanning, line 35: >Select [](http://example.com/codeflow?start=3&length=6)

Why are we insisting on the debug solution configuration? Users can debug release as well. The guidance as stated here actually builds release in step #2 and then debug here, which seems a bit duplicative. We could just advise the users to `Build Solution`. `Debug` will be the default for them in any case.

Alternately you could drop the running of BuildAndTest.cmd from the root of the enlistment and just advice jumping into the IDE, selecting `Debug` and building.

In general, a debug build does do better when debugging mostly due to the absence of certain code optimizations that cause unexpected behaviors in the IDE (such as not being able to set breakpoints in some places).

michaelcfanning (approved):
:shipit:

docs/DebuggingGuide.md
HulonJenkins, line 39: >Then run [](http://example.com/codeflow?start=3&length=8)

Not sure if this is even relevant or important at all, but when opening VS a CMake prompt pops up. I hit "do not enable", but if this is necessary to note, then a clarification here would be useful to a new person.

HulonJenkins: It helped me realize I didn't have the solution open once I opened the folder in VS. I made a separate comment about this issue though.

PR #635, opened 2022-08-08
Avoid special characters error when following DebuggingGuide.md

HulonJenkins:
## Changes

Updated DebuggingGuide.md to clarify instructions to avoid path name misinterpretation error from newline, tab, or carriage return characters.


docs/DebuggingGuide.md
```diff
@@ -45,7 +45,7 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
...
+3. Create a response file (e.g., in a file named `d:\testFiles\SpamArguments.rsp`) that contains the following analysis arguments (it is fine for a response file to break its arguments across multiple lines, just as they are here). Note that in this example `d:\src\sarif-pattern-matcher\` refers to the root of the `sarif-pattern-enlistment`. Also note that the .rsp file should NOT be in a directory that contains /n, /t, or /r as this will cause incorrect behavior when parsing the path name containing a newline, tab, or carriage return character and won't allow launching of the debugging session. The output location referenced here directs SARIF log file output to a special directory that the VS SARIF viewer extension watches. VS will automatically load and display any SARIF results written to this special location (i.e., a directory named `\.sarif\` that exists alongside the current VS solution file).
```
marmegh, line 48: >Also note that the .rsp file should NOT be in a directory that contains /n, /t, or /r as this will cause incorrect behavior when parsing the path name containing a newline, tab, or carriage return character and won't allow launching of the debugging session.  [](http://example.com/codeflow?start=344&length=259)

Highlighting the only change for all reviewers.

```
Also note that the .rsp file should NOT be in a directory that contains /n, /t, or /r as this will cause incorrect behavior when parsing the path name containing a newline, tab, or carriage return character and won't allow launching of the debugging session.
```

marmegh: The sentence reads like it is missing punctuation.

Perhaps we can clean this up a little?

```
Also note, avoid referencing response files in any directory beginning with `/n`, `/t`, or `/r` (which will be translated to newline, tab, or carriage return) as this will cause the debugging session to fail on launch.
```

marmegh (approved):
:shipit:

docs/DebuggingGuide.md
```diff
@@ -45,7 +45,7 @@ Additional setup guidance can be found in the [Contributing.md](https://github.c
...
+3. Create a response file (e.g., in a file named `d:\testFiles\SpamArguments.rsp`) that contains the following analysis arguments (it is fine for a response file to break its arguments across multiple lines, just as they are here). Note that in this example `d:\src\sarif-pattern-matcher\` refers to the root of the `sarif-pattern-enlistment`. Also note that the .rsp file should NOT be in a directory that contians '/n', '/t', or '/r' (a newline, tab, or carriage return character). This will cause incorrect behavior when parsing the path name and won't allow launching of the debugging session. The output location referenced here directs SARIF log file output to a special directory that the VS SARIF viewer extension watches. VS will automatically load and display any SARIF results written to this special location (i.e., a directory named `\.sarif\` that exists alongside the current VS solution file).
```
marmegh, line 48: >NOT [](http://example.com/codeflow?start=380&length=3)

I object to the use of the word `not` when advising someone, but acknowledge this is the clearest documentation approach. While most people will struggle to internalize the advice on first read of this doc; they will at least have a clear path to understanding the error if/when it occurs. No change needed, this is just a note about how the audience for this document may struggle with information assimilation due to some common cognitive lapses.

PR #644, opened 2022-08-23
Aligning package reference

eddynaka:
## Changes

This change will align the newtonsoft.json package to be the same as internal references.
Consumers of the library must use a more secure version (13.0.1).

It also changes the Sarif.PatternMatcher projects to depend on RE2.Managed/Strings.Interop package to prevent some warnings when building in the cmd/vs.

### DIFF
--- a/Src/Plugins/Tests.Security/EndToEndTests.cs
+++ b/Src/Plugins/Tests.Security/EndToEndTests.cs
@@ -104,8 +104,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                 {
                     TargetUri = new Uri(filePath, UriKind.Absolute),
                     FileContents = logContents,
-                    Logger = logger,
-                    MaxFileSizeInKilobytes = 1024,
+                    Logger = logger
                 };
 
                 using (context)
--- a/Src/ReleaseHistory.md
+++ b/Src/ReleaseHistory.md
@@ -8,7 +8,7 @@
 - SDK: Sarif.PatternMatcher projects will start using a fixed version of `RE2.Managed` and `Strings.Interop`. [#638](https://github.com/microsoft/sarif-pattern-matcher/pull/638)
 - Bump Microsoft.Security.Utilities from 1.1.0 to 1.3.0. [#642](https://github.com/microsoft/sarif-pattern-matcher/pull/642)
 - BUG: Fix false negative when `SearchSkimmer` is invoked directly and `MaxFileSizeInKilobytes` is not set. This will now default to approximately 10MB. [#637](https://github.com/microsoft/sarif-pattern-matcher/pull/637)
-- Dependency alignment of `Newtonsoft.Json` to 12.0.3. [#644](https://github.com/microsoft/sarif-pattern-matcher/pull/644)
+- BUG: Loosen `Newtonsoft.Json` minimum version requirement to 12.0.3 for `Sarif.PatternMatcher` project. [#644](https://github.com/microsoft/sarif-pattern-matcher/pull/644)
 
 ## *v1.10.0*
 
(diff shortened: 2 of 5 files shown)


Src/Sarif.PatternMatcher.Cli/Sarif.PatternMatcher.Cli.csproj
```diff
@@ -19,6 +19,7 @@
   <ItemGroup>
     <PackageReference Include="CommandLineParser" Version="2.7.82" />
     <PackageReference Include="Microsoft.Data.Sqlite" Version="5.0.5" />
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
eddynaka, line 22: >Newtonsoft.Json [](http://example.com/codeflow?start=31&length=15)

This will guarantee that our tool is safe to use.

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -104,8 +104,7 @@ protected override string ConstructTestOutputFromInputResource(string inputResou
                 {
                     TargetUri = new Uri(filePath, UriKind.Absolute),
                     FileContents = logContents,
-                    Logger = logger,
-                    MaxFileSizeInKilobytes = 1024,
```
eddynaka, line 108: >MaxFileSizeInKilobytes = 1024 [](http://example.com/codeflow?start=20&length=29)

Since our default value changed to 10MB, we don't need to set this.

Src/ReleaseHistory.md
```diff
@@ -8,6 +8,7 @@
 - SDK: Sarif.PatternMatcher projects will start using a fixed version of `RE2.Managed` and `Strings.Interop`. [#638](https://github.com/microsoft/sarif-pattern-matcher/pull/638)
 - Bump Microsoft.Security.Utilities from 1.1.0 to 1.3.0. [#642](https://github.com/microsoft/sarif-pattern-matcher/pull/642)
 - BUG: Fix false negative when `SearchSkimmer` is invoked directly and `MaxFileSizeInKilobytes` is not set. This will now default to approximately 10MB. [#637](https://github.com/microsoft/sarif-pattern-matcher/pull/637)
+- Dependency alignment of `Newtonsoft.Json` to 12.0.3. [#644](https://github.com/microsoft/sarif-pattern-matcher/pull/644)
```
marmegh, line 11: >Dependency alignment of `Newtonsoft.Json` to 12.0.3. [](http://example.com/codeflow?start=2&length=52)

Is this dependency alignment visible to someone using spam? Are there customer use cases where they would be integrating directly with the Sarif.PatternMatcher.dll as opposed to the CLI?
Also, this suggests that the CLI is also downgraded to 12.0.3 as written.

Src/Sarif.PatternMatcher/Sarif.PatternMatcher.csproj
```diff
@@ -10,18 +10,16 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
```
marmegh, line 13: ><PackageReference Include="Newtonsoft.Json" Version="12.0.3" /> [](http://example.com/codeflow?start=4&length=63)

This may retrigger the Newtonsoft alert and dependabot.

marmegh, line 13: >12.0.3 [](http://example.com/codeflow?start=57&length=6)

Reading the description, I don't see an explanation of the reason for this change. Is this to prevent a conflict with the versions of RE2.Managed and Strings.Interop? How does this affect package flattening?

eddynaka: Observed an issue with Hulon when compiling in vs/cmd which shows many warnings.

With that, I want to fix this and make things cleaner :)

eddynaka: Actually, I wrote it in the description: `It also changes the Sarif.PatternMatcher projects to depend on RE2.Managed/Strings.Interop package to prevent some warnings when building in the cmd/vs.`

marmegh: So, the part that is missing from the description is that this is to cleanup build warnings by aligning newtonsoft.json package between spam and internal references. Cool.

marmegh (approved):
:shipit:

PR #645, opened 2022-08-31
Add test to enforce expected result levels

marmegh:
## Changes

Created test to reproduce bug observed in feature testing, where the `FailureLevel` for static analysis results was overridden to always return `FailureLevel.Note` unless dynamic analysis was also completed.

Removed the single line causing this bug and observed the new test passed with expected SARIF results generated.

This is a breaking change because it is likely to result in new findings, as demonstrated by the need to rebaseline.

* [x] Added unit tests


Src/ReleaseHistory.md
```diff
@@ -1,5 +1,8 @@
 # Release History
 
+## Unreleased
+- BREAKING: Fix bug resulting in static validators returning `FailureLevel.Note` despite configured `FailureLevel`. [#645](https://github.com/microsoft/sarif-pattern-matcher/pull/645)
+
```
marmegh, line 5: Release note for this change.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -28,7 +28,7 @@ public AnalyzeCommand(IFileSystem fileSystem = null)
         {
             engine ??= RE2Regex.Instance;
 
-            var validators = new ValidatorsCache();
```
marmegh, line 31: >var validators = new ValidatorsCache(); [](http://example.com/codeflow?start=12&length=39)

Extending our ability to mock throughout the targeted code path.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -364,7 +364,6 @@ public override void Analyze(AnalyzeContext context)
 
                 case ValidationState.Unknown:
                 {
-                    level = FailureLevel.Note;
```
marmegh, line 367: >level = FailureLevel.Note; [](http://example.com/codeflow?start=20&length=26)

Removing the overwriting of the failure level if only static validation has completed. This is the actual code change for this PR.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -529,5 +742,50 @@ private static string GetSingleLineRuleDefinition()
 
             return JsonConvert.SerializeObject(definitions);
         }
+
+        private static string GetSingleLineRuleDefinitionFailureLevel(FailureLevel level)
```
marmegh, line 746: >GetSingleLineRuleDefinitionFailureLevel [](http://example.com/codeflow?start=30&length=39)

This method generates the shared strings and json files needed to support end to end testing of static and dynamic validation. These files will then be returned by the mocked file system during the relevant tests. This allowed for the cleanest repro of the conditions observed in the reported bug.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -453,6 +548,124 @@ private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
...
+            else if (sarifLog.Runs[0]?.Results[0]?.Level != expectedFailureLevel)
+            {
+                string message = $"Expected `FailureLevel` to be `{expectedFailureLevel}` but found " +
+                    $"{sarifLog.Runs[0]?.Results[0]?.Level} for `{validationScenario}` test scenario {testScenarioMode}.";
+
+                StringBuilderFormatAndAppendNewLine(message, ref stringBuilder);
+            }
+        }
+
+        private void StringBuilderFormatAndAppendNewLine(string data, ref StringBuilder stringBuilder)
```
marmegh, line 660: >StringBuilderFormatAndAppendNewLine [](http://example.com/codeflow?start=21&length=35)

This is the first step in a multi-step process of streamlining test validation. Ultimately, this method and similar will be moved to security utilities to facilitate a common test assertion pattern across all team test projects. Consider this the proof of concept to try out in this test project until adoption.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -453,6 +548,124 @@ private SarifLog RunAnalyzeCommand(string definitionsText, string fileContents)
...
+                if (File.Exists(sarifLogFileName))
+                {
+                    File.Delete(sarifLogFileName);
+                }
+            }
+
+            return sarifLog;
+        }
+
+        private void ValidateFailureLevelByValidationType(
```
marmegh, line 635: >ValidateFailureLevelByValidationType [](http://example.com/codeflow?start=21&length=36)

Split out because this validation was used more than once and the potential for use in other tests. Again, this is part of an effort to demonstrate one option in standardizing our test assertions.

marmegh:
There is a significant amount of churn with this PR. Key changes have been highlighted with comments. A number of sarif files were rebaselined and at least one unit test required updating to account for a change from "note" for static analysis only to the default level found in the corresponding json.

michaelcfanning:
>               " (no validation occurred as it was not enabled. Pass '--dynamic-validation' on the command-line to validate this match)",

Something looks wrong here. This message appears to indicate there is a dynamic validator available but it wasn't enabled. This case should never return an 'error', that case should return a 'warning'. ?

---
Refers to: Src/Plugins/Tests.AzureDevOpsConfiguration/TestData/ServiceConnectionSecurity/ExpectedOutputs/AZC102_190.serviceconnection.DoNotShareWithAllPipelines.sarif:59 in 0ba90a7. [](commit_id = 0ba90a73dfc00af3d4f2b60b0a6dc9d99f8de955, deletion_comment = False)

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_006.GitHubPat.sarif
```diff
@@ -48,7 +48,6 @@
         {
           "ruleId": "SEC101/006",
           "ruleIndex": 0,
-          "level": "note",
```
michaelcfanning, line 51: >note [](http://example.com/codeflow?start=20&length=4)

This delta makes sense. Note disappears, meaning that we now have a warning (by default). A rule with a dynamic validator should return a warning when dynamic analysis is not enabled.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
michaelcfanning: Excellent.

You can reduce this call to simply pass 'fileSystem', however.

The redundancy in 'fileSystem: fileSystem' doesn't encode any additional information. 

DRY :)

marmegh:
>               " (no validation occurred as it was not enabled. Pass '--dynamic-validation' on the command-line to validate this match)",

This matches the configuration in the `AZC102.ServiceConnectionSecurity.json`:

```
"Id": "AZC102",
"Name": "ServiceConnectionShouldBeSecure",
"Level": "Error",
```

---
In reply to: [1242508967](https://github.com/microsoft/sarif-pattern-matcher/pull/645#issuecomment-1242508967) [](http://example.com/codeflow?ancestors=1242508967,1242508967)

---
Refers to: Src/Plugins/Tests.AzureDevOpsConfiguration/TestData/ServiceConnectionSecurity/ExpectedOutputs/AZC102_190.serviceconnection.DoNotShareWithAllPipelines.sarif:59 in 0ba90a7. [](commit_id = 0ba90a73dfc00af3d4f2b60b0a6dc9d99f8de955, deletion_comment = False)

Src/Sarif.PatternMatcher/SearchSkimmer.cs
michaelcfanning: Ah, but your change is incomplete, I think. You found the reason that this data was getting clobbered. You did *not*, however, distinguish handling between validators that do and do *not* have a dynamic validator and my supposition is that these cases aren't identical.

michaelcfanning:
>                             validatorMessage += " " + DynamicValidationNotEnabled;

In this code path, I *believe* you need to set the failure level explicitly to 'Warning'. If I'm correct, that raises the question, why didn't your test uncover this problem? 

1) test gap?
2) matrix not correct?
3) @michaelcfanning is wrong in his analysis. <g>

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:377 in 0ba90a7. [](commit_id = 0ba90a73dfc00af3d4f2b60b0a6dc9d99f8de955, deletion_comment = False)

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -277,6 +280,98 @@ public void AnalyzeCommand_ShouldThrowExceptionForConflictingFileSizeOptions()
...
 
+
+        [Fact]
+        public void AnalyzeCommand_FailureLevelShouldBeDefaultUnlessDynamicallyValidated()
+        {
+            var sb = new StringBuilder();
+
+            var testCases = new[] {
+                new {
+                    jsonLevel = FailureLevel.Note,
```
michaelcfanning, line 291: These aren't great names, I don't understand them clearly. Maybe you can try making them more descriptive? You could have added a code comment for it?

IOW, I can't easily review your matrix to see whether I agree that it is complete and correct. If I'm having trouble, it is extremely likely that a new engineer on the project would be entirely lost. :)

michaelcfanning, line 291: >jsonLevel [](http://example.com/codeflow?start=20&length=9)

OK, reading the rest of the test code, I get it. Try the following:

'jsonLevel': 'failureLevelConfiguredInDefinitionsJson'
'dynamicValidationEnabled' -> 'dynamicValidationEnabledOnCommandLine'
'expectedNoValidation' -> 'expectedFailureLevelIfNoValidatorsExist'
'expectedStaticOnly' -> 'expectedFailureLevelIfOnlyStaticValidatorExists'
expectedStaticDynamic -> 'expectedFailureLevelIfStaticAndDynamicValidatorExists'

These are long but descriptive and precise. You can take the expedient route of creating more concise names, but only if you please author informative code comments. I note again here that I have seen very few useful code comments in the PR's coming through our pipeline. :) Try to find the right ones to author.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -277,6 +280,98 @@ public void AnalyzeCommand_ShouldThrowExceptionForConflictingFileSizeOptions()
...
+                    expectedStaticOnly = FailureLevel.Warning,
+                    expectedStaticDynamic = FailureLevel.Error
+                },
+                new {
+                    jsonLevel = FailureLevel.Error,
+                    dynamicValidationEnabled = true,
+                    expectedNoValidation = FailureLevel.Error,
+                    expectedStaticOnly = FailureLevel.Error,
+                    expectedStaticDynamic = FailureLevel.Error
+                }
```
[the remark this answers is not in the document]
michaelcfanning: Yes! And this matrix looks good. So I'm confused as to why we're seeing the log file output I noticed. What condition are we hitting? Is it possible we have a different bug where the literal failure levels are correct but we're incorrectly emitting 'dynamic analysis was not enabled' for a rule that doesn't even have a dynamic validator?

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -277,6 +280,98 @@ public void AnalyzeCommand_ShouldThrowExceptionForConflictingFileSizeOptions()
...
+
+            foreach (var testCase in testCases)
+            {
+                string definitionsText = GetSingleLineRuleDefinitionFailureLevel(testCase.jsonLevel);
+                string testScenarioName = "NoValidation";
+                SarifLog logFile = RunAnalyzeCommandWithDynamicValidation(definitionsText,
+                                                                          testScenarioName,
+                                                                          testCase.dynamicValidationEnabled);
+
+                ValidateFailureLevelByValidationType(expectedFailureLevel: testCase.expectedNoValidation,
```
michaelcfanning, line 342: >expectedFailureLevel [](http://example.com/codeflow?start=53&length=20)

please consider my suggested renames. if you took the suggestion, i think you'd agree that the test case property name itself is sufficient documentation of its purpose and you wouldn't need a named parameter. that would be the goal.

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -277,6 +280,98 @@ public void AnalyzeCommand_ShouldThrowExceptionForConflictingFileSizeOptions()
...
+                                                     sarifLog: logFile,
```
michaelcfanning, line 343: >sarifLog: logFile [](http://example.com/codeflow?start=53&length=17)

same note. does these duplicated names, 'sarifLog: logfile' help use? can you collapse them?

for example, what if you named your local variable 'sarifLog'?

in general, your goal should be for both your parameter names in callees and call sites to be sufficiently descriptive to be standalone. with the result that you do *not* have to provide a named argument in order for the code to be readable. there's certainly no excuse for that when you own both the variable names and the called methods, as here. :)

the named arguments are most helpful when providing standalone literals and when wiring up other data where you don't own the name...

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -529,5 +742,50 @@ private static string GetSingleLineRuleDefinition()
...
+                        Name = "FailureLevelTest", Id = "TEST001",
+                        Level = level,
+                        Message = "A problem occurred in '{0:scanTarget}'.",
+                        Description = "Failure Level Testing Rules",
+                        MatchExpressions = new List<MatchExpression>(new[]
+                        {
+                            new MatchExpression()
+                            {
+                                Id = "TEST001/001",
+                                Name ="NoValidation",
```
michaelcfanning, line 768: >NoValidation [](http://example.com/codeflow?start=39&length=12)

Your test names do not encode sufficient information, please try to provide a more descriptive scenario.

'DoesNotWork',
'NoValidation'
'FailsQuickly'

these are general descriptions. 

'StaticOnly' static what? Pretend you are new to the team, what would you make of this?

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
michaelcfanning, line 635: >ValidateFailureLevelByValidationType [](http://example.com/codeflow?start=21&length=36)

this isn't a great method name. what is a 'validation type'?

are you really talking about a 'type' here (recall that it's a very overloaded term). you are really talking about a rule implementation detail right?

Also you've duplicated the term 'validate' to refer to your testing but it's also our term for the thing under test. Confusing.

What you're actually testing is the behavior of analysis returning (or not) configured rule failure levels, yes?

Want to try for a better name? :) want me to suggest one?

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
michaelcfanning, line 660: >ref [](http://example.com/codeflow?start=70&length=3)

why is this a ref arg? it doesn't need to be. you don't test it for null (and create one if it is null).

Src/Plugins/Tests.AzureDevOpsConfiguration/TestData/ServiceConnectionSecurity/ExpectedOutputs/AZC102_190.serviceconnection.DoNotShareWithAllPipelines.sarif
```diff
@@ -48,7 +48,7 @@
         {
           "ruleId": "AZC102/190",
           "ruleIndex": 0,
-          "level": "note",
+          "level": "error",
```
michaelcfanning, line 51: >error [](http://example.com/codeflow?start=20&length=5)

So, yes, this is wrong and it's the bug I pointed out. This rule has a static/dynamic validator but dynamic analysis isn't enabled. So this level *should* have been warning but it is not. I pointed out the code change that I think is responsible for this.

But now we have to ask? why are your tests passing for this case? :) something isn't right, can you figure out what it is?

michaelcfanning (changes requested):
:clock1:

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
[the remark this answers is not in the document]
marmegh: AZC102_190 does actually have a dynamic validator. Please see the `AZC102_190.DoNotGrantAllPipelinesAccessValidator`

marmegh:
>               " (no validation occurred as it was not enabled. Pass '--dynamic-validation' on the command-line to validate this match)",

There is a dynamic validator for this class.

---
In reply to: [1242511201](https://github.com/microsoft/sarif-pattern-matcher/pull/645#issuecomment-1242511201) [](http://example.com/codeflow?ancestors=1242511201,1242508967)

---
Refers to: Src/Plugins/Tests.AzureDevOpsConfiguration/TestData/ServiceConnectionSecurity/ExpectedOutputs/AZC102_190.serviceconnection.DoNotShareWithAllPipelines.sarif:59 in 0ba90a7. [](commit_id = 0ba90a73dfc00af3d4f2b60b0a6dc9d99f8de955, deletion_comment = False)

marmegh:
>                             validatorMessage += " " + DynamicValidationNotEnabled;

test gap

---
In reply to: [1242511925](https://github.com/microsoft/sarif-pattern-matcher/pull/645#issuecomment-1242511925) [](http://example.com/codeflow?ancestors=1242511925,1242511925)

---
Refers to: Src/Sarif.PatternMatcher/SearchSkimmer.cs:377 in 0ba90a7. [](commit_id = 0ba90a73dfc00af3d4f2b60b0a6dc9d99f8de955, deletion_comment = False)

Src/Test.UnitTests.Sarif.PatternMatcher.Cli/AnalyzeCommandTests.cs
```diff
@@ -277,6 +280,98 @@ public void AnalyzeCommand_ShouldThrowExceptionForConflictingFileSizeOptions()
...
+                                                     validationScenario: testScenarioName,
+                                                     isDynamicAnalysis: testCase.dynamicValidationEnabled,
+                                                     ref sb);
+
+                testScenarioName = "StaticOnly";
+                logFile = RunAnalyzeCommandWithDynamicValidation(definitionsText,
+                                                                 testScenarioName,
+                                                                 testCase.dynamicValidationEnabled);
+
+                ValidateFailureLevelByValidationType(expectedFailureLevel: testCase.expectedStaticOnly,
```
marmegh, line 353: >testCase.expectedStaticOnly [](http://example.com/codeflow?start=75&length=27)

This was the test gap, we were validating against the previous test case instead of the current test case.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -28,7 +28,7 @@ public AnalyzeCommand(IFileSystem fileSystem = null)
...
+            var validators = new ValidatorsCache(fileSystem: fileSystem);
```
michaelcfanning, line 31: >fileSystem: fileSystem [](http://example.com/codeflow?start=49&length=22)

Please remove the redundancy in this, i.e., 

pass 
  fileSystem

instead of 
  fileSystem: fileSystem

Make sense? the latter pattern repeats information unnecessarily.

michaelcfanning (approved):
:shipit:

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
marmegh: This is necessary to disambiguate which nullable parameter we are using to instantiate the ValidatorsCache. Taking this change causes build errors. Chaining the constructors for ValidatorsCache to allow this change, causes build errors elsewhere. This is the cleanest solution.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -28,7 +28,8 @@ public AnalyzeCommand(IFileSystem fileSystem = null)
...
```
michaelcfanning, line 31: >fileSystem: fileSystem [](http://example.com/codeflow?start=49&length=22)

validatorBinaryPaths: null, fileSystem

michaelcfanning, line 31: >fileSystem: fileSystem [](http://example.com/codeflow?start=49&length=22)

IEnumerable<string> validatorBinaryPaths = null;
var validators = new ValidatorsCache(validatorBinaryPaths, fileSystem);

PR #647, opened 2022-09-02
Loosen probing logic to look for files directly alongside client exe.

michaelcfanning:
We now look for any definitions file, by name, in the binary execution directory.

@cfaucon

### DIFF
--- a/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
+++ b/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
@@ -33,9 +33,23 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 
             var skimmers = new HashSet<Skimmer<AnalyzeContext>>();
 
-            // TODO exception handling for bad search definitions files
-            foreach (string searchDefinitionsPath in searchDefinitionsPaths)
+            foreach (string inputSearchDefinitionsPath in searchDefinitionsPaths)
             {
+                string searchDefinitionsPath = inputSearchDefinitionsPath;
+
+                if (!fileSystem.FileExists(searchDefinitionsPath))
+                {
+                    string coreToolDirectory = typeof(AnalyzeContext).Assembly.Location;
+                    coreToolDirectory = Path.GetDirectoryName(coreToolDirectory);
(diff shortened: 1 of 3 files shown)


cfaucon (approved):

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -33,9 +33,23 @@ public AnalyzeCommand(IFileSystem fileSystem = null)
...
+                    string coreToolDirectory = typeof(AnalyzeContext).Assembly.Location;
+                    coreToolDirectory = Path.GetDirectoryName(coreToolDirectory);
+                    searchDefinitionsPath = Path.GetFileName(searchDefinitionsPath);
+                    searchDefinitionsPath = Path.Combine(coreToolDirectory, searchDefinitionsPath);
+
+                    if (!fileSystem.FileExists(searchDefinitionsPath))
+                    {
+                        throw new ArgumentException($"Could not locate specified definitions path: '{inputSearchDefinitionsPath}'");
+                    }
+                }
```
marmegh, line 51: This is an okay mitigation. However, we should really be checking that the file exists before any uses of `File.ReadAllText`. We should consider a follow up item to add this check directly to `fileSystem.FileReadAllText`. Alternatively, we would want to make sure we're confirming a file exists for all invocations of `fileSystem.FileReadAllText`. The first option seems like a cleaner and more maintainable option.

michaelcfanning: An excellent thought and something we should always consider: is some condition/invariant something that should be enforced in the callee or at every call site?

One thing to keep in mind here is that the file system object is a wrapper on top of System.IO, to assist with mocking this .NET namespace. This allows us to intercept .NET calls and replace their behavior. Generally, however, this replacement would mock an existing .NET behavior.

.NET today, by design, throws FileNotFoundException when calling File.ReadAllText, which seems like a good behavior. In our case here, we add more code to improve on the error reporting that .NET would provide otherwise.

We can talk more offline, this is a great conversation. Other things to criticize about this change include whether this code could/should have been factored into a helper. I also see other IO-related API in this method (like 'Path.GetDirectoryName') that makes me wonder whether our file system mock is complete/sufficiently utilized.

PR #650, opened 2022-09-09
No memory leaks resulting from file-specific fingerprint hash

michaelcfanning:
This change pushes the per-file fingerprint hash onto the per-file analyze context.

Previously, this data was maintained by each validator. This complicates managing this data (including proactively releasing it in an API use case scenario).

We accomplish this by providing a set to every static validation call (which comprises the per-file fingerprint hash). This change explicitly requires we manage on a per-file basis (previously caching was by unique file hash). It's an interesting design question whether this is correct.

As part of making this change, we uncovered an issue in the SQL rules where MySql, SQL and PostGres produce common validation fingerprints. The related tests are disabled pending investigation.

This change does *not* resolve the file region caching, which needs to be examined before building/shipping any packages.


Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs
```diff
@@ -9,7 +11,37 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
...
+        /// The type of this cache is peculiar. We use a concurrent dictionary
+        /// because .NET does not ship a concurrent HashSet class and we do
+        /// not have time currently to author one. The choice of 'byte' as a
+        /// value type for the dictionary is arbitrary; this data is never
+        /// consumed.
+        /// </param>
+        /// <returns> One or ValidationResults indicating zero or more valid findings.
+        /// </returns>
+        public IEnumerable<ValidationResult> IsValidStatic(IDictionary<string, FlexMatch> groups,
+                                                           ConcurrentDictionary<string, byte> perFileFingerprintCache)
```
michaelcfanning, line 44: >ConcurrentDictionary [](http://example.com/codeflow?start=59&length=20)

IDictionary, since call in theory could use a single thread?

Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs
```diff
@@ -20,16 +53,14 @@ public IEnumerable<ValidationResult> IsValidStatic(IDictionary<string, FlexMatch
                     continue;
                 }
 
-                string scanTarget = groups["scanTargetFullPath"].Value;
-                string key = $"{scanTarget}#{validationResult.Fingerprint}";
```
eddynaka, line 24: >{scanTarget}# [](http://example.com/codeflow?start=31&length=13)

Per description, this is a perfile fingerprint hash.
With that, I would expect to see this in the new changes as well.

Otherwise, we have a fingerprintCache only.

Src/Sarif.PatternMatcher.Sdk/DynamicValidatorBase.cs
```diff
@@ -16,6 +16,7 @@ public abstract class DynamicValidatorBase : StaticValidatorBase
             resultLevelKind = default;
 
             if (shouldUseDynamicCache &&
+                FingerprintToResultCache != null &&
```
eddynaka, line 19: >FingerprintToResultCache [](http://example.com/codeflow?start=16&length=24)

Is this change needed?

looking at the changes in the ValidatorBase, you still initialize the FingerprintToResultCache in the constructor.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
```diff
@@ -55,8 +59,22 @@ public AnalyzeContext()
...
         public long MaxMemoryInKilobytes { get; set; } = -1;
 
+        public FileRegionsCache FileRegionsCache { get; set; }
+
+        public IEnumerable<Skimmer<AnalyzeContext>> Skimmers { get; set; }
+
+        /// <summary>
+        /// Gets a hashset that stores observed fingerprints in the
+        /// current scan target. This data is used to prevent firing
+        /// multiple instances of he same logically unique apparent
```
eddynaka, line 69: >he [](http://example.com/codeflow?start=34&length=2)

the

Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs
```diff
@@ -1,15 +1,48 @@
...
 using Microsoft.RE2.Managed;
 
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public abstract class StaticValidatorBase : ValidatorBase
     {
-        public IEnumerable<ValidationResult> IsValidStatic(IDictionary<string, FlexMatch> groups)
+        /// <summary>
+        /// Examines the groups output by a positive regex match and
+        /// optionally performs additional validated to determine
```
eddynaka, line 17: >validated [](http://example.com/codeflow?start=43&length=9)

validation

Src/Sarif.PatternMatcher.Sdk/StaticValidatorBase.cs
```diff
@@ -1,15 +1,48 @@
...
+        /// observed previously. Our scanner will only detect and validate
+        /// a unique fingerprint once per file. Many regexes will produce
+        /// multiple matches that resolve to the same unique credential in
+        /// a file (one of the perils of multiline regex matching). It is
+        /// possible that this cache may drop the location of a second
+        /// actual match that happens to be duplicated in the file. In
+        /// practice, we will not worry about this scenario: it will be
+        /// sufficient that we flag one instance of the unique secret.
+        ///
+        /// The type of this cache is peculiar. We use a concurrent dictionary
```
eddynaka, line 36: >We use a concurrent dictionary [](http://example.com/codeflow?start=48&length=30)

Here we say we are using a ConcurrentDictionary with <string, byte> but, in code, we are using an ISet<string>.

Looking at the AnalyzeContext, we are using a HashSet<string>.

Can you update the text or switch back to concurrentDictionary?

michaelcfanning: 'DRAFT ONLY: don't review' :)

Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -46,6 +48,30 @@ protected EndToEndTests(ITestOutputHelper outputHelper) : base(outputHelper)
         {
             var inputFiles = parameter as List<string>;
             var results = new Dictionary<string, string>();
+
+            results = Debugger.IsAttached
```
michaelcfanning, line 52: >Debugger [](http://example.com/codeflow?start=22&length=8)

In practice, debugging the end-to-end tests when multithreaded is very difficult. The common scenario is to debug one or a few files and it is hard to step through a discrete repro w/many other threads traveling the same breakpoints. With this change, we will be single-threaded when debugging. See code comment for remaining thoughts.

Src/Sarif.PatternMatcher.Cli/StressCommand.cs
```diff
@@ -0,0 +1,111 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Runtime.Serialization;
+using System.Text;
+
+using Microsoft.CodeAnalysis.Sarif.Driver;
+using Microsoft.CodeAnalysis.Sarif.Multitool;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Cli
```
michaelcfanning, line 15: >Microsoft [](http://example.com/codeflow?start=10&length=9)

This code is entirely speculative at this point, please don't review.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
michaelcfanning: thanks, fixed.

cfaucon (approved):

PR #664, opened 2022-10-04
RE2.Managed & Strings.Interop Package -> Project References.

HulonJenkins:
## Changes
Swapped RE2.Managed from Package Reference to Project Reference.
Swapped Strings.Interop from Package Reference to Project Reference.

We had the projects but were using NuGet packages instead. Removed the NuGet packages and added reference to projects.


michaelcfanning (approved):
:shipit:

PR #665, opened 2022-10-05
Create cache for UTF8/16 index map when invoking patterns with groups.

michaelcfanning:
RE2 supports named groups using a particular regex syntax, e.g., (?P<myGroup>). Retrieving groups requires special invocation from our RE2 managed API, however. This layer first parses the input file in order to handle UTF8/UTF16 interoperability between .NET and RE2. This parsing occurred on every call for every pattern, greatly compromising performance.

This interim change allows for the scan tool to build that per-file UTF8/UTF16 index single time. We should next test and see how this performs relative to a second approach (which would entail eliding the groups altogether, calling the RE2 helpers with no special group processing and then (but only in the event of a match attempting to resolve the groups).

This latter optimization actually could be accomplished strictly within RE2 (rather than, say, falling back onto .NET)

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -835,7 +835,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
             // end of lines as well as the beginning or end of the search text.
             string lineRegex = $"(?m)^.*{firstRegex}.*";
 
-            if (Matches(lineRegex,
+            if (!Matches(lineRegex,
                         searchText,
                         out List<Dictionary<string, FlexMatch>> singleLineMatches,
                         context))
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -789,15 +789,10 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 
                 Debug.Assert(!contentsRegex.StartsWith("$"), $"Unexpanded regex variable: {contentsRegex}");
 
-                long maxMemoryInKB =
-                    context.MaxMemoryInKilobytes == -1
-                        ? context.MaxMemoryInKilobytes
-                        : 1024 * context.MaxMemoryInKilobytes;
-
                 if (!Matches(contentsRegex,
-                                                 searchText,
-                                                 out List<Dictionary<string, FlexMatch>> matches,
-                                                 context))
+                             searchText,
+                             out List<Dictionary<string, FlexMatch>> matches,
+                             context))
                 {
                     if (matchExpression.IntrafileRegexMetadata[i] == RegexMetadata.Optional)
                     {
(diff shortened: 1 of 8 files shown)


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -1,4 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
michaelcfanning, line 2: >Copyright [](http://example.com/codeflow?start=3&length=9)

Oops, what?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -789,7 +789,15 @@ private void RunMatchExpressionForIntrafileRegexes(AnalyzeContext context, Match
 
                 Debug.Assert(!contentsRegex.StartsWith("$"), $"Unexpanded regex variable: {contentsRegex}");
 
-                if (!_engine.Matches(contentsRegex, searchText, out List<Dictionary<string, FlexMatch>> matches))
+                long maxMemoryInKB =
```
michaelcfanning, line 792: >maxMemoryInKB [](http://example.com/codeflow?start=21&length=13)

Delete this.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -789,7 +789,15 @@ private void RunMatchExpressionForIntrafileRegexes(AnalyzeContext context, Match
...
+                    context.MaxMemoryInKilobytes == -1
+                        ? context.MaxMemoryInKilobytes
+                        : 1024 * context.MaxMemoryInKilobytes;
+
+                if (!Matches(contentsRegex,
+                                                 searchText,
```
michaelcfanning, line 798: >searchText [](http://example.com/codeflow?start=49&length=10)

fix alignment

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -1068,6 +1095,35 @@ private void RunMatchExpressionForSingleLineRegexes(AnalyzeContext context, Matc
...
+
+
+            long maxMemoryInKB =
+                context.MaxMemoryInKilobytes == -1
+                    ? context.MaxMemoryInKilobytes
+                    : 1024 * context.MaxMemoryInKilobytes;
+
+            if (re2regex != null)
+            {
+                return ((RE2Regex)_engine).Matches(contentsRegex,
```
michaelcfanning, line 1114: >_engine [](http://example.com/codeflow?start=34&length=7)

Call through 're2regex'

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -4,7 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
```
michaelcfanning, line 7: >IO [](http://example.com/codeflow?start=13&length=2)

clean these up.

PR #672, opened 2022-10-10
Performance fix to match Legacy speed

HulonJenkins:
## Changes
This PR was created off the same branch as https://github.com/microsoft/sarif-pattern-matcher/pull/665

Update caching so that minimum work is repeated to improve performance of the tool. 

Now adds all relevant information as a dictionary. This caching update is now 1:1 with Legacy in terms of performance with the same ruleset.

Started by just caching the int[], then added caching for the String8 and byte[] buffer as well after noticing that there was non-negligible time spent in the String8 conversion function.

From there, tests started failing because some files output raw and base64 decoded text to be scanned, and the cached String8/byte[] weren't being updated accordingly since it was different input text, but from the same file. So updated the caching to group input text (whether it was raw or base64, or anything else) to an String8 and byte[] tuple.

From there, noticed String8 to string comparison failures and lots of time spent in comparing new text coming in to see if it is already cached. Text coming in was basic string, key in dictionary was String8, so comparing the two took very long. Swapped to using basic string as dictionary key to speed up string comparison, and began storing String8, int[], and byte[] together as a tuple.

### DIFF
--- a/Src/RE2.Managed/Regex2.cs
+++ b/Src/RE2.Managed/Regex2.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 using System.Reflection;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -196,18 +197,14 @@ namespace Microsoft.RE2.Managed
                                           out List<Dictionary<string, FlexMatch>> matches,
                                           long maxMemoryInBytes)
         {
-            int[] indexMap = null;
-            String8 expression8 = String8.Empty;
-            byte[] buffer = null;
-            return Matches(pattern, text, out matches, ref indexMap, ref expression8, ref buffer, maxMemoryInBytes);
+            var textToIdMap = new Dictionary<String8, Tuple<byte[], int[]>>();
+            return Matches(pattern, text, out matches, ref textToIdMap, maxMemoryInBytes);
         }
 
         public static unsafe bool Matches(string pattern,
                                           string text,
                                           out List<Dictionary<string, FlexMatch>> matches,
-                                          ref int[] indexMap,
-                                          ref String8 expression8,
-                                          ref byte[] buffer,
+                                          ref Dictionary<String8, Tuple<byte[], int[]>> textToIdMap,
                                           long maxMemoryInBytes)
         {
             ParsedRegexCache cache = null;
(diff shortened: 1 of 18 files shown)


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -1,4 +1,5 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
```
HulonJenkins, line 2: oops

Src/Sarif.PatternMatcher.Cli/StressCommand.cs
```diff
@@ -4,13 +4,21 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Diagnostics;
```
HulonJenkins, line 7: >iagnost [](http://example.com/codeflow?start=14&length=7)

Remove extra using statements

Src/Test.UnitTests.RE2.Managed/Regex2Tests.cs
```diff
@@ -4,7 +4,9 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
```
HulonJenkins, line 7: >System.IO; [](http://example.com/codeflow?start=6&length=10)

usings again

Src/RE2.Managed/Regex2.cs
```diff
@@ -207,9 +221,33 @@ public static unsafe bool Matches(string pattern, string text, out List<Dictiona
...
                 byte[] buffer = null;
-                var expression8 = String8.Convert(text, ref buffer);
-                int[] indexMap = GetMapOfUtf8ToUtf16ByteIndices(buffer);
+                int[] indexMap = null;
+                bool saveToDictionary = false;
+                bool updateIndexMap = false;
+
+                foreach (KeyValuePair<string, Tuple<String8, byte[], int[]>> entry in textToIdMap)
+                {
+                    if (entry.Key.Length == text.Length && entry.Key == text)
```
HulonJenkins, line 237: >entry.Key.Length == text.Length [](http://example.com/codeflow?start=24&length=31)

Was hoping to short circuit and avoid a string comparison if not needed, but I'm not sure if this actually helps us at all. Thoughts?

Src/RE2.Managed/Regex2.cs
```diff
@@ -207,9 +220,30 @@ public static unsafe bool Matches(string pattern, string text, out List<Dictiona
...
+                foreach (KeyValuePair<string, Tuple<String8, byte[], int[]>> entry in textToRE2DataMap)
+                {
+                    if (entry.Key.Length == text.Length && entry.Key == text)
+                    {
+                        expression8 = entry.Value.Item1;
+                        buffer = entry.Value.Item2;
+                        indexMap = entry.Value.Item3;
+                        break;
+                    }
+                }
```
eddynaka, line 240: You could replace your foreach loop with a dict.TryGetValue.
If it exists, just populate expression8, buffer, and indexMap.
If it does not exist, then, you create expression8.

Just to be aware, I'm not saying that this will be faster than what you currently have.

michaelcfanning: Right. Actually, I suggest that you convert this code to use a HybridDictionary, which will use a list internally until the dictionary count exceeds 10 (which shouldn't happen in our utilization!).

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -962,13 +970,18 @@ private void RunMatchExpressionForSingleLineRegexes(AnalyzeContext context, Matc
                                                    ? Decode(binary64DecodedMatch.Value)
                                                    : context.FileContents;
 
+            long maxMemoryInKB =
```
michaelcfanning, line 973: >maxMemoryInKB [](http://example.com/codeflow?start=17&length=13)

you don't need this, the work is done in Matches, isn't it?

Src/RE2.Managed/Regex2.cs
michaelcfanning: Looks like maybe there isn't a generic hybrid dictionary?

next suggestion, convert this data structure to an actual list of tuples (and add the scan input string as a new tuple record member).

michaelcfanning (approved):

PR #673, opened 2022-10-12
PatternInvariantTest with 2 Json Files

HulonJenkins:
## Changes

Updated PatternInvariantTests to pull shared strings file from the field located in the json file. This ensures that the shared strings file that will be used is the one being checked. Before the shared strings file was passed in separately and could pass tests, while not actually being the file that the json would reference.
Also updated to loop through json files since there are now 2.

### DIFF
--- a/Src/Test.UnitTests.Sarif.PatternMatcher/PatternInvariantTests.cs
+++ b/Src/Test.UnitTests.Sarif.PatternMatcher/PatternInvariantTests.cs
@@ -71,7 +71,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
         {
             string definitionsFileContents = File.ReadAllText(definitionsFilePath);
             SearchDefinitions sdObject = JsonConvert.DeserializeObject<SearchDefinitions>(definitionsFileContents);
-            
+
             // Load shared strings from file in JSON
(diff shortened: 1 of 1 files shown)


Src/Test.UnitTests.Sarif.PatternMatcher/PatternInvariantTests.cs
```diff
@@ -102,44 +108,58 @@ public static void VerifyAllJsonRulesExist(string definitionsFilePath, string sh
...
         /// </summary>
-        public static void VerifyAllRuleFilenamesMatchDefinitions(string definitionsFilePath, string validatorsFolderName)
+        public static void VerifyAllRuleFilenamesMatchDefinitions(string definitionsFileDirectory)
         {
-            //Read the json file content to get the rule names
-            string content = File.ReadAllText(definitionsFilePath);
-            SearchDefinitions sdObject = JsonConvert.DeserializeObject<SearchDefinitions>(content);
+            // Get all json files from the folder passed with correct security division
+            string validatorsFolderName = "SecurePlaintextSecretsValidators";
+            string securityDivision = "SEC101";
```
michaelcfanning, line 115: >securityDivision [](http://example.com/codeflow?start=19&length=16)

securePlainSecretsRuleId

michaelcfanning (approved):
:shipit:

PR #682, opened 2022-11-01
Redact sensitive data in presence of `--redact-secrets`

michaelcfanning:
In the presence of `redact-secrets` we will elide plaintext passwords from region and context region snippets. We will not emit the plaintext secret and validation fingerprints.

This change also adjusts the truncation logic to prefer the rightmost 6 characters (excluding '=' chars suffixing base64-encoded things).


Src/Plugins/Tests.Security/TestData/ReviewPotentiallySensitiveData/ExpectedOutputs/SEC102_001.EmailAddress.sarif
```diff
@@ -82,9 +82,9 @@
           "fingerprints": {
             "SecretHashSha256/current": "a4965ae12f3568daff154d92dc41cddff82ae7b6674138279bbdac60217d7d1b",
             "AssetFingerprint/current": "{\"host\":\"example.com\",\"id\":\"email\"}",
+            "ValidationFingerprintHashSha256/current": "6064325b0dd768a320862c00c981779e3bccd9a3fa3ce9dc1473a9a41f188f53",
```
michaelcfanning, line 85: >ValidationFingerprintHashSha256 [](http://example.com/codeflow?start=13&length=31)

This is unfortunate, changes in code have resulted in reordered dictionary keys. We should consider using a sorted dictionary to preserve a deterministic order.

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_001.HttpAuthorizationRequestHeader.sarif
```diff
@@ -51,7 +51,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "SomeAu…",
+              "…111111",
```
michaelcfanning, line 54: >111111 [](http://example.com/codeflow?start=16&length=6)

An essential change in behavior, we now prefer the right-most six characters of a randomized token rather than the leftmost. This is because the leftmost portion tends to contain fixed data for many security models (such as databricks keys which are prefixed with dapi_).

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_005.SlackApiKey.sarif
```diff
@@ -54,7 +54,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "xoxb-8…",
```
michaelcfanning, line 57: >xoxb-8 [](http://example.com/codeflow?start=15&length=6)

Here's a great example of why we do this, observe there is only a single character of randomization in this token truncation.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -307,8 +307,10 @@ internal static void UpdateLevelKind(MatchExpression matchExpression)
                     DefaultTraces.None;
 
             context.Retry = options.Retry;
+            context.RedactSecrets = options.RedactSecrets;
             context.EnhancedReporting = options.EnhancedReporting;
             context.DynamicValidation = options.DynamicValidation;
+            context.DataToInsert = options.DataToInsert.ToFlags();
```
michaelcfanning, line 313: >DataToInsert [](http://example.com/codeflow?start=20&length=12)

Bug! We didn't previously pass this data on the context object. It was apparently unused previously but the recent region populating changes resulted it being consulted in some code paths. Found via unit testing.

Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -26,23 +26,11 @@ public class AnalyzeOptions : AnalyzeOptionsBase
...
         [Option(
             "deny-regex",
             HelpText = "A regular expression used to suppress scanning for any file or directory path that matches the regex.")]
         public string FileNameDenyRegex { get; internal set; }
 
-        [Option(
-            "file-size-in-kb",
-            HelpText = "The maximum file size (in kilobytes) that will be analyzed.",
-            Default = 1024)]
-        public int FileSizeInKilobytes { get; internal set; }
```
michaelcfanning, line 44: >FileSizeInKilobytes [](http://example.com/codeflow?start=19&length=19)

this arg obsoleted by max-file-size-in-kb, which we now pick up from the driver framework.

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -581,14 +579,23 @@ private static void MergeDictionary(IList<Dictionary<string, FlexMatch>> mergeFr
...
+            var fingerprints = new Dictionary<string, string>()
             {
                 { SecretHashSha256Current, fingerprint.GetSecretHash() },
                 { AssetFingerprintCurrent, fingerprint.GetAssetFingerprint() },
-                { SecretFingerprintCurrent, fingerprint.GetSecretFingerprint() },
-                { ValidationFingerprintCurrent, fingerprint.GetValidationFingerprint() },
                 { ValidationFingerprintHashSha256Current, fingerprint.GetValidationFingerprintHash() },
             };
+
+            // Add fingerprints that expose sensitive data in plaintext
```
michaelcfanning, line 590: >fingerprints [](http://example.com/codeflow?start=19&length=12)

This code path will redact the validation-relevant JSON that exposes credential passwords, etc.

cfaucon (approved):

PR #685, opened 2022-11-08
Disable DFA oom error as scary. 

michaelcfanning:
Will follow up with RE2 to consider lowering this output to warning.


michaelcfanning:
@HulonJenkins

michaelcfanning:
@eddynaka

eddynaka:
Can you update the releasehistory showing that we are updating RE2 version as well?

---
In reply to: [1307979293](https://github.com/microsoft/sarif-pattern-matcher/pull/685#issuecomment-1307979293) [](http://example.com/codeflow?ancestors=1307979293,1307979293)

eddynaka (approved):
:shipit:

PR #690, opened 2022-12-13
Properly version fingerprints hierarchical strings accrding to SARIF spec.

michaelcfanning:
We are not versioning our fingerprint hierarchical strings properly according to the SARIF spec.

This PR fixes that.

@rtaket @EasyRhinoMSFT @cfaucon 

https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317441


rtaket (changes requested):
Can we update the tool output to use the new fingerprint key?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -19,11 +19,11 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public class SearchSkimmer : Skimmer<AnalyzeContext>
     {
-        public const string SecretHashSha256Current = "SecretHashSha256/current";
-        public const string AssetFingerprintCurrent = "AssetFingerprint/current";
-        public const string SecretFingerprintCurrent = "SecretFingerprint/current";
-        public const string ValidationFingerprintCurrent = "ValidationFingerprint/current";
-        public const string ValidationFingerprintHashSha256Current = "ValidationFingerprintHashSha256/current";
+        public const string SecretHashSha256Current = "SecretHashSha256/v0";
+        public const string AssetFingerprintCurrent = "AssetFingerprint/v0";
```
rtaket, line 23: Did you want to move these non-fingerprints out of Fingerprints?

Src/Plugins/Tests.Security/TestData/UseSecureApi/ExpectedOutputs/SEC104.Memory.Allocation_malloca.sarif
```diff
@@ -84,11 +84,11 @@
             }
           ],
           "fingerprints": {
-            "SecretHashSha256/current": "a4965ae12f3568daff154d92dc41cddff82ae7b6674138279bbdac60217d7d1b",
-            "AssetFingerprint/current": "{\"id\":\"_malloca\"}",
-            "ValidationFingerprintHashSha256/current": "582290aa80b7a22d4aa98aa1227346b61831fcdc156e2a570bc6ad5b8763ccf2",
-            "SecretFingerprint/current": "{}",
-            "ValidationFingerprint/current": "{\"id\":\"_malloca\"}"
+            "SecretHashSha256/v0": "a4965ae12f3568daff154d92dc41cddff82ae7b6674138279bbdac60217d7d1b",
```
rtaket, line 87: I realize this wouldn't be part of your change, but it would be nice if the SDK somehow parsed and exposed the fingerprint versions. Maybe something like GetFingerprints(string key) that returned a list of versions and their values? It would prevent users from having to understand the versioning rules.

rtaket (approved):

HulonJenkins (approved):

PR #692, opened 2022-12-14
.Net 6.0

HulonJenkins:
## Changes

Updated build.app.props to include .net6.0. This outputs a folder in Sarif.PatternMatcher.Cli

* [ ] Added unit tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,11 +7,11 @@ You can contribute to this project from a Windows, macOS or Linux machine.
 On all platforms, the minimum requirements are:
 
 * Git client and command line tools.
-* .NET Core 3.1+
+* .NET Core 3.1+ and .NET 6.0
 
 ### Windows
 
-* Visual Studio 2019, any edition
+* Visual Studio 2019 or 2022(recommended), any edition
 * In the Visual Studio Installer, install the following:
   * Workloads
     * .NET desktop development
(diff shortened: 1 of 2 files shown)


michaelcfanning:
??  Who gave you this information? :)  Why would it be a good idea not update the packages?

---
In reply to: [1353702206](https://github.com/microsoft/sarif-pattern-matcher/pull/692#issuecomment-1353702206) [](http://example.com/codeflow?ancestors=1353702206,1353567475)

michaelcfanning:
Eddy had a real time wrestling with some recent package flattening work and I'm sure we advised him to not worry about this repo as a simplification. And considered zapping this package, as It is true that we don't see a lot of adoption of this tool via nuget.org. We'd never simply abandon that versioning effort, however, if we complete this conversation and stop investing, we should probably remove our package from NuGet. I probably noted as well that if we continue to ship in NuGet.org we should have a package that's compatible with dotnet tool installation. 

That's probably the right call: transition to .net tool install + the .net 6.0 build. Reach out to me via teams for more conversation on this if you like.

---
In reply to: [1353711153](https://github.com/microsoft/sarif-pattern-matcher/pull/692#issuecomment-1353711153) [](http://example.com/codeflow?ancestors=1353711153,1353567475)

michaelcfanning (approved):

PR #694, opened 2023-01-05
Move SPAM to unsigned build.

michaelcfanning:
@suvamM @EasyRhinoMSFT


EasyRhinoMSFT:
We should use the term "strong named" rather than "signed".

PR #695, opened 2023-01-05
Disable SEC101/001.HttpAuthorizationReuestHeader

HulonJenkins:
## Changes
Disabled SEC101/001.HttpAuthorizationReuestHeader due to timeouts being caused in PPE.
Also need to update SPMI.

Next steps are to investigate timeouts and reenable this rule with a fix.


Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_001.HttpAuthorizationRequestHeader.sarif
```diff
@@ -10,267 +10,14 @@
           "product": "Microsoft.TestHost",
           "fullName": "testhost 15.0.0.0",
           "version": "15.0.0.0",
-          "semanticVersion": "15.0.0",
-          "rules": [
-            {
-              "id": "SEC101/001",
```
suvamM, line 16: @HulonJenkins Why do we need to remove this? Isn't it enough to just remove the rule from the plug-in?

suvamM (changes requested):

Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_001.HttpAuthorizationRequestHeader.sarif
michaelcfanning: We leave this shell because we plan to re-enable the rule. Currently, the supporting rule test input is still being assessed (with no results because the rule is gone).

michaelcfanning (approved):
:shipit:

suvamM (approved):

PR #700, opened 2023-01-17
Actually disable SEC101/029

HulonJenkins:
## Changes
Disable SEC101/029.AlibabaCloudCredentials.
Update Debugging guide for clarity.
Update ReleaseHistory.md

* [ ] Added unit tests


suvamM (approved):

Src/ReleaseHistory.md
```diff
@@ -13,6 +13,7 @@
 - Rename `SEC101/050.IdentifiableNpmLegacyAuthorToken` to `SEC101/050.NpmIdentifiableAuthorToken` [#683](https://github.com/microsoft/sarif-pattern-matcher/pull/683)
 - Rename `SEC101/017.NpmAuthorToken` to `SEC101/017.NpmAuthorToken` [#683](https://github.com/microsoft/sarif-pattern-matcher/pull/683)
 - Rename `SEC101/006.GitHubPat` to `SEC101/006.GitHubLegacyPat` [#678](https://github.com/microsoft/sarif-pattern-matcher/pull/678)
+- Disable `SEC101/029.AlibabaCloudCredentials` [#700](https://github.com/microsoft/sarif-pattern-matcher/pull/700)
```
michaelcfanning, line 16: >AlibabaCloudCredentials [](http://example.com/codeflow?start=22&length=23)

Provide an explanation of why? Typically we would reference the exception that this rule is raising. Are you going to bring this check back? Think of it from the user standpoint.

Disable `SEC101/029.Alibaba` as it raises `XX` exceptions during validation producing messages such as `XXX`. These exceptions are caused by incompatibilities between the Alibaba supporting code and .net core 3.1 and 6.0. We should investigate restoring the rule when these dependencies are updated.

Src/ReleaseHistory.md
```diff
@@ -13,6 +13,9 @@
 - Rename `SEC101/050.IdentifiableNpmLegacyAuthorToken` to `SEC101/050.NpmIdentifiableAuthorToken` [#683](https://github.com/microsoft/sarif-pattern-matcher/pull/683)
 - Rename `SEC101/017.NpmAuthorToken` to `SEC101/017.NpmAuthorToken` [#683](https://github.com/microsoft/sarif-pattern-matcher/pull/683)
 - Rename `SEC101/006.GitHubPat` to `SEC101/006.GitHubLegacyPat` [#678](https://github.com/microsoft/sarif-pattern-matcher/pull/678)
+- Disable `SEC101/029.AlibabaCloudCredentials` which throws ScanErrors with message: 
+	"ValidationError:Could not load file or assembly 'AlibabaCloud.OpenApiClient, Version=0.1.4.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)"
```
michaelcfanning, line 17: >ValidationError [](http://example.com/codeflow?start=2&length=15)

What does this look like in the markdown view, did you check? I probably would have tried a block quote here, i.e., prefix the message with a >.

HulonJenkins: I already merged, but it just looks like text in double quotes. That is a good idea though.

PR #702, opened 2023-01-24
Initial matchexpression rule disable stub

HulonJenkins:
## Changes

Add functionality to disable a rule in json file. By adding RuleEnabledState = "Disabled", the rule will not run any analysis. Future efforst will have this same field determine failure level since those fields are included in the RuleEnabledState enum.

* [ ] Added unit tests

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchDefinition.cs
+++ b/Src/Sarif.PatternMatcher/SearchDefinition.cs
@@ -3,6 +3,8 @@
 
 using System.Collections.Generic;
 
+using Microsoft.CodeAnalysis.Sarif.Driver;
+
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 {
     public class SearchDefinition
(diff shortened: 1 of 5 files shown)


Src/Sarif.PatternMatcher/SearchDefinition.cs
```diff
@@ -19,6 +21,8 @@ public class SearchDefinition
 
         public ResultKind Kind { get; set; }
 
+        public RuleEnabledState RuleEnabledState { get; set; }
```
michaelcfanning, line 24: >RuleEnabledState [](http://example.com/codeflow?start=32&length=16)

This all looks right but where's the release note? What confidence do we have you've made a good change? Ad hoc testing from the command-line? Can we create a simple unit test somewhere appropriate?

HulonJenkins: I created some tests that verify functionality, and intend to use Yong's shell PR (https://dev.azure.com/mseng/1ES/_git/Spmi/pullrequest/698351) which has many rules disabled in this fashion to ensure that everything still works.

michaelcfanning: Roger that. Let's add a release note, clean up indentation in that one location and get this in. :)

michaelcfanning (approved):
:shipit:

PR #722, opened 2023-03-10
ExtensionMethods update

LingZhou-gh:
Minor (not significant) change
Add some extension functions to check whether a string contains digits and/or letters.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -18,6 +18,9 @@
 - UER => Eliminate unhandled exceptions in rule.
 - UEE => Eliminate unhandled exceptions in engine.
 
+## v4.2.0 Unreleased
+- BRK: Add/Rename functions in ExtensionMethods. [#722](https://github.com/microsoft/sarif-pattern-matcher/pull/722)
+
 ## v4.2.0 Released 03/07/2023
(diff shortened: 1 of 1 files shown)


Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs
```diff
@@ -39,7 +39,7 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
 
             // It is highly likely we do not have a key if we can't
             // find at least one letter and digit within the pattern.
-            if (!ContainsDigitAndChar(secret.Value))
+            if (!secret.Value.ToString().ContainsDigitAndLetter())
```
michaelcfanning, line 42: >ToString() [](http://example.com/codeflow?start=30&length=10)

are these ToString() calls actually necessary?

I think secret.Value is already a string.

michaelcfanning (approved):
:shipit:

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs
LingZhou-gh: I thought the same thing before. 
secret.Value is FlexString. 
secret.Value.ContainsDigitAndLetter() won't work here.

LingZhou-gh: image

PR #733, opened 2023-03-23
Update ctor init of ValidationResult.ValidationState (to Unknown).

michaelcfanning:
`Unknown` clearly seems like the appropriate default state for `ValidationState` in a newly allocated `ValidationResult` instance.

We not init to this in the ctor. Allows us to drop this explicit initialization in 50 rules.


yongyan-gh (approved):
:shipit:

PR #735, opened 2023-03-27
New validation helpers

michaelcfanning:
This changes adds two prospective methods to support cleaner validator logic. They may be helpful as permanent public helpers, in which case they should be added to the release notes (and format unit tests added).

See the doc comments for detailed information on the helpers.

### DIFF
--- a/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
@@ -1,10 +1,42 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class ExtensionMethods
     {
+        public static HttpResponseMessage ResponseHeadersRead(this HttpClient httpClient, HttpRequestMessage request)
+        {
+            return httpClient
+                .SendAsync(request, HttpCompletionOption.ResponseHeadersRead)
+                .GetAwaiter()
+                .GetResult();
+        }
+
+        public static void Merge(this HttpRequestHeaders httpRequestHeaders,
+                                      IEnumerable<KeyValuePair<string, string>> valuesToMerge)
+        {
+            if (valuesToMerge == null)
+            {
+                throw new ArgumentNullException(nameof(valuesToMerge));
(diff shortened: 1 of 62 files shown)


suvamM (changes requested):

Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
```diff
@@ -1,10 +1,42 @@
...
     {
+        public static HttpResponseMessage ResponseHeadersRead(this HttpClient httpClient, HttpRequestMessage request)
+        {
+            return httpClient
+                .SendAsync(request, HttpCompletionOption.ResponseHeadersRead)
+                .GetAwaiter()
+                .GetResult();
+        }
+
+        public static void Merge(this HttpRequestHeaders httpRequestHeaders,
```
LingZhou-gh, line 21: Maybe change the name to be Merge Headers to be more specific? As the request might have content, which need each rule class to add separately.

Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
```diff
@@ -1,10 +1,42 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
 namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class ExtensionMethods
     {
+        public static HttpResponseMessage ResponseHeadersRead(this HttpClient httpClient, HttpRequestMessage request)
```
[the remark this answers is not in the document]
michaelcfanning: Is is called `ReponseHeadersRead` because that's the operation specified in SendAsync and as a rule I tend to flow framework names around. `ReadResponseHeaders` is a clearer name, though, so I switched to that. Better?

Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
yongyan-gh: This is an extension method for `HttpRequestHeaders`, its caller we already know its a http header.
E.g. `request.Headers.Merge(...)`

michaelcfanning: But this code doesn't merge headers into the request headers instance, it merges any collection of key value pairs.

michaelcfanning: Utilization looks like this and I think it's pretty clear:

~~~
myHttpRequestHeaders.Merge(myDictionaryofValues);
~~~

suvamM (approved):

PR #736, opened 2023-03-29
Require asset information for unhandled exception reporting.

michaelcfanning:
- BRK: `ValidatorBase.ReturnUnhandledException` not requires an `asset` parameter. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
- BUG: Consistently report asset data (which may be the truncated secret) when reporting unhandled exceptions in dynamic validators. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)

@LingZhou-gh previously discovered an issue with our unhandled exception reporting, which would report an empty asset name when users failed to provide that argument to  `ValidatorBase.ReturnUnhandledException`
~~~
An unexpected exception was caught attempting to validate '': System.NullReferenceException: Object reference not set to an instance of an object.
~~~

The obvious immediate fix is to simply require this parameter, which will prompt us to review all rules to confirm whether we are passing an available asset id. Ideally, this data points to the compromised thing. Only as a fallback should we provide the truncated secret as an asset. We do this to provide some uniqueness to the reporting output, so that users know what credential provoked the unhandled exception.

There is an open item here, we need to find a find to convert certificate data that we process into some identifier. The easiest solution here would be to base64 encoded it and take the first 6 characters or so.

### DIFF
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -20,6 +20,7 @@
 
 ## v4.3.8 UNRELEASED
 - BRK: `ValidationResult` constructor now sets `ValidationState` to `Unknown`. [#733](https://github.com/microsoft/sarif-pattern-matcher/pull/733)
+- BUG: Consistently report asset data (which may be the truncated secret) when reporting unhandled exceptions in dynamic validators. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
 
 ## v4.3.7 Released 03/22/2023
 - DEP: Update SARIF SDK submodule from [53b0246 to 1ff3956](https://github.com/microsoft/sarif-sdk/compare/53b0246..1ff3956). [Full SARIF SDK release history](https://github.com/microsoft/sarif-sdk/blob/1ff3956/src/ReleaseHistory.md). Adds version control provenance.
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -20,7 +20,7 @@
 
 ## v4.3.8 UNRELEASED
 - BRK: `ValidationResult` constructor now sets `ValidationState` to `Unknown`. [#733](https://github.com/microsoft/sarif-pattern-matcher/pull/733)
-- BRK: `ValidatorBase.ReturnUnhandledException` not requires an `asset` parameter. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
+- BRK: `ValidatorBase.ReturnUnhandledException` now requires an `asset` parameter. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
 - BUG: Consistently report asset data (which may be the truncated secret) when reporting unhandled exceptions in dynamic validators. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
 
 ## v4.3.7 Released 03/22/2023
--- a/ReleaseHistory.md
+++ b/ReleaseHistory.md
@@ -20,6 +20,7 @@
 
 ## v4.3.8 UNRELEASED
 - BRK: `ValidationResult` constructor now sets `ValidationState` to `Unknown`. [#733](https://github.com/microsoft/sarif-pattern-matcher/pull/733)
+- BRK: `ValidatorBase.ReturnUnhandledException` not requires an `asset` parameter. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
 - BUG: Consistently report asset data (which may be the truncated secret) when reporting unhandled exceptions in dynamic validators. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
 
 ## v4.3.7 Released 03/22/2023
--- a/Src/Plugins/AzureDevOpsConfiguration/ServiceConnectionSecurityValidators/AZC102_190.DoNotGrantAllPipelinesAccessValidator.cs
+++ b/Src/Plugins/AzureDevOpsConfiguration/ServiceConnectionSecurityValidators/AZC102_190.DoNotGrantAllPipelinesAccessValidator.cs
@@ -19,7 +19,6 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.AzureDevOpsConfigu
 {
     public class DoNotGrantAllPipelinesAccessValidator : DynamicValidatorBase
     {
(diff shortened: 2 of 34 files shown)


ReleaseHistory.md
```diff
@@ -20,6 +20,8 @@
 
 ## v4.3.8 UNRELEASED
 - BRK: `ValidationResult` constructor now sets `ValidationState` to `Unknown`. [#733](https://github.com/microsoft/sarif-pattern-matcher/pull/733)
+- BRK: `ValidatorBase.ReturnUnhandledException` now requires an `asset` parameter. [#736](https://github.com/microsoft/sarif-pattern-matcher/pull/736)
```
michaelcfanning, line 23: >BRK [](http://example.com/codeflow?start=2&length=3)

We add a release note both for the tool consumer (who sees improved error reporting for this case) and for the API consumer (who sees a breaking change in the now required parameter). This is typical for the release notes: different audiences may observe the change in compile or runtime behavior.

Src/Plugins/AzureDevOpsConfiguration/ServiceConnectionSecurityValidators/AZC102_190.DoNotGrantAllPipelinesAccessValidator.cs
```diff
@@ -88,8 +89,7 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                 }
 
                 HttpClient httpClient = CreateOrRetrieveCachedHttpClient();
-
-                string apiUri = string.Format(PipelinePermissionAPI, organization, project, serviceConnectionId);
```
michaelcfanning, line 92: >string [](http://example.com/codeflow?start=16&length=6)

in general, we should hunt down string.format utilization in favor of interpolated strings. I did some of that here but tied off the work in order to close the PR (I left test code unimproved in some cases, for example).

Src/Plugins/AzureDevOpsConfiguration/ServiceConnectionSecurityValidators/AZC102_190.DoNotGrantAllPipelinesAccessValidator.cs
```diff
@@ -75,11 +74,13 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                                                                 IDictionary<string, string> options,
                                                                 ref ResultLevelKind resultLevelKind)
         {
+            string asset = string.Empty;
             try
             {
                 string organization = fingerprint.Host;
                 string project = fingerprint.Resource;
                 string serviceConnectionId = fingerprint.Id;
+                asset = $"https://{organization}/{project}/_apis/pipelines/pipelinePermissions/endpoint/{serviceConnectionId}";
```
michaelcfanning, line 83: >asset [](http://example.com/codeflow?start=16&length=5)

Here's a good example of what we want, an asset id that shows the compromised thing, in this case, an ADO service connection. If we don't have that, we'll use the truncated secret to provide uniqueness in the output.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_005.SlackApiKeyValidator.cs
```diff
@@ -44,13 +45,15 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                                                                 ref ResultLevelKind resultLevelKind)
         {
             const string uri = "https://slack.com/api/auth.test";
+            string secret = fingerprint.Secret;
```
michaelcfanning, line 48: >secret [](http://example.com/codeflow?start=19&length=6)

note how I continue to make all the code as consistent as possible. it's a big help when making mechanical changes such as this one.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs
```diff
@@ -76,11 +76,12 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                                                                 IDictionary<string, string> options,
                                                                 ref ResultLevelKind resultLevelKind)
         {
-            string pat = fingerprint.Secret;
```
michaelcfanning, line 79: >pat [](http://example.com/codeflow?start=19&length=3)

again, everywhere else we use 'secret' as a term, so we use it here as well. this assists in regex driven changes, etc.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs
```diff
@@ -93,6 +94,7 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                     name = $" ({name})";
                 }
 
+                asset = $"{id}{name}";
```
michaelcfanning, line 97: >asset [](http://example.com/codeflow?start=16&length=5)

another good change! we report '@michaelcfanning (Michael C. Fanning)' as the relevant tested thing now when this rule raises an exception.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_015.AkamaiCredentialsValidator.cs
```diff
@@ -60,9 +60,11 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
             string id = fingerprint.Id;
             string host = fingerprint.Host;
             string secret = fingerprint.Secret;
+            string asset = secret.Truncate(); // TBD
```
michaelcfanning, line 63: >asset [](http://example.com/codeflow?start=19&length=5)

there may be a better option for this security model but someone would need to get it into the debugger to take a look. that would be a good thing to do to learn about the subtleties in this change. any takers?

Src/Plugins/Security/Utilities/CertificateHelper.cs
```diff
@@ -73,6 +73,11 @@ internal static class CertificateHelper
                                                          ref Fingerprint fingerprint,
                                                          ref string message)
         {
+            // @michaelcfanning 3/28/23
```
michaelcfanning, line 76: >michaelcfanning [](http://example.com/codeflow?start=16&length=15)

an example of our TODO format, user name, date, link to GH issues, good description of open item. please absorb and model this standard. :)

PR #743, opened 2023-04-05
UI label + fix crash when no args passed to tool (should invoke help instead)

michaelcfanning:
We provide a user-friendly rule label for UX.

We now do not crash when passing no args to the tool.

@rtaket


rtaket:
lgtm

rtaket (approved):

PR #746, opened 2023-04-06
Set redact secrets via workaround until we get a driver framework fix.

michaelcfanning:
- BUG: We do not flow `--redact-secrets` argument properly to analysis. [#746](https://github.com/microsoft/sarif-pattern-matcher/pull/746)
- NEW: All `AnalyzeOptions` properties now settable. [#746](https://github.com/microsoft/sarif-pattern-matcher/pull/746)

### DIFF
--- a/Src/Sarif.PatternMatcher/AnalyzeOptions.cs
+++ b/Src/Sarif.PatternMatcher/AnalyzeOptions.cs
@@ -20,22 +20,22 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
             "dynamic-validation",
             HelpText = "Enable dynamic validations, if any (may compromise " +
                        "performance and/or result in calls across the network).")]
-        public bool? DynamicValidation { get; internal set; }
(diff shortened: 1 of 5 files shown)


Src/Sarif.PatternMatcher/AnalyzeOptions.cs
```diff
@@ -20,22 +20,22 @@ public class AnalyzeOptions : AnalyzeOptionsBase
             "dynamic-validation",
             HelpText = "Enable dynamic validations, if any (may compromise " +
                        "performance and/or result in calls across the network).")]
-        public bool? DynamicValidation { get; internal set; }
+        public bool? DynamicValidation { get; set; }
```
michaelcfanning, line 23: >get [](http://example.com/codeflow?start=41&length=3)

All options properties should be settable.

HulonJenkins (dismissed):
:shipit:

HulonJenkins (approved):
:shipit:

PR #749, opened 2023-04-12
Remove test dependency

michaelcfanning:
- BUG: Remove unnecessary `Test.Utilities.Sarif` reference from `Sarif.PatternMatcher` project. This test binary isn't built into our tool NuGet package and the erroneous reference causes missing file messages in some contexts.

@Jeremiah-Johnson @suvamM


HulonJenkins (approved):
:shipit:

PR #750, opened 2023-04-13
Redaction changes

michaelcfanning:
- DEP: Update SARIF SDK submodule from [1ff3956 to 36b4792](https://github.com/microsoft/sarif-sdk/compare/1ff3956..36b4792). [Full SARIF SDK release history](https://github.com/microsoft/sarif-sdk/blob/36b4792/src/ReleaseHistory.md). Adds version control provenance.
- NEW: All `AnalyzeContext` parameters can now be expressed as XML configuration. [#750](https://github.com/microsoft/sarif-pattern-matcher/pull/750)
- BUG: We now report the anonymized base64-encoded secret (rather than it decoded form) for rules like `SEC101/102.AdoPat` with this support. [#750](https://github.com/microsoft/sarif-pattern-matcher/pull/750)

### DIFF
--- a/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_102.AdoPat.sarif
+++ b/Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_102.AdoPat.sarif
@@ -64,7 +64,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "…wbbz5q",
+              "…Yno1cQ==",
               "an apparent ",
               "base64-encoded",
               "Azure DevOps personal access token",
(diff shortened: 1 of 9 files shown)


Src/Plugins/Tests.Security/TestData/SecurePlaintextSecrets/ExpectedOutputs/SEC101_102.AdoPat.sarif
```diff
@@ -64,7 +64,7 @@
           "message": {
             "id": "Default",
             "arguments": [
-              "…wbbz5q",
```
michaelcfanning, line 67: >wbbz5q [](http://example.com/codeflow?start=16&length=6)

Previously we showed a piece of the decoded token to users (when finding a base64 encoded ADO pat). This was confusing. We now reflect the data that is in the actual target file.

PR #755, opened 2023-04-28
Expand the Crc32 class by adding support for Castagnoli's polynomial

LingZhou-gh:
## Changes
The static validator of Yandex Cloud Iam Access Secret needs to use Castagnoli's polynomial in Crc32 checksum computation.
This change expands the Crc32 class by adding support for Castagnoli's polynomial without affecting the existing public methods and their reference. 
The default crc32 polynomial will be IEEE polynomial.

### DIFF
--- a/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
@@ -13,9 +13,17 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
         // This is the 'reversed representation' polynomial for
         // little-endian implementations, i.e., the bitwise
         // reflection of 0x04C11DB7;
-        private const uint Crc32Polynomial = 0xedb88320u;
+        // CRC-32-IEEE 802.3 polynomial
+        private const uint Crc32IEEEPolynomial = 0xedb88320u;
 
-        private static readonly uint[] Crc32Table = CreateCrcTable(Crc32Polynomial);
+        // Castagnoli's polynomial, used in iSCSI.
+        // Has better error detection characteristics than IEEE.
+        // https://dx.doi.org/10.1109/26.231911
+        private const uint Crc32CastagnoliPolynomial = 0x82f63b78;
+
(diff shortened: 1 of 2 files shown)


Src/Sarif.PatternMatcher.Sdk/Crc32.cs
```diff
@@ -29,30 +37,46 @@ public static uint Calculate(byte[] buffer)
...
+        public static uint Calculate(string text, uint[] crc32Table)
+        {
+            byte[] data = Encoding.ASCII.GetBytes(text);
+            return Calculate(data, crc32Table);
+        }
+
+        public static uint Calculate(byte[] buffer, uint[] crc32Table)
+        {
+            return Calculate(0, buffer, 0, buffer.Length, crc32Table);
+        }
```
yongyan-gh, line 53: instead of creating a new set of functions, consider adding an optional parameter to existing functions, less code duplication.

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
```diff
@@ -29,30 +37,46 @@ public static uint Calculate(byte[] buffer)
         }
 
         public static uint Calculate(uint checksum, byte[] buffer, int offset, int length)
+        {
+            return Calculate(checksum, buffer, offset, length, Crc32IEEETable);
+        }
+
+        public static uint Calculate(string text, uint[] crc32Table)
```
yongyan-gh, line 44: >uint[] crc32Table [](http://example.com/codeflow?start=50&length=17)

do we allow user to pass in custom crc32table data or any uint array?

if not, I would create an enum for all kinds of crc32 tables, use it as the optional parameter, and change the public crc32 table properties to private since its encapsulated internal detail.

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
LingZhou-gh: changed

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
LingZhou-gh: changed

HulonJenkins (approved):
:shipit:

PR #756, opened 2023-05-09
Sniff support

michaelcfanning:
This changes provides a new `SniffRegex` pattern to the `AnalyzeContext` object. When populated, this regex is applied to the contents of each scan target and any files that do not match against the regex will be skipped at analysis time.

This is a clean but limited change for testing. Certainly we should author notes or trace messages for skipped files in the future if we retain this functionality.


cfaucon (approved):

suvamM:
@michaelcfanning Wanted to check: why is the `SniffRegex` required?

PR #757, opened 2023-05-15
ETW events for scan analysis.

michaelcfanning:
- DEP: Update SARIF SDK submodule from [36b4792 to 140d730](https://github.com/microsoft/sarif-sdk/compare/36b4792..bc8cb57). [Full SARIF SDK release history](https://github.com/microsoft/sarif-sdk/blob/bc8cb57/ReleaseHistory.md). Adds version control provenance.
- NEW: Added ETW event tracing support.


Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
```diff
@@ -93,7 +93,7 @@ public void AnalyzeCommand_SniffRegex()
             sarifLog.Runs?[0]?.Results?.Count().Should().Be(0);
         }
 
-        [Fact]
+        [Fact(Skip = "Recent file regions cache change broke this test.")]
```
michaelcfanning, line 96: >Skip [](http://example.com/codeflow?start=14&length=4)

@shaopeng-gh , it looks as though your file region cache broke in-memory hashing (or perhaps we weren't hashing what we thought we were, i.e., the in-memory copy was incorrect?'.

can you investigate? this shows the importance of unit testing this area, it is complex and fragile.

when we fix this, we should improve the tests to not only verify that hashes exist but also verify the hash values themselves.

ReleaseHistory.md
```diff
@@ -18,6 +18,10 @@
 - UER => Eliminate unhandled exceptions in rule.
 - UEE => Eliminate unhandled exceptions in engine.
 
+## v4.5.0 UNRELEASED
```
michaelcfanning, line 21: >UNRELEASED [](http://example.com/codeflow?start=10&length=10)

This is not a draft PR but it is an intermediate, exploratory sort of check-in. i.e., this change attempts to use the new SARIF SDK event collection mechanism but also demonstrates how to add a custom event (in our case, the 'sniff' functionality). I'd like to review/refine this a bit, and then we can discuss a general test methodology. Ideally, this methodology would exist in the SARIF SDK however.

Src/Test.UnitTests.Sarif.PatternMatcher/AnalyzeCommandTests.cs
shaopeng-gh: will take a look at this one

shaopeng-gh: @michaelcfanning I checked, first this test is indeed broken by my fix.
The reason is, before my fix, we were re-using the **text** cache data to do the hashing, which was wrong. In a real scenario not these test codes, **text** cache data are read from the file, which could be binary, and text could not represent all bytes in a binary file, so to make sure we get the correct hash, we can not re-use the **text** cache to do the hashing, the hashing now is always calculated from directly hashing from the file on disk.

The testing code here is trying to feed a string and it will be in the cache, but the test case does not actually alter the file in disk. 
e.g. "Example.txt" it does not even exists. So the code will not get any hash.

Will need direction how do we want to fix this,

What my fix broken, is it a convenient way of testing without the need of actual file in the disk or mocking the file system? Or it is a real user use case.

PR #759, opened 2023-05-19
Exposing MatchExpressions in SearchSkimmer

suvamM:
## Changes

Various recent activities have the following pattern: we consume the JSON plug-in, and parse the regular expressions. The typical way to do this is to deserialize the JSON into a `SearchDefinitions` object, and then access the `MatchExpressions` list. However, this logic already exists in the `CreateSkimmersFromDefinitions` method, which creates a set of `Skimmer` objects. The problem: the match expressions were private.

This PR exposes the `MatchExpression` set as `public` in the `SearchSkimmer` class. 

### Sample usage:

```
ISet<Skimmer<AnalyzeContext>> skimmers = CreateSkimmersFromDefinitions(pluginsPath);
foreach (Skimmer<AnalyzeContext> skimmer in skimmers)
{
  var ss = (SearchSkimmer)skimmer;
  var regex = ss.MatchExpression.ContentsRegex;
}
```

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -27,6 +27,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
 
         public const string DynamicValidationNotEnabled = "No validation occurred as it was not enabled. Pass '--dynamic-validation' on the command-line to validate this match";
 
+        public readonly IList<MatchExpression> MatchExpressions;
+
         private const string DefaultHelpUri = "https://github.com/microsoft/sarif-pattern-matcher";
         private const string Base64DecodingFormatString = "\\b(?i)[0-9a-z\\/+]{0}";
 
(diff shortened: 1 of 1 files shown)


yongyan-gh (approved):
:shipit:

LingZhou-gh:
The match expression was changed from Private to Public in the PR, then where and how this public property will be actually used for? That is what is the need to make it public?

suvamM:
> The match expression was changed from Private to Public in the PR, then where and how this public property will be actually used for? That is what is the need to make it public?

@LingZhou-gh I tried to describe this in the PR description: we have several scenarios where we are parsing the regular expressions related to the rules. Concrete examples include the synthetic dataset generation, and also sniff regex generation. Without making the field `public`, we would have to duplicate the parsing logic.

suvamM:
> Should we add a release record for it?

@yongyan-gh I thought about it, but I don't really think this change is so large as to warrant an entry in the Release Notes.

HulonJenkins (approved):
:shipit:

PR #760, opened 2023-05-19
Code cleanup - eliminate build warnings

schlaman-ms:
The only purpose of this PR is to eliminate build warnings.
There are no new features or changes to logic or fixes.

## Changes
For two warnings that could not be eliminated, $(MSBuildWarningsAsMessages) was used.
For the Newtonsoft version conflicts, MSB3277.
For the copy files errors, MSB3030.
The copy file error occurs even though the files eventually do copy.

Here is an example of the MSBuildWarningsAsMessages
  <PropertyGroup>
 <MSBuildWarningsAsMessages>
     $(MSBuildWarningsAsMessages);MSB3277;MSB3030
 </MSBuildWarningsAsMessages>
  </PropertyGroup>

The rest of the fixes are like xml comments missing, and correcting order of static vs public, fields vs properties.

* [ ] Added unit tests

### DIFF
--- a/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
+++ b/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.AzureDevOpsConfiguration</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MSBuildWarningsAsMessages>
+	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+    </MSBuildWarningsAsMessages>
+  </PropertyGroup>
+	
   <ItemGroup>
     <None Remove="AdoPat.txt" />
     <None Remove="AZC102.ServiceConnectionSecurity.json" />
--- a/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
+++ b/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
@@ -13,8 +13,17 @@
 
   <PropertyGroup>
     <MSBuildWarningsAsMessages>
-        $(MSBuildWarningsAsMessages);MSB3277;MSB3030
-    </MSBuildWarningsAsMessages>
+		// MSB3277 - Found conflicts with different versions of the same
+		//  dependent assembly that could not be resolved. Specifically
+		//  for Newtonsoft where we need older versions for compatibility
+		// MSB3030 - This error occurs when a file that was supposed to be
+		//  copied isn't found at the expected location.
+		//  There are copy commands that have a continueOnError=true because
+		//  there is a time in the copy process where a copy fails before it
+		//  succeeds for those commands with the continueOnError. Those copy
+		//  commands have been set to continueOnError for almost two years.
+		$(MSBuildWarningsAsMessages);MSB3277;MSB3030
+	</MSBuildWarningsAsMessages>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
+++ b/Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
@@ -13,10 +13,10 @@
 
   <PropertyGroup>
     <MSBuildWarningsAsMessages>
-	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+        $(MSBuildWarningsAsMessages);MSB3277;MSB3030
     </MSBuildWarningsAsMessages>
   </PropertyGroup>
-	
+
   <ItemGroup>
     <None Remove="AdoPat.txt" />
     <None Remove="AZC102.ServiceConnectionSecurity.json" />
--- a/Src/Plugins/SalModernization/SalModernization.csproj
+++ b/Src/Plugins/SalModernization/SalModernization.csproj
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.SalModernization</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+	<MSBuildWarningsAsMessages>
+	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+	</MSBuildWarningsAsMessages>
+  </PropertyGroup>
+	
   <ItemGroup>
     <Content Include="build\Sarif.PatternMatcher.SalModernization.targets">
       <PackagePath>build\</PackagePath>
--- a/Src/Plugins/SalModernization/SalModernization.csproj
+++ b/Src/Plugins/SalModernization/SalModernization.csproj
@@ -13,8 +13,17 @@
 
   <PropertyGroup>
     <MSBuildWarningsAsMessages>
-        $(MSBuildWarningsAsMessages);MSB3277;MSB3030
-    </MSBuildWarningsAsMessages>
+		// MSB3277 - Found conflicts with different versions of the same
+		//  dependent assembly that could not be resolved. Specifically
+		//  for Newtonsoft where we need older versions for compatibility
+		// MSB3030 - This error occurs when a file that was supposed to be
+		//  copied isn't found at the expected location.
+		//  There are copy commands that have a continueOnError=true because
+		//  there is a time in the copy process where a copy fails before it
+		//  succeeds for those commands with the continueOnError. Those copy
+		//  commands have been set to continueOnError for almost two years.
+		$(MSBuildWarningsAsMessages);MSB3277;MSB3030
+	</MSBuildWarningsAsMessages>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/Plugins/SalModernization/SalModernization.csproj
+++ b/Src/Plugins/SalModernization/SalModernization.csproj
@@ -12,11 +12,11 @@
   </PropertyGroup>
 
   <PropertyGroup>
-	<MSBuildWarningsAsMessages>
-	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
-	</MSBuildWarningsAsMessages>
+    <MSBuildWarningsAsMessages>
+        $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+    </MSBuildWarningsAsMessages>
   </PropertyGroup>
-	
+
   <ItemGroup>
     <Content Include="build\Sarif.PatternMatcher.SalModernization.targets">
       <PackagePath>build\</PackagePath>
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -11,7 +11,13 @@
     <PackageId>Sarif.PatternMatcher.Security</PackageId>
   </PropertyGroup>
 
-  <ItemGroup>
+  <PropertyGroup>
+	<MSBuildWarningsAsMessages>
+	    $(MSBuildWarningsAsMessages);MSB3030
+    </MSBuildWarningsAsMessages>
+  </PropertyGroup>
+
+	<ItemGroup>
     <Content Include="*.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -12,12 +12,12 @@
   </PropertyGroup>
 
   <PropertyGroup>
-	<MSBuildWarningsAsMessages>
-	    $(MSBuildWarningsAsMessages);MSB3030
+    <MSBuildWarningsAsMessages>
+        $(MSBuildWarningsAsMessages);MSB3030
     </MSBuildWarningsAsMessages>
   </PropertyGroup>
 
-	<ItemGroup>
+  <ItemGroup>
     <Content Include="*.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
--- a/Src/Plugins/Security/Security.csproj
+++ b/Src/Plugins/Security/Security.csproj
@@ -13,8 +13,14 @@
 
   <PropertyGroup>
     <MSBuildWarningsAsMessages>
-        $(MSBuildWarningsAsMessages);MSB3030
-    </MSBuildWarningsAsMessages>
+		// MSB3030 - This error occurs when a file that was supposed to be
+		//  copied isn't found at the expected location.
+		//  There are copy commands that have a continueOnError=true because
+		//  there is a time in the copy process where a copy fails before it
+		//  succeeds for those commands with the continueOnError. Those copy
+		//  commands have been set to continueOnError for almost two years.
+		$(MSBuildWarningsAsMessages);MSB3030
+	</MSBuildWarningsAsMessages>
   </PropertyGroup>
 
   <ItemGroup>
--- a/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
+++ b/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
@@ -39,6 +39,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         /// Sign the HTTP request using Alibaba Cloud access key and secret so it
         /// can be used to access Alibaba Cloud service REST API.
         /// </summary>
+        /// <param name="request">httpRequest.</param>
+        /// <param name="accessKey">Alibbaba Cloud access key.</param>
+        /// <param name="secret">Alibaba Cloud secret.</param>
         public void SignRequest(HttpRequestMessage request, string accessKey, string secret)
         {
             const string format = "JSON";
--- a/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
+++ b/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
@@ -40,7 +40,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         /// can be used to access Alibaba Cloud service REST API.
         /// </summary>
         /// <param name="request">Unsigned httpRequest passed in that will be signed for Alibaba Cloud service REST API.</param>
-        /// <param name="accessKey">Alibbaba Cloud access key.</param>
+        /// <param name="accessKey">Alibaba Cloud access key.</param>
         /// <param name="secret">Alibaba Cloud secret.</param>
         public void SignRequest(HttpRequestMessage request, string accessKey, string secret)
         {
--- a/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
+++ b/Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
@@ -39,7 +39,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         /// Sign the HTTP request using Alibaba Cloud access key and secret so it
         /// can be used to access Alibaba Cloud service REST API.
         /// </summary>
-        /// <param name="request">httpRequest.</param>
+        /// <param name="request">Unsigned httpRequest passed in that will be signed for Alibaba Cloud service REST API.</param>
         /// <param name="accessKey">Alibbaba Cloud access key.</param>
         /// <param name="secret">Alibaba Cloud secret.</param>
         public void SignRequest(HttpRequestMessage request, string accessKey, string secret)
--- a/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
@@ -7,9 +7,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class Crc32
     {
-        public static readonly uint[] Crc32IEEETable = CreateCrcTable(Crc32IEEEPolynomial);
-
         public static readonly uint[] Crc32CastagnoliTable = CreateCrcTable(Crc32CastagnoliPolynomial);
+        private static readonly uint[] Crc32IEEETable = CreateCrcTable(Crc32IEEEPolynomial);
 
         // https://crc32.online/
         // https://github.com/force-net/Crc32.NET
--- a/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/Crc32.cs
@@ -7,6 +7,10 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class Crc32
     {
+        public static readonly uint[] Crc32IEEETable = CreateCrcTable(Crc32IEEEPolynomial);
+
+        public static readonly uint[] Crc32CastagnoliTable = CreateCrcTable(Crc32CastagnoliPolynomial);
+
         // https://crc32.online/
         // https://github.com/force-net/Crc32.NET
         // https://en.wikipedia.org/wiki/Cyclic_redundancy_check
@@ -20,10 +24,6 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
         // Has better error detection characteristics than IEEE.
         // https://dx.doi.org/10.1109/26.231911
         private const uint Crc32CastagnoliPolynomial = 0x82f63b78;
-
-        public static readonly uint[] Crc32IEEETable = CreateCrcTable(Crc32IEEEPolynomial);
-
-        public static readonly uint[] Crc32CastagnoliTable = CreateCrcTable(Crc32CastagnoliPolynomial);
 
         public static uint Calculate(string text, uint[] crc32Table = null)
         {
--- a/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
@@ -10,7 +10,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class ExtensionMethods
     {
-        public static char RedactionChar => '?';
+        public const char RedactionChar = '?';
 
         /// <summary>
         /// Use an <cref>HttpClient</cref> instance to retrieve request response headers only.
--- a/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
+++ b/Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
@@ -10,6 +10,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class ExtensionMethods
     {
+        public static char RedactionChar => '?';
+
         /// <summary>
         /// Use an <cref>HttpClient</cref> instance to retrieve request response headers only.
         /// </summary>
@@ -83,8 +85,6 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
                    truncatedText.Substring(truncatedText.Length - lengthExclusiveOfEllipsis) +
                    (charsElided ? suffix : string.Empty);
         }
-
-        public static char RedactionChar => '?';
 
         public static string Anonymize(this string text, int lengthExclusiveOfEllipsis = 6)
         {
--- a/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
+++ b/Src/Sarif.PatternMatcher/AnalyzeCommand.cs
@@ -80,6 +80,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
                     // TBD add version details. Breaks test baselines currently.
                     // semanticVersion = fvi?.ProductVersion;
                     // version = fvi?.FileVersion;
+                    // ADO Work item: https://dev.azure.com/mseng/1ES/_workitems/edit/2066916
                 }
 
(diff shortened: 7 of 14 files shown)


Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
```diff
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.AzureDevOpsConfiguration</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MSBuildWarningsAsMessages>
+	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
```
shaopeng-gh, line 16: >MSBuildWarningsAsMessages [](http://example.com/codeflow?start=7&length=25)

Just a question if this is what we want, or fix the warning.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -350,36 +373,14 @@ internal static void UpdateLevelKind(MatchExpression matchExpression)
             return result;
         }
 
-        public override AnalyzeContext InitializeGlobalContextFromOptions(AnalyzeOptions options, ref AnalyzeContext context)
```
shaopeng-gh, line 353: >InitializeGlobalContextFromOptions [](http://example.com/codeflow?start=39&length=34)

It looks like much change here, compare to other small; changes, could you explain a bit the change here

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
yongyan-gh, line 16: >	     [](http://example.com/codeflow?start=0&length=5)

mixing of tab and space characters

yongyan-gh: same to all other csproj files

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
```diff
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.AzureDevOpsConfiguration</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MSBuildWarningsAsMessages>
```
suvamM, line 15: @schlaman-ms Can you show the before and after due to this change? The build will now result in these messages? Or will they be suppressed entirely?

Src/Plugins/SalModernization/SalModernization.csproj
```diff
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.SalModernization</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+	<MSBuildWarningsAsMessages>
+	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+	</MSBuildWarningsAsMessages>
+  </PropertyGroup>
```
suvamM, line 18: @schlaman-ms There seems to be code duplication. Can we move this to build.common.props perhaps?

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
```diff
@@ -39,6 +39,9 @@ public AlibabaEcsRequestSigner(DateTime? dateTime = null, string signatureNonce
         /// Sign the HTTP request using Alibaba Cloud access key and secret so it
         /// can be used to access Alibaba Cloud service REST API.
         /// </summary>
+        /// <param name="request">httpRequest.</param>
```
suvamM, line 42: >httpRequest. [](http://example.com/codeflow?start=34&length=12)

@schlaman-ms This isn't a great XML  comment. Can you add a bit more description?

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
```diff
@@ -29,11 +29,11 @@ internal class SpamTestRule : Skimmer<AnalyzeContext>, IOptionsProvider
         internal static Random s_random = new Random(s_seed);
 
         [ThreadStatic]
-        internal static TestRuleBehaviors s_testRuleBehaviors;
+        private static TestRuleBehaviors s_testRuleBehaviors;
 
         public SpamTestRule()
         {
-            if (s_testRuleBehaviors.HasFlag(TestRuleBehaviors.RaiseExceptionInvokingConstructor))
+            if (S_testRuleBehaviors.HasFlag(TestRuleBehaviors.RaiseExceptionInvokingConstructor))
```
suvamM, line 36: >S_testRuleBehaviors [](http://example.com/codeflow?start=16&length=19)

@schlaman-ms I didn't fully follow this change. The variable name is still s_testRuleBehaviors right? Why capitalize the S?

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
schlaman-ms: There should be no functionality changes AT ALL. This is only to make the compiler happy to have things in the order it wants.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
schlaman-ms: No messages will show in the build at all with these suppressions.
Before you had many many lines of warning from the Newtonsoft conflict. Then there were the 4 copy warning lines.
Now there are zero lines generated as a result of these suppressions.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
schlaman-ms: It is not possible to fix either of the sources of the warnings, therefore I need to suppress them.

schlaman-ms: I can't update newtonsoft to eliminate the conflict warning and I can't get rid of the copy commands that sometimes don't copy and give warning but eventually copy.

suvamM: @schlaman-ms Do we have an understanding on why this transient fault is occurring for making copies? I agree with @shaopeng-gh that we should try to fix it, rather than suppress it. Is this a known issue?

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
suvamM: Not sure if this actually resolves the underlying problem.

Src/Plugins/SalModernization/SalModernization.csproj
schlaman-ms: Well I intentionally only put the MSBuildWarningsAsMessages entries in these projects as the suppressions only occur in the project where the entry is.

If I put this in build.common.props then I would suppress those errors globally and that wouldn't be good.
There is already a certain amount of risk by suppressing that you silence something that shouldn't be silenced. If in a given project, I only need to silence one warning then that project only gets the one MSBxxxx entry.

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
schlaman-ms: Yes, there were warnings generated because there were no XML parameter entries in the XML Comment.
I'll look at bit more at the code to come up with a better comment. Easier if it was my own code

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
schlaman-ms: Yes, sorry, this was a really messy one. properties not before fields. Constructor not before fields. public not before static.
I pretty much had to completely rearrange to satisfy all the different conditions.

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
schlaman-ms: Well the issue here is to eliminate the warning I took the automatic visual studio change.
It does look like all that changed is internal to private and then s_test... to S_test...
But s_test... still exists and is the backing field and the new property S_test... is added.
That's why it looks hard to follow and odd why the case change s_ --> S_

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
schlaman-ms: Transient isn't the correct word here really.
It will always have the failure to copy error, but by the end of the build, in each case for the 4 different warnings, each of those 4 files ends up copied.

The copy parameter for ONLY those 4 has the flag set to continue on error = true. ALL the other copies are continue on error=false.

It looks like those copy commands and continue on error= true settings have not changed for about 2 years. Somebody recognized it needed to be able to continue and that those would have an error but still needed to be there.

Can we have that be a separate task/PR  to investigate about why those copies are like that? I'm guessing that if the continue on error was set to true in the past there was a decent reason for it but who knows.

schlaman-ms: Nice catch. I didn't have display of spaces/tabs turned on and didn't see that what I pasted from an external source had the tabs.
My setup doesn't add tabs just the paste from outside.

Tabs removed now.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
```diff
@@ -11,6 +11,12 @@
...
+	    $(MSBuildWarningsAsMessages);MSB3277;MSB3030
+    </MSBuildWarningsAsMessages>
+  </PropertyGroup>
+
```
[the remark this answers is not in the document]
schlaman-ms: Tabs removed

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
schlaman-ms: Michael wanted me to suppress the Newtonsoft conflict warnings.

As for the 4 files copying issue (that looks like it has been there for a couple of years). All this does is suppress the message that on at least one of the copies it fails (which is why the continue on error=true is there and only on those 4 copies).
This just suppresses the warning but doesn't affect the proper copying of the files.

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
schlaman-ms: Ok now there is better XML comment

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
shaopeng-gh: properties not before fields. Constructor not before fields. public not before static. --- got it.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
michaelcfanning: When you disable warnings in project files, how about adding an XML comment describing what's being suppressed?

Src/Plugins/Security/Security.csproj
```diff
@@ -11,6 +11,12 @@
     <PackageId>Sarif.PatternMatcher.Security</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MSBuildWarningsAsMessages>
+        $(MSBuildWarningsAsMessages);MSB3030
```
michaelcfanning, line 16: >MSB3030 [](http://example.com/codeflow?start=37&length=7)

Add comment here and elsewhere documenting what this warning is.

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
```diff
@@ -39,6 +39,9 @@ public AlibabaEcsRequestSigner(DateTime? dateTime = null, string signatureNonce
         /// Sign the HTTP request using Alibaba Cloud access key and secret so it
         /// can be used to access Alibaba Cloud service REST API.
         /// </summary>
+        /// <param name="request">Unsigned httpRequest passed in that will be signed for Alibaba Cloud service REST API.</param>
```
michaelcfanning, line 42: >httpRequest [](http://example.com/codeflow?start=43&length=11)

what is this term? it's not a variable name, it's not a type, it's not a valid english term. ? :)

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
```diff
@@ -39,6 +39,9 @@ public AlibabaEcsRequestSigner(DateTime? dateTime = null, string signatureNonce
...
+        /// <param name="accessKey">Alibbaba Cloud access key.</param>
```
michaelcfanning, line 43: >Alibbaba [](http://example.com/codeflow?start=36&length=8)

this is misspelled.

how much of an investment are you making in improving the doc comments? If we aren't going to put them in a sufficient level of polish, we shouldn't bother.

Also, weren't you going to broadcast the engineering standards to the group before committing to many fixes? i.e., was there a discussion with the team about our position on this code standard? If we don't like it, we can just disable the check. Please do schedule this conversation, probably invite the whole team and you can review individual checks with us.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -76,9 +76,10 @@ public AnalyzeCommand(IFileSystem fileSystem = null)
                     FileVersionInfo fvi = fileSystem.FileVersionInfoGetVersionInfo(Path.Combine(directory, definitions.ValidatorsAssemblyName));
 
                     name = $"{fvi?.CompanyName}/{fvi?.FileDescription}/{name}";
+
                     // TBD add version details. Breaks test baselines currently.
-                    //semanticVersion = fvi?.ProductVersion;
-                    //version = fvi?.FileVersion;
+                    // semanticVersion = fvi?.ProductVersion;
```
michaelcfanning, line 81: >semanticVersion [](http://example.com/codeflow?start=23&length=15)

So here we've made a code clean-up while ignoring  something important, a dev (me! eep) has left a work item undone. Far more important than injecting spaces is that devs look at this and either update the code comment or create a tracking issue for the work.

Welcome to code clean-up. :) This will rub your face in all our technical debt. We perform code clean-up not merely to make symptoms (warnings) for away, we do it to rub our faces in our debt and potentially drive clean-up.

Please file an issue for this open item, which is actually pretty important TODO.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
```diff
@@ -13,6 +13,16 @@ public class AnalyzeContext : AnalyzeContextBase
     {
         public static long FilesFilteredBySniffRegex;
 
+        /// <summary>
+        /// Gets or sets a dictionary linking file text with
+        /// A String8 that is used to in RE2 searching.
```
michaelcfanning, line 18: this isn't grammatically correct.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
```diff
@@ -13,6 +13,16 @@ public class AnalyzeContext : AnalyzeContextBase
...
+        /// An array of bytes that comprise a buffer used in String8 conversion
```
michaelcfanning, line 19: >array [](http://example.com/codeflow?start=15&length=5)

I think the intent was to encode some of this text as XML params args, but that wasn't done.

michaelcfanning: I now understand that you simply moved this code. If you touched it, you bought it. :) The comment is bad.

Also, please do review this code standard with the team, i.e., property placement. Are we really going to enforce that standard? Why exactly? Someone on the team can advocate for it, I would like to hear the rationale.

Notice in this file as well that you moved a property with a code comment, but we're ignoring every other public thing that has no code comments. No good.

Now, balancing my remarks, if we get to a clean condition, we can enable warnings as errors and help prevent regression. So that is a fine value that might argue for not getting lost in too many 'you touched it' morasses.

Src/Sarif.PatternMatcher/JsonLogicalLocationProcessor.cs
```diff
@@ -27,6 +67,9 @@ public class JsonLogicalLocationProcessor
         ///     - For arrays, one might report the whole array, but this gets "closer" to the match.
         ///     - For objects, spanning properties will get "closer" to the match.
         /// </remarks>
+        /// <param name="results">collection of JSON results to process.</param>
```
michaelcfanning, line 70: >collection [](http://example.com/codeflow?start=34&length=10)

I don't think that most doc comments of this kind are sentence fragments. Please review actual .NET docs comments to get a sense of what the standard is.

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
```diff
@@ -7,6 +7,10 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class Crc32
     {
+        public static readonly uint[] Crc32IEEETable = CreateCrcTable(Crc32IEEEPolynomial);
```
michaelcfanning, line 10: >public [](http://example.com/codeflow?start=8&length=6)

Why is this data public?

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
michaelcfanning: S_ is not a convention for any standard that I'm aware of.

:) if we prefer s_ it needs to remain that way and you'll need to disable warnings.

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
```diff
@@ -29,11 +29,11 @@ internal class SpamTestRule : Skimmer<AnalyzeContext>, IOptionsProvider
         internal static Random s_random = new Random(s_seed);
 
         [ThreadStatic]
-        internal static TestRuleBehaviors s_testRuleBehaviors;
+        private static TestRuleBehaviors s_testRuleBehaviors;
```
michaelcfanning, line 32: >private [](http://example.com/codeflow?start=8&length=7)

You are going to break upstream consumers of this property, I believe. We shouldn't do it but this data is visible to some SPAM things, I think, check the internals visible to attribute data.

The 'right' thing to do would be to get over it and just elevate this to a public static property.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
schlaman-ms: Added very complete XML comments to explain the 2 suppressions.

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
schlaman-ms: Fixed the embarrassingly bad misspell of Alibaba (oddly I chose to only misspell in one location - did the key get stuck down ??) :)

Setup a meeting to discuss coding standards in detail.

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
schlaman-ms: The existing code had it as public but it should have been private as that field is not referenced outside of this class and clearly should be private.
Changed it to private.

Src/Sarif.PatternMatcher.Sdk/ExtensionMethods.cs
```diff
@@ -10,6 +10,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class ExtensionMethods
     {
+        public static char RedactionChar => '?';
```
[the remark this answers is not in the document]
schlaman-ms: Yes it is and I have changed it to a const char.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
schlaman-ms: New user story created:

[Add version details to ToolComponent in CreateSkimmers](https://dev.azure.com/mseng/1ES/_workitems/edit/2066916)

schlaman-ms: There are about 50 TODOs and about 13 TBDs in BinSkim.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
schlaman-ms: I do have to say the 'you touched it' idea can be a tad onerous and discourage doing a lot of cleanup that has high value but is too tedious to attack the owned content in the same pass.

I think there is great value in having multiple passes/waves of cleanup with different types of things done in each wave and if you mix too many things it makes it quite inefficient.

And personally, I would be very happy to NOT advocate for caring about code or property placement in a file.
It does however, feel messy to have fields/properties after the construction and methods. It does make sense to have private methods at the end of the class but that is a minor concern. There are so many subtle "this before this" rules it can become an almost unsolvable puzzle to find the perfect combination that satisfies all the placement rules.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
schlaman-ms: This falls under the should the 'if you touch it, you own it' rule.

Ultimately if we decide we don't care about the order then this would not have come up and there are literally hundreds of comments I'm guessing that need help and that takes away time from perhaps other higher yield tasks or  perhaps a dedicated cleanup pass for comments.

Src/Sarif.PatternMatcher/JsonLogicalLocationProcessor.cs
schlaman-ms: Improved the XML comments.
Aware that typical comments are more complete.
Had to stop to study code to get a sense of something more appropriate to write.
Trying to balance where to spend time.

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
schlaman-ms: Ok, now I remember, the VS auto fix was to create a property and s_testRuleBehaviors turned into the backing field which is why it could become private but the property became S_testRuleBehaviors.

But, I have restored everything back and for now there is a warning about it:
warning CS0649: Field 'SpamTestRule.s_testRuleBehaviors' is never assigned to, and will always have its default value

Src/Test.UnitTests.Sarif.PatternMatcher/SpamTestRule.cs
schlaman-ms: Addressed this in the prior comment about s_testRuleBehaviors.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
```diff
@@ -11,6 +11,21 @@
     <PackageId>Sarif.PatternMatcher.AzureDevOpsConfiguration</PackageId>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <MSBuildWarningsAsMessages>
+		// MSB3277 - Found conflicts with different versions of the same
```
suvamM, line 16: Nit: `.` at the end of the sentence. Also "found conflicts *among* different versions..." might be better.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
```diff
@@ -11,6 +11,21 @@
...
+		//  dependent assembly that could not be resolved. Specifically
+		//  for Newtonsoft where we need older versions for compatibility
+		// MSB3030 - This error occurs when a file that was supposed to be
+		//  copied isn't found at the expected location.
+		//  There are copy commands that have a continueOnError=true because
```
suvamM, line 21: The text seems a little convoluted to me. How about: "Some copy commands have continueOnError=true since the copy fails, and then eventually succeeds. MSB3030 errors, in such cases, are false positives and are therefore OK to suppress. 
TODO: Investigate why the copy failure occurs even though the operation eventually succeeds."

I am not sure what value is added by saying this has been the case for almost two years. Better to add the TODO so that we can track this, and hopefully mitigate it properly.

Src/Plugins/SalModernization/SalModernization.csproj
suvamM: @schlaman-ms To me, the problems of the duplication is showing up already. If you make a refinement to the XML comment, that adds a follow-on work item to make sure it's updated everywhere. The `build.common.props` was just a suggestion. Can we not create a new suppression file, and make the relevant projects reference that instead? Do we currently have a scenario where we are suppressing things other than MSB3277 and MSB3030?

Src/Plugins/Security/Utilities/AlibabaEcsRequestSigner.cs
suvamM: +1. I couldn't parse this.

Src/Sarif.PatternMatcher/AnalyzeCommand.cs
suvamM: @schlaman-ms Thanks for opening the work item. But I think there is a subtlety here. To me, the value of work items come from the fact that it forms a connected component. For the work item you created, there are the following issues:

1. If someone is browsing the code and reaches this TODO, there is no way to click and jump to the corresponding work item.
2. If someone is browsing our ADO board and reaches the work item, there is no way to point back to this code.
3. Since this is open-source, and we drop in an ADO work-item link, it will be broken for most people. 

We should think of a coding standard for this as well:

1. For Todos in open-source, raise issues. Ensure the issue is linked from code, and there are links to the code from the issue.
2. For internal projects, replace "issues" with ADO work items.

@michaelcfanning what do you think about this suggestion?

Src/Sarif.PatternMatcher/AnalyzeContext.cs
suvamM: I have a preference for adding public methods before private ones: if someone else is reading the class to understand how to _interact_ with it (and not necessarily how it implements things internally), then having the public stuff at the top makes it easier to understand. One doesn't need to traverse different parts of the class. 

Personal preference here.

suvamM (changes requested):

Src/Sarif.PatternMatcher.Sdk/Crc32.cs
```diff
@@ -7,6 +7,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk
 {
     public static class Crc32
     {
+        public static readonly uint[] Crc32CastagnoliTable = CreateCrcTable(Crc32CastagnoliPolynomial);
```
michaelcfanning, line 10: >public [](http://example.com/codeflow?start=8&length=6)

Why is this public? This public array of items is mutable, i.e., any caller can change index 0 this data and break it. That's not good.

This data should either not be public or we should be exposing an immutable array.

Src/Sarif.PatternMatcher/AnalyzeContext.cs
michaelcfanning: Yes, you touched it. When you touch code, in our project, people should take a careful look at it. 

Here is an updated code comment. I invested more time in discussing this with you than authoring it. And btw - improving code comments is a high yield task in my opinion. :)


    /// <summary>
   /// Gets or sets a map that associates file text with data used in RE2
   /// searching, specifically, a String8 form of the file text and arrays
   /// that comprise a map of UTF8 to UTF16 byte indices. This data is
   /// required to rationalize RE2 match segments with original .NET strings.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
michaelcfanning: So, this is an appeal to authority, and I encourage everyone to reject comments on the basis of 'Michael said so'. Instead, please try to communicate the technical reason I provided: the SARIF SDK is consumed by other projects that themselves may consume outdated versions of Newtonsoft. If we update our references, we will force those projects to upgrade as well, which may invalidate our assembly from reuse. 

I say lots of things, and I'm often wrong or my information is incomplete. Trust me if you haven't had direct experience of it already.

schlaman-ms: So a better answer would be:

We have consumers of this tool that use older frameworks and if we updated to newer versions of Newtonsoft then that would break the compatibility with consumers that are not as up to date.
Because of that, it is necessary to maintain older versions within the tool and this creates warnings that there are version conflicts. For this reason I suppressed the warnings rather than update the versions.

Src/Plugins/SalModernization/SalModernization.csproj
shaopeng-gh: I agree with it is better if possible, to define in one place, even if it applies to more projects those not have to have it.
applies to more projects has not have major impact.
The benefit is 
less duplicate code, 
easier to maintain, 
and if the project that currently does not need starts to have the e.g. newton in the future then it is already covered no need to add one by one again.

Src/Plugins/AzureDevOpsConfiguration/AzureDevOpsConfiguration.csproj
shaopeng-gh: I think this is wrong, the correct xml comment should be: `<!-- ... -->` ?

PR #718, opened 2023-03-07
Disk space failure analytics

suvamM:
This PR makes the following changes:

1. Removes the schedule for running CodeQL jobs.
2. Prints out the disk usage immediately after the SARIF Pattern Matcher build, for diagnostic purposes.

PR #770, opened 2023-06-15
Restore schedule as part of solving GitHub action that is broken.

schlaman-ms:
## Restoring cron schedule that was pulled out in #718 

For now, we will leave the schedule on so the GitHub actions will trigger on every action to increase likelihood of hitting the bug.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -16,6 +16,8 @@ on:
     branches: "*"
   pull_request:
     branches: [ main ]
+  schedule:
+    - cron: '0 0 * * 0'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -16,6 +16,8 @@ on:
     branches: "*"
   pull_request:
     branches: [ main ]
+  schedule:
+    - cron: '0 0 * * 0'
```
suvamM, line 20: @schlaman-ms What is the interpretation of `0 0 * * 0`?

michaelcfanning (approved):

PR #773, opened 2023-06-21
Remove `Octokit` dependencies and refactor the GitHub Pat validator

yongyan-gh:
## Changes

- Updated `SEC101/006.GitHubLegacyPat` validator to use HTTP REST API request.
- Remove binary dependency 'Octokit.dll'.


LingZhou-gh:
>         protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary<string, FlexMatch> groups)

Just note that this SEC101_006 rule has different implementations for Static Validator in other plugins: Security.Internal and Security.PushProtection. 

---
Refers to: Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs:15 in d52894a. [](commit_id = d52894a185ecc89fd4ea347400ea02ce1e57ec6b, deletion_comment = False)

Src/Plugins/Security/SecurePlaintextSecretsValidators/GitHubTokenValidationHelper.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.Http.Headers;
+
+using Microsoft.CodeAnalysis.Sarif.PatternMatcher.Sdk;
+
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
```
LingZhou-gh, line 14: >Security [](http://example.com/codeflow?start=62&length=8)

Exactly same as GitHubTokenValidationHelper.cs in Security.Internal?

yongyan-gh:
>         protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary<string, FlexMatch> groups)

yes PushProtection has the checksum verification I think it comes from internal source (GHAS) so I didn't add it to the public validator. @HulonJenkins what do you think?

---
In reply to: [1603000582](https://github.com/microsoft/sarif-pattern-matcher/pull/773#issuecomment-1603000582) [](http://example.com/codeflow?ancestors=1603000582,1603000582)

---
Refers to: Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_006.GitHubLegacyPatValidator.cs:15 in d52894a. [](commit_id = d52894a185ecc89fd4ea347400ea02ce1e57ec6b, deletion_comment = False)

Src/Plugins/Security/SecurePlaintextSecretsValidators/GitHubTokenValidationHelper.cs
```diff
@@ -0,0 +1,165 @@
...
+{
+    internal class GitHubTokenValidationHelper
+    {
+        protected virtual string[] GitHubEndpoints => new[]
```
yongyan-gh, line 18: >GitHubEndpoints [](http://example.com/codeflow?start=35&length=15)

Make the GitHubEndpoint virtual so in the internal plugin it can be overridden with additional endpoints to support verification against GH enterprise and MS internal endpoints.

Src/Plugins/Security/SecurePlaintextSecretsValidators/GitHubTokenValidationHelper.cs
yongyan-gh: almost same except `GitHubEndpoints` property. pls see my comment about it. The internal plugin can use this class and override the `GitHubEndpoints` if needed.

ReleaseHistory.md
```diff
@@ -21,8 +21,9 @@
 ## v4.5.2 Unreleased
 - DEP: Update SARIF SDK submodule from [dd8b7b8 to a7029a3](https://github.com/microsoft/sarif-sdk/compare/dd8b7b8..a7029a3). [Full SARIF SDK release history](https://github.com/microsoft/sarif-sdk/blob/a7029a3/ReleaseHistory.md). SDK v4.2.0.
 - DEP: Update `SEC101/003.GoogleApiKey` in `Security` validating Google API key using HTTP requests.
+- DEP: Update `SEC101/006.GitHubLegacyPat` in `Security` validating GitHub PAT using HTTP requests.
```
michaelcfanning, line 24: >DEP [](http://example.com/codeflow?start=2&length=3)

You need to explicitly note that we've removed 'Oktokit.dll'. The reason is that some consumers might explicitly reference this file name in their utilization of our package.

yongyan-gh: updated.

michaelcfanning (approved):
:shipit:

PR #774, opened 2023-06-22
Add dlopen command.

michaelcfanning:
@shaopeng-gh

### DIFF
--- a/Src/RE2.Managed/NativeMethods.cs
+++ b/Src/RE2.Managed/NativeMethods.cs
@@ -16,10 +16,7 @@ namespace Microsoft.RE2.Managed
             string platform =
                 RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "win" : "linux";
 
-            if (platform == "linux")
-            {
-                return;
-            }
+            bool isLinux = platform == "linux";
 
             // Strictly speaking we don't need any platform-specific code here. I leave
             // it just in case we find out that we *do* need to perform the Linux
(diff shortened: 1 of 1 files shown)


shaopeng-gh:
Not familiar with these, looks good, will try it.

shaopeng-gh (approved):
:shipit:

Src/RE2.Managed/NativeMethods.cs
```diff
@@ -34,7 +31,7 @@ static NativeMethods()
 
                 if (File.Exists(filePath))
                 {
-                    LoadLibrary(filePath);
+                    if (isLinux) { LoadLibrary(filePath); } else { LinuxMethods.dlopen(filePath, LinuxMethods.RTLD_NOW); }
                 }
             }
             else if (platform == "win")
```
yongyan-gh, line 37: >(platform == "win") [](http://example.com/codeflow?start=20&length=19)

nit: when `platform == "win"`. `isLinux` should always be false. not needed to check `isLinux` in below if condition?

yongyan-gh (approved):
:shipit:

yongyan-gh:
looks like a P/Invoke the `dlopen` method from native dynamic library in linux platform.

---
In reply to: [1603348892](https://github.com/microsoft/sarif-pattern-matcher/pull/774#issuecomment-1603348892) [](http://example.com/codeflow?ancestors=1603348892,1603348892)

shaopeng-gh:
@michaelcfanning I think you have made it for the other direction for IronRe2, looks like we can close this PR,

PR #776, opened 2023-06-25
Restore regex extensibility

michaelcfanning:
This change restores use of alternate regex engines behind the IRegex abstraction. We can now therefore run on linux if users specify the `--engine DotNet` flag.


Src/Sarif.PatternMatcher/RegexEngine.cs
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
+{
+    public enum RegexEngine
+    {
+        None,
+        DotNet,
+        CachedDotNet,
+        RE2,
+        IronRE2,
```
michaelcfanning, line 14: >IronRE2 [](http://example.com/codeflow?start=8&length=7)

Coming next.

PR #777, opened 2023-06-26
filter out the noise password={anything} in regex

LingZhou-gh:
## Changes

Minor regex change of rule 036 & 037 & 038 to filter out the noise of the following format: password=**{**anything**}** or password=**[**anything**]** or password=**(**anything**)**. 
Release record of FPC (Regex candidate reduction) is added.


LingZhou-gh:
Abandon this PR as the noise reduction are made in Static Validation phase.

PR #778, opened 2023-06-27
Release/v4.5.5

michaelcfanning:
This change enables IronRE2 as an option, running on Windows *or* Linux environments.

We now run IronRE2 for many end-to-end tests, so the behavior looks identical.

With recent ETW work we can easily test this on some real-world tests. We should also rerun the benchmark tests we authored previously.

### DIFF
--- a/Src/Plugins/Tests.Security/EndToEndTests.cs
+++ b/Src/Plugins/Tests.Security/EndToEndTests.cs
@@ -38,7 +38,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                 AnalyzeCommand.CreateSkimmersFromDefinitionsFiles(fileSystem,
                                                                   new string[] { regexDefinitionsPath },
                                                                   tool,
-                                                                  RE2Regex.Instance);
+                                                                  IronRE2Regex.Instance);
             return skimmers;
         }
 
@@ -176,7 +176,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
 
                 using (context)
                 {
-                    ValidatorBase.RegexInstance = RE2Regex.Instance;
+                    ValidatorBase.RegexInstance = IronRE2Regex.Instance;
                     AnalyzeCommand.AnalyzeTargetHelper(context, skimmers, disabledSkimmers);
                 }
             }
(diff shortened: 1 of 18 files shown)


Src/Plugins/Tests.Security/EndToEndTests.cs
```diff
@@ -38,7 +38,7 @@ public static ISet<Skimmer<AnalyzeContext>> CreateOrRetrievedCachedSkimmer(IFile
                 AnalyzeCommand.CreateSkimmersFromDefinitionsFiles(fileSystem,
                                                                   new string[] { regexDefinitionsPath },
                                                                   tool,
-                                                                  RE2Regex.Instance);
+                                                                  IronRE2Regex.Instance);
```
michaelcfanning, line 41: >IronRE2Regex [](http://example.com/codeflow?start=66&length=12)

Just to raise confidence, we run all functional tests against IronRE2. This creates a high degree of confidence it works: all outputs of all test targets is identical using this engine.

Src/Plugins/Tests.Security/SecurePlaintextSecretsValidators/SEC101_008.AwsCredentialsValidatorTests.cs
```diff
@@ -31,7 +31,7 @@ public void AwsCredentialsValidatorValidator_MockHttpTests()
             const string secret = "abc123";
             string fingerprintText = $"[id={id}][secret={secret}]";
 
-            ValidatorBase.RegexInstance = RE2Regex.Instance;
+            ValidatorBase.RegexInstance = IronRE2Regex.Instance;
```
michaelcfanning, line 34: >IronRE2Regex [](http://example.com/codeflow?start=42&length=12)

Updating these tests to IronRE2Regex makes these tests xplat compatible.

Src/RE2.Managed/NativeMethods.cs
```diff
@@ -13,20 +13,7 @@ internal static class NativeMethods
     {
         static NativeMethods()
         {
-            string platform =
```
michaelcfanning, line 16: >platform [](http://example.com/codeflow?start=19&length=8)

removing old linux re2 experiment.

Src/Sarif.PatternMatcher.Benchmark/Benchmarks/AnalyzeCommandBenchmarks.cs
```diff
@@ -55,6 +55,12 @@ public void AnalyzeCommand_SimpleAnalysisRegex2()
             AnalyzeCommand(RE2Regex.Instance);
         }
 
+        [Benchmark]
```
michaelcfanning, line 58: >Benchmark [](http://example.com/codeflow?start=9&length=9)

We should run these benchmarks!! @HulonJenkins ?

PR #783, opened 2023-07-05
Update SARIF SDK to dd3741f to enable improved xplat testing.

michaelcfanning:
This change updates SARIF SDK module strictly to avoid test failures between net452 and netstandard 2.0 runs. No product changes.


shaopeng-gh:
I have included this PR into:
https://github.com/microsoft/sarif-pattern-matcher/pull/785 so that we can see if the build and test can be all green. 
Now looks good and green, I merged that one, and close this PR.

PR #788, opened 2023-07-15
Expand RabbitMQ validator to check loose credentials expressions.

yongyan-gh:
## Changes

- Update `SEC101/041.RabbitMqCredentials` validator to check for loose credentials. e.g.
```
{
  "settings": [
    { "name": "RABBITMQ HOST", "value": "12.23.45.78" },
    { "name": "RABBITMQ PASSWORD", "value": "..." },
    { "name": "RABBITMQ PORT", "value": "5672" }, 
    { "name": "RABBITMQ USERNAME", "value": "guest" },
    ...
  ]
}
```
```
...
env:
    RABBITQ_HOST: ...
    RABBITMQ_USERNAME: ...
    RABBITMQ_PASSWORD: ...
    RABBITMQ_VHOST: ...
```
- Ignore the password in within common variable context e.g. $(...) {...} [...]
- Make the port number and virtual host name optional.
- Update official RabbitMQ.Client package to latest version.

### DIFF
--- a/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_041.RabbitMqCredentialsValidator.cs
+++ b/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_041.RabbitMqCredentialsValidator.cs
@@ -18,11 +18,18 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
         {
             if (!groups.TryGetNonEmptyValue("id", out FlexMatch id) ||
                 !groups.TryGetNonEmptyValue("host", out FlexMatch host) ||
-                !groups.TryGetNonEmptyValue("secret", out FlexMatch secret) ||
-                !groups.TryGetNonEmptyValue("resource", out FlexMatch resource))
+                !groups.TryGetNonEmptyValue("secret", out FlexMatch secret))
             {
                 return ValidationResult.CreateNoMatch();
             }
+
+            if (FilteringHelpers.PasswordIsInCommonVariableContext(secret.Value))
+            {
+                return ValidationResult.CreateNoMatch();
+            }
+
+            groups.TryGetValue("resource", out FlexMatch resource);
+            groups.TryGetValue("port", out FlexMatch port);
 
             string hostValue = FilteringHelpers.StandardizeLocalhostName(host.Value);
 
@@ -32,8 +39,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                 {
                     Id = id.Value,
                     Host = hostValue,
+                    Port = port?.Value,
                     Secret = secret.Value,
-                    Resource = resource.Value,
+                    Resource = resource?.Value,
                 },
             };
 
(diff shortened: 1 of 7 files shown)


HulonJenkins (approved):

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_041.RabbitMqCredentialsValidator.cs
```diff
@@ -18,12 +18,19 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
         {
             if (!groups.TryGetNonEmptyValue("id", out FlexMatch id) ||
                 !groups.TryGetNonEmptyValue("host", out FlexMatch host) ||
-                !groups.TryGetNonEmptyValue("secret", out FlexMatch secret) ||
-                !groups.TryGetNonEmptyValue("resource", out FlexMatch resource))
+                !groups.TryGetNonEmptyValue("secret", out FlexMatch secret))
```
HulonJenkins, line 21: We shouldn't need this if statement check for id, host, and secret now that these are required in the regex. Can just be string id = groups["id"].value for all 3

Src/Plugins/Security/Utilities/FilteringHelpers.cs
```diff
@@ -48,5 +48,24 @@ public static bool LikelyPowershellVariable(string input)
 
             return true;
         }
+
+        public static bool PasswordIsInCommonVariableContext(string secret)
```
HulonJenkins, line 52: >InCommo [](http://example.com/codeflow?start=37&length=7)

Can our rules in SPMI use this function as well after this is merged?

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -129,9 +129,14 @@
   $SEC101/038.PostgreSqlCredentialsAdoSecret=(?i)(?:password|pwd)\s*=\s*(?P<secret>[^,;"'<\s]{8,128})(?:[,;"'<\s]|$)
   $SEC101/038.PostgreSqlCredentialsAdoResource=(?i)(?:database|db|dbname)\s*=\s*(?P<resource>[^,;"'=|&\]\[><\s]+)(?:[,;"'=|&\]\[><\s]|$)
 
-  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:"]+):(?P<secret>[^@\s]+)@(?P<host>[\w_\-\:]+)\/(?P<resource>[\w]+)(?:[^0-9a-z]|$)
+  $SEC101/041.RabbitMqCredentials=(?i)amqps?:\/\/(?P<id>[^:"]+):(?P<secret>[^@\s]+)@(?P<host>[\w_-]+)(?::?(?P<port>[0-9]{4,5}))?\/(?P<resource>[\w]+)?(?:[^0-9a-z]|$)
```
yongyan-gh, line 132: The change here is to add capture group for "port" explicitly and make the "resource" group optional

Src/Plugins/Security/Utilities/FilteringHelpers.cs
yongyan-gh: Yes, the function I copied from your previous implementation. The other projects has a reference to Security can use it.

PR #789, opened 2023-07-17
Re-enable MongoDbCredentials dynamic validator

yongyan-gh:
## Changes

- Add dynamic validation of rule `SEC101/033.MongoDbCredentials` back to `Security`.
- Add MongoDB official package `MongoDB.Driver` latest version `2.20.0` to `Security`.
- Update `SEC101/033.MongoDbCredentials` rule regex to handle different schema and connection options.


Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_033.MongoDbCredentialsValidator.cs
```diff
@@ -19,45 +23,65 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
...
+
             var validationResult = new ValidationResult
             {
                 Fingerprint = new Fingerprint()
                 {
                     Id = id.Value,
                     Host = host.Value,
                     Secret = secret.Value,
+                    Scheme = schema?.Value,
+                    QueryString = options?.Value,
```
HulonJenkins, line 37: > QueryString = options?.Value, [](http://example.com/codeflow?start=19&length=30)

Looking at fingerprint fields, is queryString the best fit? I ask because queryString was created to handle a special case in a different rule, so questioning its use here.

yongyan-gh: then I should not use `QueryString`, how about `Part`?
this is the name for the options parts of the connection string 
e.g. "mongodb://user:password@example.com:10255/?ssl=true&retrywrites=false&replicaSet=globaldb&maxIdleTimeMS=120000", these key value pairs after port number are the options for the connection.

HulonJenkins: No, I was just asking to ask, not asking for a change. After looking more, queryString is correct here. I'll try to be more explicit when I'm asking for a change vs just asking to be curious.

ReleaseHistory.md
```diff
@@ -24,6 +24,8 @@
 - BRK: Replaced `MatchLengthToDecode` property of `MatchExpression` with new `Base64EncodingMatch` class to support detecting Base64 string from strings of specific length range. [#790](https://github.com/microsoft/sarif-pattern-matcher/pull/790)
 - FNC: Update `SEC101/041.RabbitMqCredentials` in `Security` to check loose credential combinations. [#788](https://github.com/microsoft/sarif-pattern-matcher/pull/788)
 - FPD: Removed dynamic analysis entirely for `SEC101/047.CratesApiKey` rule due to outdated validation always returning status code 200 to all tokens. No API endpoint seems to return different status codes to distinguish between valid and invalid API keys. [#786](https://github.com/microsoft/sarif-pattern-matcher/pull/786)
+- DEP: Add package `MongoDB.Driver` 2.20.0. [#789](https://github.com/microsoft/sarif-pattern-matcher/pull/789)
```
HulonJenkins, line 27: >EP:  [](http://example.com/codeflow?start=3&length=4)

nit, keep release history entries sorted by type

Src/Plugins/Security/Security.SharedStrings.txt
```diff
@@ -96,7 +96,7 @@
   $SEC101/027.MailChimpApiKey=(?:[^0-9a-f]|^)(?P<secret>[0-9a-f]{32}-us[0-9]{12})
   $SEC101/029.AlibabaCloudAccessKey=(?:^|[^0-9A-Za-z])(?P<id>LTAI([0-9A-Za-z]{12}|[0-9A-Za-z]{20}))(?:[^0-9A-Za-z]|$)
   $SEC101/029.AlibabaCloudSecretKey=(?:^|[^0-9A-Za-z\-~_+/%\\])(?P<secret>[0-9A-Za-z]{30})(?:[^0-9A-Za-z\-~_+/=%\\]|$)
-  $SEC101/033.MongoDbCredentials=(?i)mongodb(?:[^:]{0,50})?:\/\/(?P<id>[0-9a-z]+):(?P<secret>[^@\s]{1,200})@(?P<host>[^\/;"<\s,]+)
+  $SEC101/033.MongoDbCredentials=(?i)mongodb(?P<schema>\+srv)?:\/\/(?P<id>[0-9a-z_-]+):(?P<secret>[^@\s]{1,200})@(?P<host>[^\/;"<\s,]+)(?P<options>/[0-9a-z?&@=_-]+)?
```
michaelcfanning, line 99: >schema [](http://example.com/codeflow?start=48&length=6)

this looks like a constant value. why have we created a named group for it?

am i wrong? doesn't this pattern *always* match `+srv` and only that?

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_033.MongoDbCredentialsValidator.cs
```diff
@@ -1,23 +1,26 @@
...
     {
         protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary<string, FlexMatch> groups)
         {
-            if (!groups.TryGetNonEmptyValue("id", out FlexMatch id) ||
-                !groups.TryGetNonEmptyValue("host", out FlexMatch host) ||
-                !groups.TryGetNonEmptyValue("secret", out FlexMatch secret))
-            {
-                return ValidationResult.CreateNoMatch();
-            }
+            groups.TryGetValue("id", out FlexMatch id);
```
michaelcfanning, line 19: >TryGetValue [](http://example.com/codeflow?start=19&length=11)

why are you using try value? your regexes seem to require some of this data?

Src/Plugins/Security/Security.SharedStrings.txt
michaelcfanning, line 99: >schema [](http://example.com/codeflow?start=48&length=6)

why is this named 'schema' and not 'scheme' (if it even requires us to have a named group.

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_033.MongoDbCredentialsValidator.cs
yongyan-gh: "mongodb+srv://" format is used to connect to a DNS host to resolve multiple host names. Its not compatible with the standard format "mongodb://".

https://www.mongodb.com/docs/manual/reference/connection-string/#srv-connection-format

Src/Plugins/Security/Security.SharedStrings.txt
yongyan-gh: "mongodb+srv://" format is used to connect to a DNS host to resolve multiple host names. Its not compatible with the standard format "mongodb://".

https://www.mongodb.com/docs/manual/reference/connection-string/#srv-connection-format

yongyan-gh: thinking about capturing the whole protocol part using `(?P<protocol>mongodb|mongodb\+srv)` instead, will it be more clear?

michaelcfanning (approved):

PR #790, opened 2023-07-19
Support detecting Base64 string from strings of specific length range

yongyan-gh:
## Changes

This is a breaking change, the rules with `MatchLengthToDecode` property set will break.
The solution is to replace it with `Base64EncodingMatch` property.

E.g. The rule definition
```
        {
          "Id": "SEC101/102",
          "Name": "AdoPat",
          "MatchLengthToDecode": 52,
          "ContentsRegex": "$SEC101/102.AdoPat",
          ...
        },
```
Need to be updated to
```
        {
          "Id": "SEC101/102",
          "Name": "AdoPat",
         "Base64EncodingMatch": { "MinSourceLength": 52, "MaxSourceLength": 52 },
          "ContentsRegex": "$SEC101/102.AdoPat",
          ...
        },
```

- Removed `MatchLengthToDecode` int property from `SearchSkimmer` class.
- Using a new class `Base64EncodingMatch` to present the decoded string length range.
- Updated existing rules using `MatchLengthToDecode` to `Base64EncodingMatch`.
- Unit tests added.

### DIFF
--- a/Src/Sarif.PatternMatcher/MatchExpression.cs
+++ b/Src/Sarif.PatternMatcher/MatchExpression.cs
@@ -51,7 +51,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
         /// decoded string to the match expression. No base64-decoding
         /// occurs when this property is 0 or less.
         /// </summary>
-        public int MatchLengthToDecode { get; set; }
+        public Base64EncodingMatch Base64EncodingMatch { get; set; }
 
         public Dictionary<string, string> Properties { get; set; }
 
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -31,7 +31,8 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
         public readonly IList<MatchExpression> MatchExpressions;
 
         private const string DefaultHelpUri = "https://github.com/microsoft/sarif-pattern-matcher";
-        private const string Base64DecodingFormatString = "\\b(?i)[0-9a-z\\/+]{{{0},{1}}}[=]{{0,2}}";
+        private const string Base64DecodingFixLengthFormatString = "\\b(?i)[0-9a-z\\/+]{{{0}}}={{{1}}}";
+        private const string Base64DecodingVarLengthFormatString = "\\b(?i)[0-9a-z\\/+]{{{0},{1}}}={{0,2}}";
         private const string Base64DecodingRegex = "^(?i)(?:[0-9a-z\\/+]{4})*(?:[0-9a-z\\/+]{2}==|[0-9a-z\\/+]{3}=)?$";
 
         private static readonly Regex namedArgumentsRegex =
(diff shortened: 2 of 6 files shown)


Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -209,25 +208,28 @@ public override void Analyze(AnalyzeContext context)
...
-                    int unpaddedLength = (int)Math.Ceiling(decimal.Divide(unencodedLength * 8M, 6M));
-                    int paddedLength = 4 * (int)Math.Ceiling(decimal.Divide(unencodedLength, 3M));
+                    int unpaddedMinLength = (int)Math.Ceiling(decimal.Divide(matchExpression.Base64EncodingMatch.MinSourceLength * 8M, 6M));
+                    int unpaddedMaxLength = (int)Math.Ceiling(decimal.Divide(matchExpression.Base64EncodingMatch.MaxSourceLength * 8M, 6M));
 
-                    // Create proper regex for base64-encoded string which includes padding characters.
+                    // Create proper regex for all strings matches base64 charset and length requirements
+                    // But these strings may be not valid base64 strings since the padding length are vary
+                    // for different length source string.
                     string base64DecodingRegexText =
```
yongyan-gh, line 220: >base64DecodingRegexText [](http://example.com/codeflow?start=27&length=23)

Couldn't figure out a single regex pattern can match all base64 encoded string of a specific length range, because the padding pattern varies.

Have to first detect the possible base64 string match the length requirement, then check if they are valid base64 strings.

let me know if any better idea

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -33,7 +31,8 @@ public class SearchSkimmer : Skimmer<AnalyzeContext>
         public readonly IList<MatchExpression> MatchExpressions;
 
         private const string DefaultHelpUri = "https://github.com/microsoft/sarif-pattern-matcher";
-        private const string Base64DecodingFormatString = "\\b(?i)[0-9a-z\\/+]{0}";
+        private const string Base64DecodingFormatString = "\\b(?i)[0-9a-z\\/+]{{{0},{1}}}[=]{{0,2}}";
+        private const string Base64DecodingRegex = "^(?i)(?:[0-9a-z\\/+]{4})*(?:[0-9a-z\\/+]{2}==|[0-9a-z\\/+]{3}=)?$";
```
HulonJenkins, line 35: >^ [](http://example.com/codeflow?start=52&length=1)

Do we need to assert start of line and end of line for this regex?

yongyan-gh: here its used to match the whole string if a valid base64.
without the start/end line, it will find some false-positives.
e.g. "abcdef" is an invalid base64 missing padding "==" at the end.
without the start/end line check, it will find a match on "abcd".

Src/Sarif.PatternMatcher/SearchSkimmer.cs
suvamM: @yongyan-gh Trying to understand this better. Can we not parameterize the regular expression based on the values specified in the range of `Base64EncodingMatch`?

yongyan-gh: @suvamM can you pls elaborate more about your idea? do you mean ignore the length check and find all valid base64 strings?

Src/Test.UnitTests.Sarif.PatternMatcher/SearchSkimmerTests.cs
```diff
@@ -29,7 +28,11 @@ public class SearchSkimmerTests
 
             return new MatchExpression
             {
-                MatchLengthToDecode = Guid.NewGuid().ToString().Length,
+                Base64EncodingMatch = new Base64EncodingMatch
+                {
+                    MinSourceLength = Guid.NewGuid().ToString().Length,
```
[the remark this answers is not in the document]
yongyan-gh: This test assumes to detect the content which is a base64 string encoded from a guid string. The length of the guid string is fixed. Will replace it with the fix length.

Src/Sarif.PatternMatcher/MatchExpression.cs
```diff
@@ -51,7 +51,7 @@ public class MatchExpression
         /// decoded string to the match expression. No base64-decoding
         /// occurs when this property is 0 or less.
         /// </summary>
-        public int MatchLengthToDecode { get; set; }
+        public Base64EncodingMatch Base64EncodingMatch { get; set; }
```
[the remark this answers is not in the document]
yongyan-gh: any suggestion for the class name and property name?

HulonJenkins (approved):
:shipit:

PR #795, opened 2023-07-21
Provide stable id look-up for validators.

michaelcfanning:
We previously used the friendly (unstable) rule name to map to validators. The reason is that this data tends to consist of readable terms that are appropriate for type names. But, this makes the system fragile: we can't find validators for historical rule names form historically generated log files.

Now we have a descriptor that tags a rule with its stable rule id. And the validator retrieval logic operates strictly from this data.

Note that in addition to the core change, I added a little (currently disabled) test that can be updated with a directory path to auto-inject this information for files that match the rule validator file name convention.

@marmegh @Jeremiah-Johnson

@HulonJenkins, note this change merges into the commit associated with v4.3.17, to assist in servicing.

### DIFF
--- a/Src/Sarif.PatternMatcher/SearchSkimmer.cs
+++ b/Src/Sarif.PatternMatcher/SearchSkimmer.cs
@@ -888,8 +888,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher
             if (_validators != null && matchExpression.IsValidatorEnabled)
             {
                 string filePath = context.CurrentTarget.Uri.GetFilePath();
-                string ruleName = matchExpression.Name ?? reportingDescriptor.Name;
-                IEnumerable<ValidationResult> validationResults = _validators.Validate(ruleName,
+                IEnumerable<ValidationResult> validationResults = _validators.Validate(matchExpression.Id,
                                                                                        context,
                                                                                        mergedGroups,
                                                                                        groups,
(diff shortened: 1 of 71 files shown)


suvamM (approved):
Looks good to me.

Src/Sarif.PatternMatcher/ValidatorsCache.cs
```diff
@@ -33,52 +33,45 @@ public ValidatorsCache(IEnumerable<string> validatorBinaryPaths = null, IFileSys
...
+                if (_ruleIdToValidationMethods == null)
                 {
                     lock (sync)
                     {
-                        if (_ruleNameToValidationMethods == null)
+                        if (_ruleIdToValidationMethods == null)
                         {
-                            _ruleNameToValidationMethods ??= LoadValidationAssemblies(ValidatorPaths);
+                            _ruleIdToValidationMethods ??= LoadValidationAssemblies(ValidatorPaths);
                         }
```
HulonJenkins, line 47: Isn't this if statement redundant because of the null assignment operator?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -991,8 +990,7 @@ private void RunMatchExpressionForSingleLineRegexes(AnalyzeContext context, Matc
 
                 if (_validators != null && matchExpression.IsValidatorEnabled)
                 {
-                    string ruleName = matchExpression.Name ?? reportingDescriptor.Name;
-                    IEnumerable<ValidationResult> validationResults = _validators.Validate(ruleName,
+                    IEnumerable<ValidationResult> validationResults = _validators.Validate(reportingDescriptor.Id,
```
yongyan-gh, line 993: >reportingDescriptor.Id [](http://example.com/codeflow?start=91&length=22)

why use `reportingDescriptor.Id` here, not `matchExpression.Id` like other 2 changes?

LingZhou-gh (approved):
:shipit:

PR #797, opened 2023-07-26
Dynamic validator update for NpmCredentials and report credentials as unknown for generic uri.

LingZhou-gh:
Dynamic validator update for NpmCredentials and report credentials as unknown for generic uri.

### DIFF
--- a/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
+++ b/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
@@ -101,7 +101,7 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NotFound ||
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NonAuthoritativeInformation)
                 {
-                    message = $"The Uri {uri} is generic, no way to validate teh credential with for account {id}";
+                    message = $"The Uri {uri} is generic, no way to validate the credential with for account {id}";
 
                     return ValidationState.Unknown;
                 }
--- a/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
+++ b/Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
@@ -101,7 +101,9 @@ namespace Microsoft.CodeAnalysis.Sarif.PatternMatcher.Plugins.Security
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NotFound ||
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NonAuthoritativeInformation)
                 {
-                    return ValidationState.NoMatch;
+                    message = $"The Uri {uri} is generic, no way to validate teh credential with for account {id}";
+
+                    return ValidationState.Unknown;
                 }
 
                 string credentials = Convert.ToBase64String(Encoding.ASCII.GetBytes(string.Format("{0}:{1}", id, secret)));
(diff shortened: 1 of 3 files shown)


Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
```diff
@@ -101,7 +101,9 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NotFound ||
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NonAuthoritativeInformation)
                 {
-                    return ValidationState.NoMatch;
+                    message = $"The Uri {uri} is generic, no way to validate teh credential with for account {id}";
```
yongyan-gh, line 104: >message [](http://example.com/codeflow?start=20&length=7)

can you show an example of entire message of unknown results?

i remember this message will be concat with another message so in some case it starts with a space and lowercase.

LingZhou-gh: npm-cli-login -u userNameHere -p P@s$w0rdHere S -e emailHere@microsoft.com -r https://registry.npmjs.org

HulonJenkins, line 104: >teh [](http://example.com/codeflow?start=77&length=3)

spelling here and in test file

LingZhou-gh: yes, corrected.

HulonJenkins: >{uri} is generi [](http://example.com/codeflow?start=40&length=15)

I think this update is an overcorrection. If a uri returns NotFound we would still return the new message, which doesn't seem right to me.

LingZhou-gh: I update the message again.

michaelcfanning: 1. bad grammar. 2. no use of apostrophes to enquote dynamic data. 3. incorrectly capitalized words. 

Use this:

$'{uri}' is a generic URI that can't be used to validate account '{account}'"

Please pay particular attention to this note, which is commonly given in code review. If you are confused about what I mean when I refer to 'dynamic data', we can follow-up in stand-up. It is always fine, of course, to ask for suggestions for user-facing content, which Hulon or I can provider (as I have here!). :)

Src/Plugins/Security/SecurePlaintextSecretsValidators/SEC101_044.NpmCredentialsValidator.cs
```diff
@@ -101,7 +101,9 @@ protected override IEnumerable<ValidationResult> IsValidStaticHelper(IDictionary
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NotFound ||
                     responseWithNoCredentials.StatusCode == HttpStatusCode.NonAuthoritativeInformation)
                 {
-                    return ValidationState.NoMatch;
+                    message = $"The Uri {uri} is generic, no way to validate the credential with for account {id}";
```
michaelcfanning: >generic [](http://example.com/codeflow?start=49&length=7)

What does it mean, the 'URI' is generic? This code path seems to suggest one possibility is the URI can't be found, how is that generic? 

Why does a return value of 'OK' mean it is too generic?

Are we properly describing this condition? Do we need separate error messages for each condition?

Why are we even fixing this? Your change makes this URI return 'unknown' which means we will validate it over and over again. Why? i.e., why isn't the status code conclusive? 

these don't look like transient conditions, they look permanent.

michaelcfanning (changes requested):
:clock1:

LingZhou-gh:
This PR has been replaced by new PR
https://github.com/microsoft/sarif-pattern-matcher/pull/799

PR #798, opened 2023-07-27
Sniff literals

michaelcfanning:


Src/Sarif.PatternMatcher/AnalyzeCommand.cs
```diff
@@ -380,7 +380,6 @@ public override AnalyzeContext InitializeGlobalContextFromOptions(AnalyzeOptions
             }
 
             ValidatorBase.RegexInstance = GetRegexEngine(context.RegexEngine);
-            context.PluginFilePaths = options.PluginFilePaths.Any() ? new StringSet(options.PluginFilePaths) : context.PluginFilePaths;
```
michaelcfanning, line 383: >PluginFilePaths [](http://example.com/codeflow?start=20&length=15)

This property is already handled in the base class.

michaelcfanning:
This change introduces 'sniff literals'. A sniff literal is a small snippet of text that is used as an IndexOf-driven pre-check of a file: if IndexOf can't find the string literal in a file, for a specific rule, that rule is filtered away from analysis. Performance analysis tells us that IndexOf is ~18x faster than calling an individual RE2-driven regex. The reason is that this method calls very low level hardware intrinsics (when available) that perform string comparisons. We gain an additional advantage from this approach: certain sniff literals are common between multiple rules/match expressions.

This change also includes events that allow us to collect timing information for the sniff literal phase. This change does *not* yet include the formulation of sniff literals for core plugins, these are being developed separately.

ReleaseHistory.md
```diff
@@ -23,6 +23,7 @@
 - DEP: Upgrade `Microsoft.Security.Utilities` from 6.2.1 to 6.5.0. [#788](https://github.com/microsoft/sarif-pattern-matcher/pull/788)
 - FNC: Update `SEC101/041.RabbitMqCredentials` in `Security` to check loose credential combinations. [#788](https://github.com/microsoft/sarif-pattern-matcher/pull/788)
 - FPD: Removed dynamic analysis entirely for `SEC101/047.CratesApiKey` rule due to outdated validation always returning status code 200 to all tokens. No API endpoint seems to return different status codes to distinguish between valid and invalid API keys. [#786](https://github.com/microsoft/sarif-pattern-matcher/pull/786)
+- NEW: Add pre-regex file `sniff` to decide whether to run match expressions.
```
suvamM, line 26: "Add pre-regex file `sniff` _option_..."

Src/RE2.Managed/IronRE2Regex.cs
```diff
@@ -94,7 +94,7 @@ private static Options ToOptions(System.Text.RegularExpressions.RegexOptions dot
             int lastUtf16Index = 0;
 
             int index = 0;
-            while (index < input.String8.Array.Length)
+            while (index < input.String8.Array?.Length == true)
```
suvamM, line 97: @michaelcfanning Basic question: why is the `== true` needed?

Src/Sarif.PatternMatcher/SearchSkimmer.cs
```diff
@@ -756,6 +756,48 @@ private static Region PopulateTextRegionProperties(AnalyzeContext context, Regio
 
         private void RunMatchExpression(FlexMatch binary64DecodedMatch, AnalyzeContext context, MatchExpression matchExpression)
         {
+            bool continueProcessing = false;
+            if (context.EnableSniffLiterals && matchExpression.SniffLiterals?.Count > 0)
+            {
+                DriverEventSource.Log.RuleReserved1Start(SpamEventNames.RunRulePhase0Sniff,
```
suvamM, line 762: Why is the event called `RuleReserved1Start`?

suvamM (approved):

PR #807, opened 2024-01-02
Update the RE2 submodule

thatsagreatidea:
- Adding missing submodule directory for RE2.
  This allows `git submodule update --init --recursive` to work on a clean enlistment.

- Pin RE2 submodule to commit 954656f47fe8fb505d4818da1e128417a79ea500.
  This allows RE2.native.sln to build.
  There are commits in RE2 after 954656f47fe8fb505d4818da1e128417a79ea500. But they require project file updates to build successfully. This change only restores the previous ability to build RE2 successfully.

Note: Can the reviewer smoke test the compiled RE2 library with the pattern matcher before closing the PR? I don't have the capacity and don't know if there were any breaking changes.


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "Src/RE2.Native/re2"]
 	path = src/RE2.Native/re2
-	url = https://github.com/google/re2.git
+	url = https://github.com/google/re2.git
```
michaelcfanning, line 3: Hello, @thatsagreatidea, I'm not sure that we should do that. What we've recently preferred is to simply check in an optimized release build of RE2.dll. Would that work for your purposes?

thatsagreatidea: yes, thank you

